PR #227, opened 2018-08-28
Calc_psf docstring edits now in a decorator

shanosborne:
The combination of the JWInstrument calc_psf docstring and the Poppy Instrument calc_psf docstring is now located in the combine_docstrings method and applied to the JWInstrument calc_psf as a decorator.

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -607,6 +607,19 @@ class SpaceTelescopeInstrument(poppy.instrument.Instrument):
         filterfits.close()
         return band
 
+    @staticmethod
+    def include_other_docstrings():
+        def wrapper(func):
+            # Allow users to see poppy calc_psf docstring too
+            if func.__name__ == "calc_psf":
+                ind0 = func.__doc__.index("add_distortion")  # pull the new parameters
+                ind1 = getattr(SpaceTelescopeInstrument, func.__name__).__doc__.index("Returns")  # end of parameters
+                func.__doc__ = getattr(SpaceTelescopeInstrument, func.__name__).__doc__[0:ind1] + func.__doc__[ind0:] \
+                               + getattr(SpaceTelescopeInstrument, func.__name__).__doc__[ind1:]
+
+            return func
+        return wrapper
+
 
 #######  JWInstrument classes  #####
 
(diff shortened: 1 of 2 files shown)


webbpsf/webbpsf_core.py
```diff
@@ -752,6 +767,7 @@ def _get_fits_header(self, result, options):
                                            "[arcmin] Det. pos. in telescope V2,V3 coord sys"), after=True)
         result[0].header["APERNAME"] = (self._detectors[self._detector], "SIAF aperture name")
 
+    @SpaceTelescopeInstrument.combine_docstrings()
```
mperrin: I'd like to suggest moving the docstring decorator out of the `SpaceTelescopeInstrument` class. It could be a standalone method in `utils.py` for instance. My motivation for suggesting this is to keep the namespace of class methods a little cleaner - i.e. it's not helpful for tab-completion on every WebbPSF class instance to include `combine_docstrings`, since this is never a function some user would want to call interactively. So we should hide it behind the curtain, so to speak.

mperrin: @shanosborne I wanted to check in with you on this - did you see my above comment on moving this function aside somewhere it won't show up in tab completion? If you agree with that, we could do that then I can merge this PR

shanosborne: Yes - I have seen this and am working on moving it to utils.py. However, for some reason now the docstring doesn't show up when I do shift+tab inside the parentheses in an ipython notebook (nor does it show up when I use the question mark in ipython or the notebook). So I'm working on figuring out why that happens.

shanosborne: Update: I fixed things by changing combine_docstrings to be a class decorator

mperrin:
I’m going to re-run the test on this since I just merged the PR #239 which should fix the Travis error I hope.

shanosborne:
@mperrin Do I need to pull the changes from the PR you just merged? Is that the problem?

mperrin:
Just stand by for now - this is still showing the results from Travis build 11 which is the one I referred to above (“just literally re ran it identically”), but build 14 (running now) is the one from after my close and reopen.

shanosborne:
@robelgeda re-ran the Travis build and it looks like the update works now (everything passes). So I'm not sure what was happening before

mperrin:
Woo! Ok that makes me happy since I had no idea why it was timing out yesterday.

Issue #226, opened 2018-08-28
NIRCam SW distortion center position


mperrin:
<a href="https://github.com/JarronL"><img src="https://avatars3.githubusercontent.com/u/17557291?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [JarronL](https://github.com/JarronL)**
_Friday Jul 20, 2018 at 00:49 GMT_
_Originally opened as https://github.com/mperrin/webbpsf/issues/226_

----

In checking out the distortion results in v0.7.0, I noticed resulting PSFs getting shifted in ways I was not expecting. These shifts depend on values specified in `detector_position` attribute. For the SW modules, I would have expected any PSF shifting to be relative to the center of the entire focal plane rather than the center of a given detector. 

For instance, setting `detector='A1'` and `detector_position=(1024,1024)` shows a well-centered PSF, whereas values of (0,0), (0,2047), (2047,0), and (2047,2047) all show PSFs shifted diagonally by about 0.47". However, SWA and SWB consist of four detectors with the center of the field situated at their four adjoining corners. I would have expected that center of the field to be the preferential position where the PSF is "unshifted." This appears to be a consequence of the fact that the `apply_distortion` function uses the detector-specific aperture (e.g., `NRCA1_FULL`), so it's always using the center of the detector as the relative position to determine PSF shifting.

mperrin:
@shanosborne are you comfortable that this issue has been addressed by your PR #229? If you agree let me know and we can close this issue.

shanosborne:
@mperrin Yes - I am comfortable that #229 will fix this issue (once the PR has been finalized- I still have that 1 mean vs median question)

mperrin:
Fixed by #229

PR #229, opened 2018-08-30
Shift distorted PSFs back to correct detector position

shanosborne:
When distorting a PSF, now the indices we interpolate from and on to have (approximately) the same center point. This fixes issue raised in #226

### DIFF
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -119,6 +119,10 @@ def apply_distortion(hdulist_or_filename=None, fill_value=0):
     xnew += xpix_center
     ynew += ypix_center
 
+    # 5) Shift the new indices so the center matches the sci indices
+    xnew += np.mean(xsci) - np.mean(xnew)
+    ynew += np.mean(ysci) - np.mean(ynew)
+
     # ###############################################
     # Interpolate from the original indices (xsci, ysci) on to new indices (xnew, ynew)
     # noinspection PyTypeChecker
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -101,6 +101,10 @@ def apply_distortion(hdulist_or_filename=None, fill_value=0):
     # Going from Idl to Sci this way allows us to add in the distortion
     xsci, ysci = aper.idl_to_sci(xidl, yidl)
 
+    # 6) Shift the sci indices so they match the PSF's position again (moved slightly off from pysiaf calculation)
+    xsci += xpix_center - np.median(xsci)
+    ysci += ypix_center - np.median(ysci)
+
     # ###############################################
     # Create an array of indices (in pixels) that the final data will be interpolated on to
     # 1) Set up blank indices (in pixels)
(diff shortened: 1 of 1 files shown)


shanosborne:
Tables show for different detector positions, the original offset between the center of the sci and new indices (the indices we interpolate from and to respectively). The 3rd column is the value we get from the average x/y new minus the average x/y sci value. And the last column is the new offset between the center of the sci and new indices if we subtract the value in the 3rd column from the new indices.

image

image

And this is the result: These show the same NIRCam PSF at detector position (0,0) before and after this change

Before:
0_0_before

After:
0_0_after

mperrin:
This looks good. Can you briefly explain / summarize the changes in algorithm you made to implement this fix?

mperrin:
We should also check with @Johannes-Sahlmann about what level of precision is appropriate here. The residuals in the table above are a few parts in 1000 of a pixel, which is indeed quite small and I expect is good enough for all current or near future purposes. Yes? Eventually I'm sure someone will be doing millipixel astrometry with JWST but not soon...

shanosborne:
So I decided on a better/more logical way to fix this issue. Basically what happens is when the sci indices are run through the idl_to_sci transformation, the output indices are no longer exactly centered on the initially set PSF location. So what this fix does is move the sci indices back to the initial PSF location after they’ve been run through the pysiaf transformation. And I do this with the lines:

# 6) Shift the sci indices so they match the PSF's position again (moved slightly off from pysiaf calculation)
xsci += xpix_center - np.median(xsci)
ysci += ypix_center - np.median(ysci)

So now the center point of the sci indices matches the center point of the new indices, and you get the results below (this is for the (0,0) case):

nircam_before

nircam_after

mperrin:
That makes sense.

Minor question, would it make any practical difference to use mean instead of median? It seems to me that the intent of this is something along the lines of 
```
xsci += xpix_center - center_value_of_xsci
```
(i.e. force the center pixel value to be the same between xsci and xpix). In practice the median value is probably very close to the mean, and to the value at the exact center of the array, so probably I'm over-thinking this. 


The Travis build appears to be failing because it's trying to run test cases on Python 2.7 still. If you git merge master back into your branch, or rebase on current master , then you ought to be able to get the updated .travis.yml file so that it only tries running the tests on Python 3.

shanosborne:
Yes, `xsci += xpix_center - center_value_of_xsci` is exactly what I'm trying to do. I was also thinking about median vs mean. The difference is small (in the case of the NIRCam F200W (0,0) PSF, the difference between the mean and median for the sci indices is around 0.01. And there is no difference between the mean and median for the new indices). I thought median might be correct since we're dealing with integer pixel indices, but maybe since that idl_to_sci transformation is applied, this is the wrong way to think about it.

Also - my .travis.yml file is the same as the version in the spacetelescope master. Is it possible that that change is only in mperrin/webbpsf?

shanosborne:
So I wanted to check back in with this - should I use the mean to find the center location of the PSF in this case? Also - do you recognize this travis failure?

mperrin:
As discussed in email and in person - the current implementation with median is accurate to about 0.01-0.02 pixels. That’s good enough for almost all purposes. So we’re going to go ahead and merge this now, and if more carefully handling of the centering becomes necessary in the future we’ll fix it up more precisely then.

Issue #230, opened 2018-09-04
Update OPD sub-classes to include functional as_fits method


laurenmarietta:
Re-implement the as_fits method in the OPD class in `opds.py` to actually work with the given subclasses.

mperrin:
Fixed by #232

PR #232, opened 2018-09-14
Fix as_fits() in opds.py

laurenmarietta:
From #230:
- Fixes the `as_fits()` method within the `OPD` class of `opds.py` to properly return an OPD as an `HDUList`.
- Updates the filename to the pupil OPD (was "jwst_pupil_revW_npix1024.fits.gz", now jwst_pupil_RevW_npix1024.fits.gz").
- Fixes a couple typos and some PEP8 violations.

A question: in order for an OPD HDU object to be used to calculate a PSF, there needs to be a header keyword for the units. Here I have hardcoded `header['BUNIT'] = 'meter'`, but I'm not sure the unit will always be meters...?

Also, a general comment: I am starting to play around with PyCharm's inspections feature, and there seem to be a lot of broken pieces in `opds.py` (namely, 48 apparently unresolved references). Just making sure someone is aware of this!

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -38,7 +38,6 @@ import astropy.table
 import astropy.io.fits as fits
 import astropy.units as u
 import logging
-import six
 
 import poppy
 import poppy.zernike as zernike
@@ -62,6 +61,7 @@ because of small differences in the parameter ranges used when
 deriving the linear approximations.
 
 """
+__location__ = os.path.dirname(os.path.abspath(__file__)) + os.sep
 
 
 ################################################################################
@@ -107,7 +107,6 @@ class OPD(poppy.FITSOpticalElement):
             slice of a datacube to load OPD from, if the selected extension contains a datacube.
 
         """
-        mypath = os.path.dirname(os.path.abspath(__file__)) + os.sep
         if opd is None and transmission is None:
             _log.debug('Neither a pupil mask nor OPD were specified. Using the default JWST pupil.')
             transmission = os.path.join(utils.get_webbpsf_data_path(), "jwst_pupil_RevW_npix1024.fits.gz")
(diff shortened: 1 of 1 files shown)


mperrin:
Surprised to hear about the unresolved references in `opds.py`. I don't see those when running the Pycharm inspector on it here on my side. In which functions are you seeing those? Let's take a look together sometime.

webbpsf/opds.py
```diff
@@ -186,22 +185,22 @@ def __sub__(self, x):
...
         -----------
         include_pupil : bool
             Include the pupil mask as a FITS extension?
         """
 
-        output = fits.HDUList([self._opdHDU])
-        output[0].header.update('EXTNAME', 'OPD')
+        output = fits.HDUList([fits.ImageHDU(self.opd)])
+        output[0].header['EXTNAME'] = 'OPD'
+        output[0].header['BUNIT'] = 'meter'  # Will this always be the case?
```
mperrin: > Will this always be the case? 

I think so. `poppy` internally works with all OPDs in SI units, i.e. meters. It is capable of reading in FITS files with OPDS in other units, like microns or nanometers, but it immediately converts the arrays into meters. (That's all inside `poppy.poppy_core.FITSOpticalElement.__init__`). For output, I don't see any need right now for a feature that would let us output in other units, so I think this is good as-is.

mperrin:
The failed Travis build is due to an unrelated problem, in which the Sphinx docs aren't building on Travis. Not a problem with this PR. 

The "failed" Coveralls build is an artifact of splitting up the same code over more lines, so the line counts change of how much test coverage there is. It's not an actual real decrease in test coverage. 

So despite those two failed tests, this PR looks good and I'm going to go ahead and merge it.

PR #236, opened 2018-09-18
Removed line because no longer needed and messes up installation.

### LINKED ISSUE #234: Set use_2to3 to False in setup.py?
@kjbrooks identified a problem when installing `webbpsf` in developer mode (i.e. either by `pip install -e .` or by `python setup.py develop`) where the package was being built and then run from  a `build/` directory, rather than the source code.

The primary effect of this is that the installation isn't really edittable; if a file in `webbpsf/` is changed, that change would not be registered in the package unless the install command was run again and the files in the `build/` directory were updated.

It seems that this behavior is coming from the `use_2to3=True` definition in `setup.py` (line 131). From what I can glean from the [setuptools documentation](https://setuptools.readthedocs.io/en/latest/python3.html) and [StackOverflow](https://stackoverflow.com/questions/35734519/why-after-setup-py-develop-new-submodules-are-not-available/35735171#35735171), it seems as though this option assumes that the source code is written in Python 2, and the code in the `build/` directory is thus the result of `setuptools` converting the code into Python 3.

Since `webbpsf` is written in Python 3 (right?), it seems like it would be beneficial to remove the `use_2to3` option, as it is unnecessary and makes developing more complicated.

kjbrooks:
Fixes #234

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -128,6 +128,5 @@ setup(name=PACKAGENAME,
       long_description=LONG_DESCRIPTION,
       cmdclass=cmdclassd,
       zip_safe=False,
-      use_2to3=True,
       **package_info
 )
--- a/webbpsf/tests/test_webbpsf.py
+++ b/webbpsf/tests/test_webbpsf.py
@@ -191,22 +191,6 @@ def test_return_intermediates():
     assert isinstance(intermediates[0], poppy.Wavefront)
     assert isinstance(psf, astropy.io.fits.HDUList)
 
-def test_unicode_filter_names():
-    """ See https://github.com/mperrin/webbpsf/issues/18
-    Bug reported by Brian York in which unicode filternames made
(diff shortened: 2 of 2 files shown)


mperrin:
oh ok, this really is that simple. Just a one line edit.

mperrin:
Huh, the 2to3 function was actually doing something. The `tests.test_webbpsf.test_unicode_filter_names` function uses a unicode function which is a Python 2 idiom, and doesn't work in Python 3 (since all strings are now unicode automatically in Py3). So this makes the tests all fail without 2to3. 

I think it's probably fine to just throw away that test case - it's a check against a long-since-fixed problem, and it's not relevant anymore in Py3 given how strings and unicode work in 3. So I think we should just get rid of it. Can you add  to this PR a commit deleting that function, please? Let's see if that can help us get the tests not all failing.

kjbrooks:
Yes, I can do that.

PR #238, opened 2018-09-20
Bug Fix: Documentation build failing on Travis

### LINKED ISSUE #233: Documentation build failing on Travis
The docs build for WebbPSF is failing consistently right now. @robelgeda could you take a look at this please, and see if you can figure out what’s going on? Thanks

robelgeda:
Closes #233 
This PR resolves the Travis doc build failure. It turns out the failure is caused by duplicate code that defines `calc_psf.__doc__` in `webbpsf_core.JWInstrument` :

[First instance (L790)](https://github.com/spacetelescope/webbpsf/blob/master/webbpsf/webbpsf_core.py#L790-L794)
[Second instance (L852)](https://github.com/spacetelescope/webbpsf/blob/master/webbpsf/webbpsf_core.py#L852-L856)

The duplicate was introduced in this [merge with master](https://github.com/spacetelescope/webbpsf/commit/6c7fc717406e6797883ea79b8b6f0117263e0448#diff-1340a819c5a9f36ab275cd31b8866138R791).

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -849,12 +849,6 @@ class JWInstrument(SpaceTelescopeInstrument):
         # Rewrite result variable based on output_mode set:
         SpaceTelescopeInstrument._calc_psf_format_output(self, result, options)
 
-    # Allow users to see poppy calc_psf docstring too
-    ind0 = calc_psf.__doc__.index("add_distortion")  # pull the new parameters
-    ind1 = SpaceTelescopeInstrument.calc_psf.__doc__.index("Returns")  # pull where the parameters list ends
-    calc_psf.__doc__ = SpaceTelescopeInstrument.calc_psf.__doc__[0:ind1] + calc_psf.__doc__[ind0:] + \
-                       SpaceTelescopeInstrument.calc_psf.__doc__[ind1:]
-
     def interpolate_was_opd(self, array, newdim):
         """ Interpolates an input 2D  array to any given size.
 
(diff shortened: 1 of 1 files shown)


mperrin:
OK, this ought to be coordinated with #227, since that's also changing the code about this doc string. It looks like you're just stripping out that part of the code, is that right?  While the PR #227 moves the same code to a decorator to work better that way.  Attention @shanosborne can you discuss this with @robelgeda, and make sure that your PR #227 will plan well with the documentation build system? Hopefully so.

mperrin:
Meanwhile I'm confused why there isn't any Travis build results showing up on this PR here. Because it did run on Travis: https://travis-ci.org/spacetelescope/webbpsf/builds/430961595

That build failed, but I think from unrelated reasons, that the Python 3.4 test case doesn't have the right packaged available. I have a fixed version of that locally which updates things to minimum Python 3.5. Stand by and I will push that momentarily

shanosborne:
@mperrin I don't know how that doc string code got there twice, but I've already removed them in #227. Also the PR is failing Travis due to the package issue you mentioned, not the documentation, so I think #227 should be fine

robelgeda:
@mperrin @shanosborne : 

> It looks like you're just stripping out that part of the code, is that right? 

This is right. I just removed the duplicate. #227 will fix this issue without this PR since it does the same. This can be closed if that works.

robelgeda:
@mperrin : This PR is now outdated and will remove necessary code.

PR #239, opened 2018-09-20
Enforce minimum Python > 3.5 and remove Python 2.7 compatibility bits

mperrin:
- Remove use of `six` and `from __future__ import x` for Python 2.7/3+ cross compatibility. 
- Enforce minimum Python 3.5 at both setup time and run time, following how this is done in the current astropy package template
 - Update Travis config to use Python 3.5 instead of 3.4 as the minimum version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,10 +49,10 @@ matrix:
         - python: 3.6
           env: SETUP_CMD='test' ASTROPY_VERSION=development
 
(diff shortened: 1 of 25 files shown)


mperrin:
OK, the Travis build for this passes for most cases, except the docs build which is the subject of the #227 and #238 PRs

PR #241, opened 2018-10-18
Adding ability to create a grid of PSFs for Photutils

shanosborne:
Adding ability to create a grid of PSFs expected to be used with Photutils for PSF-fitting photometry.

Things still to be determined:
1. What is the best practice for naming the files?
2. Should outputing a detector sampled grid of PSFs be an option?
     - And as part of this - should we only allow certain setups (e.g. add_distortion=True and must be oversampled) or give users the option to change these?
3. What should the output object be? (hdulist, nddata, photutils object)
4. What should the default values be? (see psf_grid() docstring for all current defaults)

(Note that there are some aspects that I will clean up once everyone is okay with the functionality and it's determined that they are no longer needed - E.g. the notebook will be cleaned up and some edge case fixes will be removed from the class if its determined we don't need them.)


webbpsf/tests/test_psfgrid.py
```diff
@@ -0,0 +1,195 @@
...
+from astropy.io import fits
+import numpy as np
+import os
+import pytest
+
+from .. import gridded_library
+from .. import webbpsf_core
+
+
+# @pytest.mark.skip()
```
mperrin: These tests look good. Do we still need the commented-out `# @pytest.mark.skip()` lines?

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
 
         return fits.HDUList(fits.ImageHDU(newopd, header=hdr))
 
+    def psf_grid(self, all_detectors=True, num_psfs=16, single_psf_centered=True, use_detsampled_psf=False,
+                 save=False, outfile=None, overwrite=True, **kwargs):
+        """
+        Create a PSF library in the form of a grid of PSFs across the detector based on the specified instrument,
+        filter, and detector. The output file is of the form [i, y, x] where i is the PSF position on the detector
```
mperrin:
```suggestion
        filter, and detector. The output file will contain a 3D array with axes [i, y, x] where i is the PSF position on the detector
```

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+        grid and (y,x) is the 2D PSF.
+
+        Parameters
+        ----------
+        all_detectors: bool
```
mperrin: minor formatting quirk: Annoyingly, the numpy docstring standard specifies that there must be a space _before_ the colon. Like so: 
```suggestion
        all_detectors : bool
```
and likewise for all the other parameters.

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+        -------
+        grid : fits.HDUList or list
+            Returns and/or saves 3D fits.HDUlist object (1 per instrument, detector, and filter)
+            or a list of fits.HDUlist objects if more than one detector is specified
+
+        Use
+        ----
+        nir = webbpsf.NIRCam()
+        nir.filter = "F090W"
+        nir.psf_grid(all_detectors=True, num_psfs=4)
```
mperrin:
```suggestion
        grid = nir.psf_grid(all_detectors=True, num_psfs=4)
```

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+        # Keywords that could be set before the method call
+        filters = self.filter
+
+        if all_detectors is True:
+            detectors = "all"
+        else:
+            detectors = self.detector
+
+        if single_psf_centered is True:
+            psf_location = (int(self._detector_npixels / 2), int(self._detector_npixels / 2))  # center point
```
mperrin: In fact the center location has to be computed as `(npixels-1)/2`. (You can think about the case of npixels=2 and npixels=3, and you should be able to convince yourself why this is the case, for Python's 0-based array indexing...)

```suggestion
            psf_location = (int((self._detector_npixels-1) / 2), int((self._detector_npixels-1) / 2))  # center point
```

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+        else:
+            psf_location = (self.detector_position[1], self.detector_position[0])  # (y,x)
```
mperrin:
```suggestion
            psf_location = self.detector_position
```

shanosborne: I’m not sure I agree. When the detector_position attribute is used to set the header keywords DET_X and DET_Y, we see that detector_position is (x,y) whereas the psf_location variable here is (y,x)

mperrin: Aha yes, ok, right. Good catch.  
In that case maybe `psf_location = self.detector_position[::-1]`?  That succinctly says "use the values from this other variable, but flip the order around".

mperrin:
Some general feedback. This mostly is just restating what I said on the most recent telecon. 

I like the idea of having this output a Photutils object of some kind, or something that can easily be turned into a Photutils object. But correct me if I am wrong, there's not yet a native class in Photutils to represent a grid of PSFs, right? In some sense I think the best approach in the long run may be to co-develop functionality in both classes, such that Photutils gets some class for e.g. `EPSFModelGrid` and then the `psf_grid` function here can return one of those. I think most of the functionality for `EPSFModelGrid` already exists in what's in MIRAGE, yes? 

In terms of defaults and options, I think we should focus on the use case in MIRAGE, which also happens to be the appropriate case for fitting PSFs to real data from individual exposures. The purpose of `psf_grid` should be to prepare ePSFs which are suitable for use on level 2b image data, which means: 
 * ePSFs must be at least Nyquist sampled at the shortest wavelength of interest in the filter bandpass. (note that "must be at least Nyquist sampled" is not the same as "must be oversampled"; if we want to be strict about it then this depends on the instrument and selected filter)
 * The optical distortion should be included. 
 * Other detector effects that modify the PSF such as the MIRI scattering should be included, unless the pipeline takes them out, or unless they are already modeled elsewhere. In particular IPC should _not_ be included, because there is a separate addIPC step in MIRAGE. (yes?) And there is a data pipeline step to remove IPC (though am I right that currently that step is not enabled by default?)

hcferguson:
I think it would be useful to get some advice from @drdavella on how this might be packaged in ASDF. First, the logical layout of the files might be much cleaner, and second, we should be eating our own dog food. If we are going to be turning these into photutils objects then serializing them that way in asdf from the start might make things more seamless overall, but @larrybradley should advise on what the structure is.

drdavella:
@larrybradley I am happy to help with implementing ASDF extensions for `photutils` if that seems like a good path forward.

larrybradley:
@drdavella You can find a WIP version of `GriddedPSFModel` on my branch here:
https://github.com/larrybradley/photutils/blob/griddedpsfmodel/photutils/psf/models.py#L741

It's essentially an `astropy` `Fittable2DModel` that is initialized by a `NDData` object.  The `NDData` object has a `data` attribute containing a 3D `ndarray` and a `meta` attribute containing a `dict` (whose values are `str`, `int`, or a list of `(x, y)` positions).

Currently, I think @shanosborne (correct me if I'm wrong) is writing a FITS file containing the 3D `ndarray` with the metadata stored in the FITS header.

shanosborne:
@larrybradley Yes it currently writes to a FITS object - but once there is a consensus about what the output object should be I can change this.

mperrin:
@larrybradley How will `GriddedPSFModel` handle instruments with multiple detectors? (NIRCam, NIRSpec, ACS, WFC3/UVIS, WFIRST WFI, etc...) Are you thinking "one GriddedPSFModel per detector", an enhanced GriddedPSFModel class that tracks more lookup information to span across multiple detectors, or something new like a GriddedPSFModelSet class to do that? Or is this yet to be determined? 

Not that we necessarily have to have the answer to that right away, but it may inform design decisions about what the `psf_grid` function in WebbPSF returns in this case.

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+    def psf_grid(self, all_detectors=True, num_psfs=16, single_psf_centered=True, use_detsampled_psf=False,
+                 save=False, outfile=None, overwrite=True, **kwargs):
+        """
+        Create a PSF library in the form of a grid of PSFs across the detector based on the specified instrument,
+        filter, and detector. The output file will contain a 3D array with axes [i, y, x] where i is the PSF position
+        on the detector grid and (y,x) is the 2D PSF.
+
+        Parameters
+        ----------
+        all_detectors : bool
```
eteq: I find this argument confusing.  I'd say either have the argument be the list of detectors (or something like that), or (and this is my preference) have the information come entirely from the instance.  

I'm not clear if ``"all"`` is an option on the underlying class, though?  And related, does it do anything outside of this method? If not, that argues for a method keyword.

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+            If True, run all detectors for the instrument. If False, run for the detector set in
+            the instance. Default is True
+        use_detsampled_psf : bool
```
eteq: This is tied back to that unresolved question in the telecon... But I would think this is different enough *conceptual* behavior that instead of being a keyword it should be two methods, one called, say, `epsf_grid`, and the other `psf_grid`.  They could still share most of their functionality in an internal private (leading `_`) method, but it would reinforce to the user that these are conceptually very different objects.

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+            If True, the grid of PSFs returned will be detector sampled (made by binning down the
+            oversampled PSF). If False, the PSFs will be oversampled by the factor defined by the
+            oversample/detector_oversample/fft_oversample keywords. Default is False.
+        num_psfs : int
```
eteq: Maybe this should instead/additionally be a 2-tuple?  E.g. `(4,4)` by default. Or alternatively/additionally, a list of positions in detector coordinates...

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+            The total number of fiducial PSFs to be created and saved in the files. This
+            number must be a square number. Default is 16.
+            E.g. num_psfs = 16 will create a 4x4 grid of fiducial PSFs.
+        single_psf_centered : bool
+            If num_psfs is set to 1, this defines where that psf is located. If True it will be the
+            center of the detector, if False it will be the location defined in the WebbPSF
+            attribute detector_position (reminder - detector_position is (x,y)). Default is True
+        save : bool
+            True/False boolean if you want to save your file. Default is False.
+        outfile : str
```
eteq: Instead of `save` + `outfile` I'd suggest a single keyword where `None` means "don't save", and otherwise the user has to give the filename.  If you still want the default behavior, you could use standard python string formatting and tell the user to do `'instr_det_filt_fovp{}_samp{}_npsf{}.fits'` or similar.
 
(Caveat: the comment I'm about to leave in the "general" thread might make both of these go away.)

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+            If "save" keyword is set to True, your current file will be saved under
+            "{outfile}_det_filt.fits". Default of None will save it in the current
+            directory as: instr_det_filt_fovp#_samp#_npsf#.fits
+        overwrite : bool
+            True/False boolean to overwrite the output file if it already exists. Default
+            is True.
+        **kwargs
+            Add any extra arguments to the WebbPSF calc_psf() method call. This includes
```
eteq: Opinions vary here... but in my experience it's dangerous to enumerate kwargs like this.  Inevitably someone will update `calc_psf` down the line and not know/forget that they need to update this method's docstring.  So I would just leave only the first sentence here, but as 
```
Add any extra arguments to the WebbPSF :meth:`calc_psf` method call
```
so that the html docs get a hyperlink.  That at least makes it easy to go look it up in the web docs.

webbpsf/webbpsf_core.py
```diff
@@ -920,6 +921,87 @@ def load_was_opd(self, inputWasOpd, size=1024, save=False, filename='new_was_opd
...
+            oversampled PSF). If False, the PSFs will be oversampled by the factor defined by the
+            oversample/detector_oversample/fft_oversample keywords. Default is False.
+        num_psfs : int
+            The total number of fiducial PSFs to be created and saved in the files. This
+            number must be a square number. Default is 16.
+            E.g. num_psfs = 16 will create a 4x4 grid of fiducial PSFs.
+        single_psf_centered : bool
+            If num_psfs is set to 1, this defines where that psf is located. If True it will be the
+            center of the detector, if False it will be the location defined in the WebbPSF
+            attribute detector_position (reminder - detector_position is (x,y)). Default is True
```
eteq: Similar to above, what about having `detector_position` be the One True Way of indicating this?  I.e., let `detector_position` be `'center'` or similar and have that be the same as the `bool` being True.

And what happens if `num_psfs` is greater than 1?  I would think you still need the centering information.  (Although if you adopt my suggestion above for `num_psfs` being a list of psf locations, maybe that makes this superfluous.)

eteq:
Some more high-level items:

On the output question: I like the idea of `GriddedPSFModel` being the output. I think it probably needs a bit of "metadata" (which can just be implemented as attributes that `webbpsf` attaches to the object) - certainly the detector and instrument names, but possibly some other useful information (but don't worry too much about this as it can always be added to later).

 To take it one step further: perhaps the method under discussion her in `webbpsf` should *only* return that, and not do *any* saving?  Either `photutils` or `webbpsf` could instead implement a `save_psf_library` function that takes the output of this method (or a for loop of multiple detectors) and actually does the saving, and possibly a matched `load_psf_library`.  Then users who need the "saved library" functionality can do that, possibly even with their own tweaked versions of these PSFs, but those who are happy with the in-memory-only version can totally ignore the saving.

One caveat/additional is: do we need to have both `GriddedPSFModel` and `GriddedEPSFModel`? (or similar - those names are dangerously close, but the idea here is "thing with the Pixel Response Function convolved" and "without")  They would seem to be necessary to differentiate between to then actually use with PSF photometry.  Not clear to me if there implementation differences here, but definitely *conceptual* differences.

@mperrin:
> one GriddedPSFModel per detector

I'm strongly in favor of this. The data structure here is already mind-bendingly complicated, and "detector" seems like a natural place to break it apart for the user to more easily mentally consume.

hcferguson:
I don't mind a separate step to save to disk, but if this is going to be an object anyway, I would suggest just adding a `writeto()` method, since that seems like a fairly common paradigm.

mperrin:
A counter argument on “one per detector”. Well, two related counter arguments. (1) That’s not the common practice on HST.  Jay’s ePSF libraries for ACS and WFC3/UVIS package up the PSFs from both CCDs together for each instrument. (2)  Consider the common use case of trying to fit a point source in a set of dithered image observations.  There’s no guarantee the point source will always fall in the same detector of e.g. ACS or NIRCam in all of the dithered exposures.  In fact it probably won’t.  So if each model only spans one detector, then there is a strong need for a higher level construct that would enable the fitting over the whole set of detectors. 

It seems to me the most useful level of abstraction to present to the end user is an instrument mode. MIRI Imaging mode has 1 detector, NIRCam long wave imaging has 2, NIRCam short has 8, WFI will have 18. Regardless, the common task will be “fit the astrometry or photometry of these sources in this list of exposures, automatically handling the details of whichever detector the point source is in for each exposure”. Yes?

Doing it that way would certainly be more complicated for us behind the scenes, no question about that.  But I think could be usefully easier for users in the long run.

larrybradley:
@mperrin I agree with you in principle, but there aren't any tools currently in `photutils` to perform combined photometry over multiple images (i.e. dithers and/or detectors).  For now, `GriddedPSFModel` should be for a single detector since that's how the photometry tools currently work (one model for one image).  We can work on the more complex case after Build 5.

shanosborne:
@larrybradley - I need to add photutils as an installation requirement, what version are you about to release (that will contain the GriddedPSFModel)?

mperrin:
@shanosborne Alternatively we could set it up so that photutils is an optional import rather than a requirement. I.e. a user who doesn't have photutils installed would only have an error if they tried to run psf_grid, but otherwise would be ok.  I wonder if that would be better to minimize user support needs? Or if it's safe to assume that most people are installing from conda these days so adding a dependency on photutils is not a big deal.

larrybradley:
I would recommend making `photutils` an optional requirement since that one method is the only functionality in `webbpsf` that requires it.  Simply move the `photutils` import to inside the method.  You can put the import in a `try... except ImportError` block to give a nicer error message (e.g. "this method requires photutils 0.6").

larrybradley:
@shanosborne The next release of `photutils` will be v0.6.

shanosborne:
Yes it's in photutils, my guess is that's what has been causing travis to fail (all the errors seemed to be cython-related)

shanosborne:
@mperrin - There was an issue with the `SETUP_CMD = ‘test --coverage’` run which resulted in `RecursionError: maximum recursion depth exceeded while calling a Python object`. With some help from Robel, we found that this is somewhat of a known problem (Joseph even submitted an issue about it) and downgrading to `sphinx=1.5.6` fixed the issue for most people. I downgraded sphinx here to check if it would fix the test, and it appears to have done so, but this doesn’t seem like a great long-term solution. Do you have any thoughts?

mperrin:
Nope, that's beyond my expertise! I'm confused, that error was with the coverage test, not the docs build test? Why does sphinx even matter for that? Weird. Probably going to have to call in the real experts on this one, on the Python slack channel or by asking around in DATB or elsewhere...

shanosborne:
So we're thinking that sphinx is a dependency of a dependency somewhere in the coverage test process, and it's installing the wrong version. Actually let me try something that might be a better method and see if it works

shanosborne:
Okay, so none of our attempts worked (basically just trying the newer versions of sphinx). I've reverted back to the method that was working - this may be a problem for another time. Also it looks like the sphinx dependency is coming from astropy_helpers which is called in setup.py

mperrin:
Thanks for all your work on this one @shanosborne. 

Sounds good - we'll proceed with the current sphinx problem workaround and can sort it out better later.

PR #242, opened 2018-11-02
PEP8 + Docstring updates to distortion code

shanosborne:
I added/cleaned up a bunch of doctrings and comments in distortion.py and where distortion.py is called in webbpsf_core.py. I also removed some imports that were no longer being used. 

There are no changes to the code or how it runs - only changes to docstrings, comments, and imports

### DIFF
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -1,4 +1,3 @@
-
 import copy
 
 import astropy.convolution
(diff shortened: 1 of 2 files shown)


mperrin:
This all looks good and I think is good to merge. Are you planning to add any more commits or should I go ahead and merge? 

Semi-relatedly: reading this code got me thinking about the rotation and distortion steps for NIRCam, and whether I understand the order those are applied in. Right now it's rotation first then distortion second. That makes sense to me (I think), but I wanted to double check that also makes sense to you & Johannes as the correct order.

shanosborne:
Yes I'm done committing things, so you can merge it whenever. 

I mean, it definitely wouldn't hurt to double check the order of application. I don't know exactly how the SIAF measured its distortion values, but if they only come from the detector and not any of the mirrors (which I think they do), then wouldn't the rotation of the detector not matter since the coefficients are found in the reference frame of the detector? I know I did find a difference in the PSF outputs based on the order of operation, but I'm wondering if that came from the imperfect method of using interpolation to rotate the array. What was your thought process for why rotation needs to come first?

PR #250, opened 2018-12-05
 STScI doc theme

robelgeda:
This PR introduces the STScI doc theme. It also fixes some minor doc-string errors. 
screen shot 2018-12-05 at 5 54 16 am

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,8 @@ matrix:
         # Check for Sphinx doc build errors
         - python: 3.6
           env: SETUP_CMD='build_sphinx'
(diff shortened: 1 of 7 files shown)


robelgeda:
You can test this locally on your Mac Via:
`python setup.py build_docs`
`open build/sphinx/html/index.html`

mperrin:
This looks good, thanks. 

I wonder if some of the dependencies could be trimmed out - do we still need six, now that this codebase is Python 3 only? - but we can leave that as a separate issue for later. I’m going to go ahead and merge this now.

PR #256, opened 2018-12-10
Add unit test for defocus and weak lens. Fixes #59

### LINKED ISSUE #59: Add test for optional defocus in calculations
<a href="https://github.com/josePhoenix"></img></a> **Issue by [josePhoenix](https://github.com/josePhoenix)**
_Tuesday Mar 03, 2015 at 21:51 GMT_
_Originally opened as https://github.com/mperrin/webbpsf/issues/59_

----

The code to introduce defocus in PSF calculations through the `options` dictionary was not being exercised by our test suite, resulting in a typo slipping through.

mperrin:
This PR fixes the long-standing issue (#59) of there not being a test case for any of the defocus options. 
No changes in package functionality; this just adds a test.

### DIFF
--- a/webbpsf/tests/test_nircam.py
+++ b/webbpsf/tests/test_nircam.py
@@ -311,6 +311,7 @@ def test_defocus(fov_arcsec=1, display=False):
     assert np.allclose(psf[0].data, psf_2[0].data), "Defocused PSFs calculated two ways don't agree"
 
     if display:
+        import webbpsf
         plt.figure()
         webbpsf.display_psf(psf)
         plt.figure()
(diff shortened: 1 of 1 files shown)


mperrin:
Hmm, I thought this was going to work after the fixes to shift in the recent PR to poppy. 
https://github.com/spacetelescope/poppy/pull/274

I held off on this PR until after that. But let me recheck things and update the PR. I may not have pushed all the relevant changes from my local machine to github perhaps? Thanks for checking. 

And those values should only be subject to change if someone were to open up NIRCam and change out that lens inside it... so hopefully not!

mperrin:
Right, I completely missed also PRing the associated changes to make WebbPSF make use of the changes in the poppy PR 274. Stand by for that one in a separate PR

shanosborne:
I re-ran this test with the changes in #257 and it now passes in the default case. However, when `display=True`, it fails since `webbpsf` is not defined (only `webbpsf_core` is).

mperrin:
fixed the import. Thanks for catching that.

PR #262, opened 2018-12-13
Update make-minimal-datafiles.py

shanosborne:
File truncation had an error with flush()

I don't know why there are days-old, already-merged commits here. But the file changes only show the most recent commit


mperrin:
Rejecting this to try for a cleaner PR without the duplicate commits

PR #273, opened 2018-12-21
Add R062 Fix to relnotes.rst

robelgeda:
This adds a small note on the the R026 bandpass fix to the `docs/relnotes.rst`


mperrin:
Minor: it’s “R062” not “R026”, isn’t it?

robelgeda:
That is correct, thanks for catching that @mperrin !

PR #278, opened 2019-02-26
fix astropy.tests.pytest_plugins error

shanosborne:
Fix for astropy.tests.pytest_plugins problem- same fix as was in poppy. See discussion of issue here: https://github.com/spacetelescope/poppy/issues/289

### DIFF
--- a/webbpsf/conftest.py
+++ b/webbpsf/conftest.py
@@ -2,7 +2,11 @@
 # by importing them here in conftest.py they are discoverable by py.test
 # no matter how it is invoked within the source tree.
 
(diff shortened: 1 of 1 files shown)


mperrin:
This looks good to me of course. 

The Travis build appears to be failing because of an unrelated issue - I removed an old deprecated function call from `poppy`, and the WFI class needs to be updated to use the PEP8-compliant syntax for that. Should be a straightforward one-line fix for that.

shanosborne:
Okay, I can fix that if you want (unless you were planning to). What function was it?

Issue #246, opened 2018-12-04
SI WFE interpolation is suboptimal for SIs with few measurements from CV3


mperrin:
The interpolation of zernikes in WebbFieldDependentAberration uses `griddata` with cubic interpolation. This seems to work not very well for cases where we have few data points measured from the ISIM CV3 phase retrieval. The WFE model appears to "snap" from one to the other, rather than smoothly interpolating. Specific examples include the "right side" (more negative V2) of the MIRI FOV, and the "left side" of the NIRISS FOV, each of which only have 2 measurements in the corners. 

See attached where the PSF speckles are identical then change abruptly, rather than varying smoothly from place to place

![miri-psf-diff-example](https://user-images.githubusercontent.com/1151745/49470198-63407c80-f7d7-11e8-9a47-5db00dce378e.png)

![niriss-psf-diff-example](https://user-images.githubusercontent.com/1151745/49470310-a995db80-f7d7-11e8-89c2-56b8aa39c248.png)

Tried `method="linear"` but it did not work better. May need to investigate other methods of interpolation for these SIs.

JarronL:
Is this because those sampled field points are outside the region where the SI WFE is sampled? In that case, griddata returns a NaN, so the code simply selects the closest field point.

mperrin:
Hmm, maybe! That would make sense. Good suggestion for investigation, thanks.

JarronL:
Now that I've verified the Zemax models have good agreement to the CV3 measurements, I can provide you with a WFE measurements for a border of field points just outside the NIRCam FOV. So, for NIRCam at least, we will always be sampling within a defined grid region.

JarronL:
Here's some code that performs a linear extrapolation on regions outside the bounds of the measured field points. This would drop in around line 1115 of optics.py (inside `class WebbFieldDependentAberration(poppy.OpticalElement)`) to replace the line `cf = self.row[zkey]`. Rather than just picking the nearest neighbor, the code creates a fine-meshed cubic fit over the known field points, fixes any NaN's using `RegularGridInterpolator`, then again uses `RegularGridInterpolator` on the fixed data to extrapolate the requested field point. 

In principle, the first call of `RegularGridInterpolator` can be used to used to extrapolate the requested field point to eliminate the intermediate step, but I wanted to be sure to use all the real data rather than the trimmed data set.

```
from scipy.interpolate import griddata, RegularGridInterpolator

# Want to perform extrapolation if outside of bounds
[...]

Issue #115, opened 2018-08-28
Update various NIRCam coronagraph reference data


mperrin:
<a href="https://github.com/mperrin"><img src="https://avatars0.githubusercontent.com/u/1151745?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mperrin](https://github.com/mperrin)**
_Thursday Apr 28, 2016 at 05:04 GMT_
_Originally opened as https://github.com/mperrin/webbpsf/issues/115_

----

Per various emails from Stansberry:
- The ND squares are not actually truly neutral, they definitely have some color to them. We could add the appropriate relative spectral response. 
- From https://confluence.stsci.edu/pages/viewpage.action?pageId=52920601: "The wedge element in the SW pupil wheel has chromatic effects which impact PSF shape for wide-bandpass filters, so PSF subtraction will be particularly tricky unless target and PSF reference star are very similar in color.” That’s an effect that is totally neglected in WebbPSF right now.

mperrin:
<a href="https://github.com/mperrin"><img src="https://avatars0.githubusercontent.com/u/1151745?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mperrin](https://github.com/mperrin)**
_Friday Apr 29, 2016 at 18:13 GMT_

----

Coronagraph wedges info. 

From NIRCam Engineering Memorandum NIR-EM_0424 "Coronagraph Wedge Angle Tolerance Effect on Focal Plane Image Position" (provided by Stansberry):
- SW wedge is BaF_2, angle is 5 degrees
- LW wedge is fused silicon, angle is 0.5 degrees. 
  - pupil stop diameter is 42 mm 
  - note also the pupil wheel mechanical repeatability was estimated to result in up to 2 pixels of PSF shifts in the direction parallel to the axis of the bar occulter. But that was a circa-2007 estimate and is thus not the  most up to date numbers. 

From "NIRCam wedge dispersion trade_Si-BaF2-CaF2-Fused Silica.ppt" for the BaF2 prism

JarronL:
With respect to the wavelength dependence brought on by the Lyot wedges, would it be a process similar to the one we performed for the defocus Zernike component? That is, we would calculate how tip/tilt changes with wavelength using the Zemax data, then modify the OPD map using the tip/tilt offset values?

mperrin:
Yes, that would be the way to do it. Conceptually identical, except for modifying a different coefficient.

[...]

PR #283, opened 2019-03-14
NIRCam Zernike and focus updates

JarronL:
1. Include NIRCam coronagraphy Zernike field points `si_zernikes_coron_zemax.fits`
2. Provide extrapolation method for field points that are outside the bounds of Zernike table data.
3. Rework NIRCam focus model such that F323N is a special case.
4. Use polynomial coefficients fit to focus model instead of interpolation of table data, providing seamless extrapolation.


webbpsf/optics.py
```diff
@@ -1111,8 +1117,42 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
 
             # Cubic interpolation of of non-uniform 2D grid
             cf = griddata((v2, v3), zvals, (v2_tel, v3_tel), method='cubic').tolist()
+            # Want to perform extrapolation if outside of bounds
             if np.isnan(cf):
-                cf = self.row[zkey]
+                #cf = self.row[zkey]
```
mperrin: can delete this leftover commented-outline

webbpsf/optics.py
```diff
@@ -1111,8 +1117,42 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
...
+                xgrid = np.arange(v2.min(), v2.max()+dstep, dstep)
+                ygrid = np.arange(v3.min(), v3.max()+dstep, dstep)
+                X, Y = np.meshgrid(xgrid,ygrid)
+
+                # Cubic interpolation of all points
+                zgrid = griddata((v2, v3), zvals, (X, Y), method='cubic')
+    
+                # There will be some NaNs along the border that need to be replaced
+                ind_nan = np.isnan(zgrid)
+                # Remove rows/cols 1 by 1 until no NaNs
```
mperrin: This part seems kind of ad hoc. In particular it doesn't account for any significantly non-square border (i.e. you'd ideally exclude different numbers of lines in X and Y). That may not matter in practice sufficiently to do anything different - I'm just asking.

webbpsf/optics.py
```diff
@@ -1111,8 +1117,42 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
 
             # Cubic interpolation of of non-uniform 2D grid
             cf = griddata((v2, v3), zvals, (v2_tel, v3_tel), method='cubic').tolist()
+            # Want to perform extrapolation if outside of bounds
             if np.isnan(cf):
-                cf = self.row[zkey]
+                #cf = self.row[zkey]
+
```
mperrin: Perhaps add here this explanatory text, adapted from your comment on #246:

```
    # To extrapolate outside the measured field points, we proceed in two steps. 
    # This first creates a fine-meshed cubic fit over the known field points, 
    # fixes any NaN's using RegularGridInterpolator, then again uses 
    # RegularGridInterpolator on the fixed data to extrapolate the requested field point.

    # In principle, the first call of RegularGridInterpolator can be used to extrapolate 
    # the requested field point to eliminate the intermediate step, but this method
    # enables use of all the real data rather than the trimmed data set.
```

webbpsf/optics.py
```diff
@@ -1243,37 +1283,15 @@ def __init__(self, instrument, **kwargs):
...
-
-        long_wavelengths_mask = model_wavelengths > 2.45
-        # (n.b. row where model_wavelengths == 2.45 is nan)
-        self.fm_long = scipy.interpolate.interp1d(
-            model_wavelengths[long_wavelengths_mask],
-            model_defocus[long_wavelengths_mask],
-            kind='cubic'
-        )
-
+        # Polynomial equations fit to defocus model (RMS WFE).
```
mperrin: Can you add here please some explanatory comments about where these polynomial coefficient values come from?

webbpsf/optics.py
```diff
@@ -1297,24 +1314,13 @@ def get_opd(self, wave):
             opd_ref_wave = 3.23
             focusmodel = self.fm_long
 
-        try:
+        # F323N deviates from the model and has it's own focus power
```
mperrin: typo: it's -> its

webbpsf/optics.py
JarronL: I agree it's certainly not ideal, especially for MIRI and NIRSpec. I'm open to suggestions of a more optimal way to trim the rows/columns with NaNs (or a better way to perform the extrapolation). I've investigated a number of methods, and this was the best I could come up with for such sparsely samples data. For NIRCam and NIRISS WFE maps, it's fairly robust and quick. Below is an image showing the resulting extrapolation to the V2/V3 limits. I'm a little wary of NIRSpec and MIRI results beyond. Perhaps choosing the closest field point for those two instruments would be better?

si_wfe_extrap_Z4

mperrin: Excellent plot - that really helps visualize things. Agrees that NIRSpec is a bit of a mess with that method. 

One relatively easy thing to try for NIRSpec and MIRI would be to apply a rotation to the X and Y grids, so the grid axes are more closely aligned with the SI FOVs. That would make the mechanics of the trimming work better for those, I think. Then do all the extrapolation and replacement in that rotated grids.

webbpsf/optics.py
```diff
@@ -1111,8 +1117,42 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
...
+                zgrid2 = zgrid
+                ygrid2 = ygrid
+                xgrid2 = xgrid
+                while np.isnan(zgrid2.sum()):
+                    zgrid = zgrid2[1:-1,1:-1]
+                    ygrid = ygrid2[1:-1]
+                    xgrid = xgrid2[1:-1]
+
+                # Create regular grid interpolator function for extrapolation of NaN's
+                func = RegularGridInterpolator((ygrid2,xgrid2), zgrid2, method='linear',
```
shanosborne: `RegularGridInterpolator` is never imported

webbpsf/optics.py
```diff
@@ -1111,8 +1117,42 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
...
+                # Cubic interpolation of all points
+                zgrid = griddata((v2, v3), zvals, (X, Y), method='cubic')
+    
+                # There will be some NaNs along the border that need to be replaced
+                ind_nan = np.isnan(zgrid)
+                # Remove rows/cols 1 by 1 until no NaNs
+                zgrid2 = zgrid
+                ygrid2 = ygrid
+                xgrid2 = xgrid
+                while np.isnan(zgrid2.sum()):
```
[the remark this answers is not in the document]
JarronL: Thanks for checking this. Yes, I caught this bug last Thursday and have fixed it in my local distribution (should be `zgrid2 = zgrid2[1:-1,1:-1]`, etc.). I'm still working on Marshall's comments (but was out of town Fri-Mon) so have not yet updated the pull request. Hoping to implement everything by the end of the week.

webbpsf/optics.py
JarronL: Good catch. I thought I copied over all the code changes into the new branch, but apparently missed a couple things. My github pull request workflow is still a little awkward for contributed projects.

mperrin: @JarronL I'd be happy to chat sometime re pull requests and branches if you want. There's about a thousand possible workflows with git but I can tell you what I use (which is actually a couple different options, in different repos)

webbpsf/optics.py
```diff
@@ -1129,26 +1151,51 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
 
                 # Cubic interpolation of all points
                 zgrid = griddata((v2, v3), zvals, (X, Y), method='cubic')
-    
+
+                # Want to rotate zgrid image of some SIs to minimize NaN clipping
+                if 'NIRSpec' in lookup_name:
+                    rot_ang = 42
+                elif 'MIRI' in lookup_name:
+                    rot_ang = -25
```
mperrin: Where does a rotation angle of -25 deg for MIRI come from? The MIRI detector axes are only rotated ~ 4.5 deg relative to the V2V3 frame. Likewise I'm surprised to see a 5 deg rotation for NIRISS, which is less than a degree rotated relative to V2V3.

JarronL: It's more about the shape of the polygons of the measured field points rather than the detector orientation. These rotation angles maximize the number of remaining field points after trimming NaNs on the edges.

JarronL: For instance, have a look at the level of skewness for the MIRI WFE map:

image

JarronL: Ok, I decided to look into developing a more optimized NaN trimming algorithm. The new algorithm does a much better job keeping real data rather than over-trimming. Optimal rotation angles are much closer to the real instrument rotation relative to V2V3.

JarronL: Here's a final visualization. You'll probably want to disable SI WFE measurements for MIRI coronagraphy given there are no measurements in that part of the field, and RGI simply spits out artifacts.

Figure_6

mperrin: Thanks for continuing to work on improving this!  I'll take a look and test it a bit sometime when I'm not in the middle of a wavefront rehearsal in the MOC. 

Can you share your code or notebook for making these plots, please? Could be nice to include some version of these in the WebbPSF documentation. (That doesn't necessarily need to be in the repo - you could just email me that if it's easier, or you could PR it into the dev_utils subdir. However you prefer.)

JarronL: Uploaded a notebook for making the plots. Also included a couple minor improvements that reduces artifacts on extrapolated regions.

JarronL:
Ok, I think I've gone as far as I can go here. Travis was failing because `si_zernikes_coron_zemax.fits` doesn't yet exist in webbpsf data directory. I've commented that section out, so NIRCam coronagraphy is still using `si_zernikes_isim_cv3.fits` for now; I think that will fix the Travis build issues. I've also added another section (currently commented out) that will appropriately offset the desired V3 extrapolation region for coronagraphy.

mperrin:
@JarronL can you send me by email the file si_zernikes_coron_zemax.fits, please? I'll add it into the files used by the CI system.

mperrin:
I realized at least one additional change is needed in this PR related to the F323N filter's special defocus: we also need to take into account that the OPD model for all the Zernikes was derived using F323N at JSC OTIS. So the code for computing the _delta_ focus needs to be updated for LW to use the special F323N defocus as the reference, instead of just calling the focus model at 3.23 microns.

JarronL:
> I realized at least one additional change is needed in this PR related to the F323N filter's special defocus: we also need to take into account that the OPD model for all the Zernikes was derived using F323N at JSC OTIS. So the code for computing the _delta_ focus needs to be updated for LW to use the special F323N defocus as the reference, instead of just calling the focus model at 3.23 microns.

Oh, right. I had overlooked the fact that all of the WFE focus map data would be contaminated by the F323N filter defocus. I'll look into the best way to accommodate this issue, unless you already have a clear idea of what needs changing?

mperrin:
> a clear idea of what needs changing? 

See https://github.com/spacetelescope/webbpsf/blob/1e6d5975141fe9cdaaef8c1e26e8d3995f6f2f34/webbpsf/optics.py#L1510

It's computing the delta focus relative to a function call rather than the special defocus for F323.

I think we need to edit that part of the code to be something like:
```
ref_defocus = 1.206e-7 if self.instrument.channel.upper() == 'LONG' else focusmodel(opd_ref_wave)
deltafocus = focus_at_wave - ref_defocus
```

mperrin:
I have added the file `si_zernikes_coron_zemax.fits` to the version of the data files used by the CI system, so you ought to be able to re-enable the tests using it on this PR.

JarronL:
Updated focus offset function to take into account the F323N reference. Streamlined it a bit to minimize duplicating referencing the hardcoded 1.206e-7 value. Also, made F212N and F323N explicitly use a deltafocus of 0.

Just saw you comment about adding the Zemax coronagraphy Zernikes. Will uncomment those lines in a few.

JarronL:
Looks like Travis is failing because the coronagraphic PSFs with their new SI WFE no longer match previously calculated PSFs, which is to be expected.

docs/jwst.rst
```diff
@@ -223,14 +217,31 @@ SI WFE
 SI internal WFE measurements are from ISIM CV3 testing (See JWST-RPT-032131 by David Aronstein et al.)
 The SI internal WFE measurements are distinct for each of the modules and
 channels. When enabled, these are added to the final pupil of the optical
-train, i.e. after the coronagraphic image planes.
-
-**Caution** WFE is not yet included for the coronagraph field points; it will use the closest imaging field point. See issue #180.
+train, i.e. after the coronagraphic image planes. For field-points outside of
+the measurement bounds, WebbPSF performs an extraction routine.
```
mperrin: should this be 'extrapolation' rather than extraction?

docs/jwst.rst
```diff
@@ -223,14 +217,31 @@ SI WFE
...
+field-dependent WFE maps.
 
 Wavelength-Dependent Focus Variations
 ---------------------------------------
 
-**TODO**  Add documentation here for the focus variations vs wavelength and how webbpsf models those.
+NIRCam's wavelength-dependent defocus was measured during ISIM CV2 at a given field point
+(See JWST-RPT-029985 by Randal Telfer). Overall, the measurements are consistent with
+predictions from the nominal optical model. The departure of the data from the 
+model curve has been determined to be from residual power in individual filters.
```
mperrin: Should we add here something like "In particular the F323N filter has a significant extra defocus; WebbPSF includes this measured defocus if the selected filter is F323N."

JarronL: Added.

docs/jwst.rst
JarronL: Yup. Midnight typos.

JarronL:
Will look into adding the wavelength-dependent PSF offsets due to the Lyot wedges (#115).

robelgeda:
Hey @JarronL, @shanosborne and I have been looking at this and #298. We thought that your method was the best approach. We would like to generalize it so it can be used for WFIRST as well. 
I was thinking about moving the interpolation/extrapolation code and generalizing certain parts. How would you feel if I made a PR against your fork? That way we can use this PR and keep the other changes.

mperrin:
@robelgeda: Can you explain please why you think we need a version of this for WFI? We're getting the Zernikes from the WFI optical models at the center and 4 corners of each detector, so there should not be any need to extrapolate.   As I've noted before, the WFI Zernikes are provided over the entire photosensitive region of the detectors (from [4,4] to 2044,2044]). 

So I think we ought to be able to make use of a much simpler solution for WFI: Before doing the interpolation between field points, we should just use `numpy.clip` to restrict the pixel coordinate to be in the range `4` to `2044`, e.g. something like
```
effective_position = np.clip(self.field_position, 4, 2044)
```
inside `first.FieldDependentAberration.get_aberration_terms`. Is there some reason that wouldn't work?

robelgeda:
That works just as well. I was thinking long term it might be better to generalize this method for estimating terms for all instruments and missions. I suppose the WFI data will behave well for a long while (since its being simulated). We can go ahead and merge this for Webb and implement clipping for WFI.

shanosborne:
@mperrin - Robel and I were thinking that this will have to be done for WFIRST eventually (once there is real WFE data), so it’d be a better practice and probably faster to just implement it now while it’s in our heads. And since this only adds ~0.5 seconds to a PSF calculation, we wouldn’t be greatly slowing down the WFIRST calculations for no reason. But if you’d rather we just merge this now and deal with the generalization later, I think that’s fine

JarronL:
> Hey @JarronL, @shanosborne and I have been looking at this and #298. We thought that your method was the best approach. We would like to generalize it so it can be used for WFIRST as well.
> I was thinking about moving the interpolation/extrapolation code and generalizing certain parts. How would you feel if I made a PR against your fork? That way we can use this PR and keep the other changes.

@robelgeda I'm good with you making a PR against my fork to implement a generalized version. I still need to clean up a couple things on that fork, but have been preoccupied with modifications to our GTO science program that are due next week. I probably won't get back to it until sometime in July.

shanosborne:
Hey @JarronL - We were actually hoping to get a new version of webbpsf released before the end of June. How would you feel about us merging this PR now, and then any future improvements you want to make can go in a different PR when you get the time? Or would you rather leave this entire PR until later? Is there anything in this PR right now that shouldn’t be merged into the code base?

JarronL:
@shanosborne This PR can get merged as is without any problem. The only thing left to do was address #115, but that can go in a separate PR.

mperrin:
I have finally (apologies for the delays) had time to take a deep look at the results of this extrapolation model, and how it works in a wide range of circumstances. Plots attached. Each plot shows the OPD inferred for a grid of field points across an SI; the field points are a mix of interpolated values, extrapolated values, and (rarely) exact measurements from ISIM testing (shown for cases in which a grid point is within 6 arcsec of a ground test point, in which case interpolation is skipped). 

@JarronL, I'd definitely appreciate your thoughts on how these plots look to you. 
And I'll tag @shanosborne, @Skyhawk172, @obi-wan76 as FYIs that you all might find this interesting. I'll ask Randal to take a look at this too, but as far as I know he's not on GitHub yet.

Plots and commentary below. 

**NIRCam A LW**:
gridplot_nircam_a_lw

Overall looks pretty reasonable. Smoothly varying, and the extrapolated part seems not inconsistent with the interpolated part. The WFE does tend to peak in the corners which are extrapolated; but I think that's likely a true property of the instrument, so these specific values seem plausible.   

**NIRCam B LW**
gridplot_nircam_b_lw

Note, different wavelengths shown for A and B, in part to demonstrate the wavelength dependence part is working. 
If we set the filter to F323N, we see the lookup table adds the increased defocus inherent in that filter. 

gridplot_nircam_b_lw_f323n

**NIRISS**

Mostly OK, but some weird stuff is happening in the interpolation along the right side of the FOV, which I suspect are likely artifacts. Need to consult with a NIRISS optics expert for better insights. This is probably hard to do much better on, given the sparse sampling in ISIM testing.  

 (Note, this plot is on a harder stretch than the NIRCam one, because NIRISS is generally lower WFE internally)
gridplot_niriss

**MIRI**
This is also quite limited by the FOV coverage. We should limit this to the imager FOV probably, rather than extrapolating out into the coronagraph FOV regions since I am not sure we are getting useful model results there. The upper right corner is also not great. But again, we're not likely to do better just on the basis of the 5 field points from ISIM testing, and I'm not sure anyone has bandwidth right now to dig into the optical models to help improve on this. 

gridplot_miri

PR #286, opened 2019-03-18
WFI aberrations handle gracefully wavelengths outside the ref data

### LINKED ISSUE #285: Cannot build PSF for R062 filter in WebbPSF 0.8.0
Despite the documentation saying WebbPSF adds the R062 filter, the aberrations do not appear to support the wavelength range of the filter.  Using the code

```
from webbpsf import wfirst
wfi = wfirst.WFI()
wfi.filter = 'r062'
wfi.detector = 'SCA09'
wfi.detector_position = (2048, 2048)
wfi.options['parity'] = 'odd'
wfi.options['output_mode'] = 'both'

psf = wfi.calc_psf(oversample=4)
```

I first get warning ``The supplied file, $WEBBPSF_PATH/WFI/filters/R062_throughput.fits, does not have a WAVEUNIT keyword. Assuming it is Angstroms.`` and then the error ``RuntimeError: Attempted to get aberrations at wavelength outside the range of the reference data``. This runs correctly if ``wfi.filter = 'z087`` is used instead, which has ``WAVEUNIT= 'angstrom'`` in its header. I believe that the defaulting units are correct, as given by

```
>>> import astropy.io.fits as pyfits
>>> a = pyfits.open('$WEBBPSF_PATH/WFI/filters/R062_throughput.fits')[1].data
>>> a
FITS_rec([( 4000., 0.), ( 4799., 0.), ( 4800., 1.), ( 8000., 1.),
          ( 8001., 0.), (30000., 0.)],
         dtype=(numpy.record, [('WAVELENGTH', '>f8'), ('THROUGHPUT', '>f8')]))
>>> b = pyfits.open('$WEBBPSF_PATH/WFI/filters/Z087_throughput.fits')[1].data
>>> b
FITS_rec([(  5000., 0.), (  7599., 0.), (  7600., 1.), (  9770., 1.),
          (  9771., 0.), (300000., 0.)],
         dtype=(numpy.record, [('WAVELENGTH', '>f4'), ('THROUGHPUT', '>f8')])
```

I have confirmed my module version and webbpsf-data version via ``conda list webbpsf`` giving ``webbpsf 0.8.0 <pip>`` and ``version.txt`` in webbpsf-data also showing ``0.8.0``. Are there any known issues with using the R062 filter in WebbPSF, or any other reasons my PSF creation may have failed? One potential explanation I have found is that the documentation says that the Zernike aberrations only extend to 0.76 microns, but also says that the wavelength range should be extended in Cycle 7. Are my Zernike aberrations incorrect in some way; is there a more up to date table of these values that I do not have in my webbpsf-data?

mperrin:
Should fix #285 

Rather than raising a RuntimeError, it now prints a warning and gives the best estimate result anyway.

### DIFF
--- a/webbpsf/wfirst.py
+++ b/webbpsf/wfirst.py
@@ -65,8 +65,12 @@ class WavelengthDependenceInterpolator(object):
             # we have to interpolate @ this wavelength
             aberration_terms = griddata(self._wavelengths, self._aberration_terms, wavelength, method='linear')
             if np.any(np.isnan(aberration_terms)):
-                raise RuntimeError("Attempted to get aberrations at wavelength "
-                                   "outside the range of the reference data")
+                import warnings
+                wavelength_closest = np.clip(wavelength.value, np.min(self._wavelengths), np.max(self._wavelengths))
+                _log.warn("Attempted to get aberrations at wavelength {:.2g} "
+                        "outside the range of the reference data; clipping to closest wavelength {:.2g}".format(wavelength, wavelength_closest))
+
+                aberration_terms = griddata(self._wavelengths, self._aberration_terms, wavelength_closest, method='linear')
             return aberration_terms
 
 


robelgeda:
Tests failed, re-running

PR #291, opened 2019-03-26
Change removal of different keywords

shanosborne:
@laurenmarietta is using the psf grid functionality in mirage and these changes should make it's implementation there easier while also not changing anything important for users following the original workflow.

This PR:
1. In `to_model()` and `to_griddedpsfmodel()`: No longer deletes the duplicate keywords of `DET_YX*` and `OVERSAMP` (so now normal users will just have some extra keywords which shouldn't hurt anything)
2. In `to_griddedpsfmodel()`: Now uses pop to delete un-needed keys so that if those keys aren't present, it doesn't throw an error

### DIFF
--- a/webbpsf/utils.py
+++ b/webbpsf/utils.py
@@ -719,8 +719,9 @@ def to_griddedpsfmodel(HDUlist_or_filename=None, ext=0):
 
     # Convert header to meta dict
     header = header.copy(strip=True)
-    header.remove('COMMENT', remove_all=True)
-    header.remove('', remove_all=True)
+    header.pop('COMMENT', None)
+    header.pop('', None)
+    header.pop('HISTORY', None)
     meta = OrderedDict((a, (b, c)) for (a, b, c) in header.cards)
 
     ndd = NDData(data, meta=meta, copy=True)
@@ -734,7 +735,7 @@ def to_griddedpsfmodel(HDUlist_or_filename=None, ext=0):
         ndd.meta['oversampling'] = ndd.meta['OVERSAMP'][0]  # pull the value
 
     # Remove keys with duplicate information
-    ndd.meta = {key.lower(): ndd.meta[key] for key in ndd.meta if 'DET_YX' not in key and 'OVERSAMP' not in key}
+    ndd.meta = {key.lower(): ndd.meta[key] for key in ndd.meta}
 
     # Create model
     model = GriddedPSFModel(ndd)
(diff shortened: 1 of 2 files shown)


mperrin:
This looks reasonable, and the two of you know this part of the code better than I do. That said I'm not sure I follow all the logic here - why is it necessary to delete the COMMENT and HISTORY keywords, rather than just leaving them in place? Given your point 1 above - it's OK to leave in place extra DET_YX keywords - then why isn't it also OK to leave in place the HISTORY keywords in point 2?

webbpsf/utils.py
```diff
@@ -719,8 +719,9 @@ def to_griddedpsfmodel(HDUlist_or_filename=None, ext=0):
 
     # Convert header to meta dict
     header = header.copy(strip=True)
-    header.remove('COMMENT', remove_all=True)
-    header.remove('', remove_all=True)
+    header.pop('COMMENT', None)
+    header.pop('', None)
+    header.pop('HISTORY', None)
```
mperrin: Why is it desirable or necessary to delete these header keywords? I'm probably missing some simple explanation...

webbpsf/utils.py
```diff
@@ -734,7 +735,7 @@ def to_griddedpsfmodel(HDUlist_or_filename=None, ext=0):
         ndd.meta['oversampling'] = ndd.meta['OVERSAMP'][0]  # pull the value
 
     # Remove keys with duplicate information
-    ndd.meta = {key.lower(): ndd.meta[key] for key in ndd.meta if 'DET_YX' not in key and 'OVERSAMP' not in key}
+    ndd.meta = {key.lower(): ndd.meta[key] for key in ndd.meta}
```
mperrin: With this change, the comment line immediately above about removing keys with duplicate information no longer applies. Instead this is just "turn all metadata keys into lowercase" now, I think.

webbpsf/utils.py
shanosborne: The idea was that we don’t want these keys in a meta dictionary because they’re either unhelpful or incomplete. E.g. getting a meta entry of  `'': ('', '')` isn’t useful. And for the `comment`/`history` cards, right now you only get the first line rather than the entire entry and we determined that the `comment`/`history` cards don't contain important enough information to do the work to get the all the lines in the meta dict (the information in them can be found in documentation). But if you think that those lines are useful, I can make it so its all included.

mperrin: I don’t think they’re so useful as to be necessary. All the other keywords that do get preserved should maintain sufficient information about the parameters used in case someone later needs to investigate how a given PSF grid was generated, right?

shanosborne: Yes exactly. There should be no information in the comment/history cards that can't be found elsewhere (either in other keys or in the webbpsf/photutils notebooks)

Issue #292, opened 2019-04-01
Pupil mask value is not present in `psf_grid` output


bhilbert4:
I've been thinking about the easiest way to allow Mirage users to specify fits files containing PSF grids. The two options I've been thinking about involve a rigid filename format that includes all relevant instrument and webbpsf details (detector/filter/pupil/oversampling/OPD/etc), or having Mirage look through the headers of all fits files in a given directory to identify the proper file.

But in the course of producing gridded PSF files, I've found that the value of the pupil mask used by WebbPSF is not present in the output file header, nor in the (automatically generated) filename. For example, in the code below, running with and without the `nrc.pupil_mask` line produces files with the same name and header info, but different PSFs.

```
nrc = webbpsf.NIRCam() 
nrc.filter = 'F200W' 
nrc.detector = 'NRCB1' 
nrc.pupil_mask = 'WEAK LENS +8'                                                                                                                                                                   

grid = nrc.psf_grid(num_psfs=4, all_detectors=False, oversample=2, fov_pixels=301, 
                                add_distortion=False, save=True) 
```

bhilbert4:
I should also point out that the OPD realization number is also not present in the file header nor filename.

shanosborne:
For both of these comments, it was a conscious choice not to include the information in the default file name, because at some point it just becomes too long (and if people want it they can use the `outfile` keyword arg to change the file name). But I agree that this information should be in the header/meta dictionary. Now as to whether or not they’re in there:
* For your first comment, yes you’re right this is a bug. I’ll add it so you’ll get `PUPIL` and `CORONMSK` keywords if you set the `pupil_mask` and `image_mask` respectively.
* For your second comment, this information is in fact in the header/meta dictionary. If you set your `pupilopd` to be a certain realization (e.g. `nrc.pupilopd = ('OPD_RevW_ote_for_NIRCam_predicted.fits.gz', 7)`), your `PUPILOPD` keyword will say `PUPILOPD= 'OPD_RevW_ote_for_NIRCam_predicted.fits.gz slice 7' / Pupil OPD source`. If you don’t specify the slice (so webbpsf defaults to the first slice), your  `PUPILOPD` keyword will just say `PUPILOPD= 'OPD_RevW_ote_for_NIRCam_requirements.fits.gz' / Pupil OPD source`. This part does require the knowledge that not specifying the slice will result in using the default first slice, but it also exactly matches what the header will look like if you were to create a PSF using `calc_psf`.

[...]

PR #293, opened 2019-04-05
Fix bug in header of psf_grid output

shanosborne:
Now when `image_mask` or `pupil_mask` is set, that information will be included in the header/meta data for the output of the `psf_grid` method. Fixes issue #292 

Also some random PEP8 updates


mperrin:
Looks good to me. Once all the tests pass this is good to merge I think. 

For the requests in #292, we should separately think about adding a new keyword with the slice number in the OPD rather than just sticking it in the string with the filename - I don’t think that was one of my better decisions and in retrospect it’s both confusing and harder to use that way. But that can & should be a separate fix for calc_psf outputs in general.

Issue #290, opened 2019-03-20
Extend psf_grid functionality to WFIRST WFI class


mperrin:
As discussed in Slack with @shanosborne - 

Right now `psf_grid` is defined on `JWInstrument` so it won't work for the WFIRST `WFI` class. This should be relocated to `SpaceTelescopeInstrument` or else a comparable implementation added to `WFIRSTInstrument` to enable the creation of PSF grids for WFI with the same programming interface. 

@shanosborne to lead modification of `psf_grid`, @robelgeda to support with any necessary knowledge of WFIRST code, please. Seems like this ought to be straightforward.

mperrin:
@shanosborne, @ojustino I think this may be complete? Can this issue be closed?

shanosborne:
Yep, looks like the PR back in 2019 fixed this, and I just used the wrong wording so it did auto-close the issue when the PR merged.

PR #294, opened 2019-04-15
Update psf_grid for WFI class

shanosborne:
I updated the `psf_grid` method so it could also be used for the `WFI` class. I moved the method up to the `SpaceTelescopeInstrument` class and added some checks for the add_distortion keyword which doesn't exist for the `WFI` class. 

I added new tests for this functionality. Currently one of these tests is failing because of a possible bug (TBD) that occurs when you create a PSF on the edge of any of the WFI detectors (e.g. setting `detector_position` to (0,0) or (1,1) causes an error). @robelgeda and I are looking into why this is happening.

I also made a few clean up changes to `gridded_library.py`


shanosborne:
Fixes issue #290

shanosborne:
Fixes small issues/improvements brought in in #303

mperrin:
Ready to test and merge once #309 is merged.

mperrin:
FYI @shanosborne the big WFIRST cycle 8 PR is now merged

shanosborne:
I'll look into this test failing tomorrow

shanosborne:
The test wasn't passing because it didn't include the scale factor to take into account the normalization as expected in the `photutils.GriddedPSFModel` (see #302). It should be fixed now.

webbpsf/gridded_library.py
```diff
@@ -121,6 +130,8 @@ def __init__(self, instrument, filters="all", detectors="all", num_psfs=16, psf_
         # Set PSF attributes for the 3 kwargs that will be used before the calc_psf() call
         if "add_distortion" in kwargs:
             self.add_distortion = kwargs["add_distortion"]
+            if self.webb.name == "WFI":
```
mperrin: Minor: `if self.webb.name == "WFI"` seems awkward. We might want to consider renaming the `.webb` attribute to `.instrument` or something similar.

webbpsf/gridded_library.py
```diff
@@ -329,10 +341,12 @@ def create_grid(self):
             meta["DIFFLMT"] = (psf[ext].header["DIFFLMT"], "Diffraction limit lambda/D in arcsec")
             meta["FFTTYPE"] = (psf[ext].header["FFTTYPE"], "Algorithm for FFTs: numpy or fftw")
             meta["NORMALIZ"] = (psf[ext].header["NORMALIZ"], "PSF normalization method")
-            meta["JITRTYPE"] = (psf[ext].header["JITRTYPE"], "Type of jitter applied")
-            meta["JITRSIGM"] = (psf[ext].header["JITRSIGM"], "Gaussian sigma for jitter [arcsec]")
             meta["TEL_WFE"] = (psf[ext].header["TEL_WFE"], "[nm] Telescope pupil RMS wavefront error")
 
+            if self.webb.name != "WFI":
```
mperrin: Hmm, I'm not a fan of having these keywords be instrument-dependent. It seems like it would be more consistent overall to write these keywords for WFI too, but with values like `None` or `0` to indicate that such models are not applied. Better yet, we should add in the actual convolution step for WFI eventually. (This is out of scope for this PR, just want to note it's something we should do eventually) . Do you agree?

webbpsf/tests/test_psfgrid.py
```diff
@@ -204,3 +206,62 @@ def test_saving(tmpdir):
 
     # Remove temporary directory
     tmpdir.remove()
+
+
+def test_2d_to_griddedpsfmodel():
+    """Test that utils.to_griddedpsfmodel function works fof a 2D HDUList input"""
```
mperrin: minor: typo "fof"

webbpsf/utils.py
```diff
@@ -676,19 +676,21 @@ def combine_docstrings(cls):
...
     Parameters
     ----------
-    HDUlist_or_filename : string
+    HDUlist_or_filename : HDUList or str
         Either a fits.HDUList object or a filename of a FITS file on disk
-    ext : int
-        Extension in that FITS file
+    ext_data : int
+        Extension of the data in the FITS file
+    ext_header : int
```
mperrin: What's the use case for pulling the data and header from different extensions?

webbpsf/tests/test_psfgrid.py
mperrin: The motivation for this test is just to ensure there's not any error for the degenerate case of a single-PSF "grid", is that correct?

webbpsf/gridded_library.py
shanosborne: That sounds fine to me! The only thing is that `self.instr` exists as an attribute for the instrument name, so I'm not sure if `instr` and `instrument` would be too similar (i.e. if may be better to then also change `self.instr` to something like `instr_name`). This change would be a pretty basic find/replace and then also double checking if the documentation needs to be updated

webbpsf/gridded_library.py
shanosborne: Right,`WFI` PSFs don't have `JITRTYPE` or `JITRSIGM` keywords, so they can't be carried through here from the PSF header to the `GriddedPSFModel` object. I can set these values to 0 for the `GriddedPSFModel` object here, but they still wouldn't exist for a `WFI` PSF, since that change would have to come from outside this grid method. At this point (knowing it'll be fixed outside this special case in the future), do you think it's better for the `GriddedPSFModel` header to match the PSF header for `WFI` or to have the model's all have the same headers but have WFI model not match it's PSF?

webbpsf/tests/test_psfgrid.py
shanosborne: Yes, the creation of a single PSF "grid" is tested in `test_one_psf()` and then this test checks turning the `HDUList` object back into a `GriddedPSFModel` object using `utils.to_griddedpsfmodel` (since it's 1 extra step compared to a 3D input).

Also I'll update that typo real quick

webbpsf/utils.py
shanosborne: This was just in case someone had their own data (e.g. a single PSF in a fits file like from WebbPSF) that they wanted to turn into a `GriddedPSFModel` object, with the header information in the primary extension and the data in the 1st extension, which is a fairly common practice. I think this was requested by Lauren/MIRAGE, although I can't remember if it was for a case that still exists, or if it was for a patch of some kind.

webbpsf/gridded_library.py
mperrin: Oh I'm just going to modify the WFI class so that every WFI PSF also has the same keywords for jitter. I didn't realize they were inconsistent or we would have just fixed that long ago so they all match.

webbpsf/utils.py
mperrin: OK, I can't see immediately what that case might be, but that's fine in any case.

PR #295, opened 2019-04-16
Feature/sm sur

AldenJurling:
This branch adds support for SM moves to the move_sur() method. Because segment moves are made through move_seg_local and SM moves are made through move_sm_local, this requires a check in move_sur on whether the segment is, in fact, the SM, and take a different code path if it is. 

In the course of this I also implemented the group= keyword argument which was previously ignored. The behaviour was unspecified so I made it take a zero-based index into the groups and execute that one. 

I also added support for moves in meters because the SUR in order to process the GA SUR I was working with it. 

I can rebase this branch off of your current master or latest merge commit if you prefer a cleaner history.

I'm not sure why my webbpsf master branch picked up the 2018-12 base commit its branched from.

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -1705,7 +1705,11 @@ class OTE_Linear_Model_WSS(OPD):
         import jwxml
 
         sur = jwxml.SUR(sur_file)
-        for grp in sur.groups:
+        if group is not None:
+            groups = [sur.groups[group]]
+        else:
+            groups = sur.groups
+        for grp in groups:
             for update in grp:
                 if verbose:
                     print("Move seg {} by {}".format(update.segment, str(update)))
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -1699,7 +1699,20 @@ class OTE_Linear_Model_WSS(OPD):
             self.update_opd(display=display)
 
     def move_sur(self, sur_file, group=None, verbose=False):
-        """ Move using a JWST Segment Update Request file
+        """
+        Move using a JWST Segment Update Request file
+
+        Parameters
+        ----------
+        sur_file : file name
+            Path to SUR XML file
+        group : zero-based int index
+            Index to a single group to run. Default is to run all groups.
+        verbose : bool
+            Flag controlling whether moves are printed.
+
+        Returns
+        -------
 
         """
         import jwxml
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -1325,7 +1325,7 @@ class OTE_Linear_Model_WSS(OPD):
         roc : float
             radius of curvature mechanism adjustment, in microns.
         trans_unit : str
-            Unit for translations. Can be 'micron', 'millimeter','nanometer', 'mm', 'nm', 'um'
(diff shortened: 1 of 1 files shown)


webbpsf/opds.py
```diff
@@ -1716,16 +1735,26 @@ def move_sur(self, sur_file, group=None, verbose=False):
...
-                                        rot_unit=rot_unit,
-                                        trans_unit=trans_unit,
-                                        delay_update=True)
+                    if update.segment == 'SM':
+                        self.move_sm_local(xtilt=update.moves['X_TILT'],
+                                           ytilt=update.moves['Y_TILT'],
+                                           xtrans=update.moves['X_TRANS'],
+                                           ytrans=update.moves['Y_TRANS'],
+                                           piston=update.moves['PISTON'],
+                                           rot_unit=rot_unit,
```
mperrin: It looks like the `update.absolute` parameter is not passed through to the SM move command. I'm guessing this is an unintentional omission? I can't remember the details but I assume that absolute moves are allowed for the SM just as for any primary segment.

mperrin:
@AldenJurling - wanted to ping you re: my comments on this PR from several weeks ago. Could you take a moment to look at those and let me know how you want to proceed, please? Should only take a minute or two. Thanks.

AldenJurling:
@mperrin - I admit I don't remember exactly what I was thinking about absolute moves, but looking at the code, absolute doesn't appear to be a valid argument to move_sm_local. I'm not sure if I missed it, or just didn't pass it through since it wasn't a legal argument. 
```
    def move_sm_local(self, xtilt=0.0, ytilt=0.0, rot_unit='urad',
                      xtrans=0.0, ytrans=0.0, piston=0.0, trans_unit='micron', display=False,
                      delay_update=False):
        """ Move the secondary mirror in pose, using segment-local control coordinates.

        These motions are always commanded in the segment-local "Control" coordinate systems,
        which are distinct for each segment. The SM is also handled a bit differently than all the
        PMSAs in terms of its local coordinate system, which this function handles behind the scenes.

        Parameters
        -----------
        segment : str
            Segment name, e.g. 'A1'. Use 'SM' for the secondary mirror.
        xtilt, ytilt, clocking : floats
            Tilt angle, in microradians by default. 'xtilt' means tilt *around the X axis*, and similarly for ytilt.
        xtrans, ytrans, piston : floats
            Displacement distance, in microns by default.
        trans_unit : str
            Unit for translations. Can be 'micron', 'millimeter','nanometer', 'mm', 'nm', 'um'
        rot_unit : str
            Unit for rotations. Can be 'urad', 'radian', 'milliradian', 'arcsec', 'arcmin', 'milliarcsec'
        display : bool
            Display after moving?
        delay_update : bool
            hold off on computing the WFE change? This is useful for computational efficiency if you're
            moving a whole bunch of segments at once. Incompatible with display=True.
        """
```

I may have just assumed it wasn't valid for the SM since it was supported downstream, but I don't have any reason to know that. If its a valid option in the SUR, which it probably is, it needs to be implemented in move_sm_local and then threaded through to it.  That should be pretty easy, but I won't be able to do it right now.

PR #298, opened 2019-04-19
Use nearest aberration terms for out of bound field points (WFI)

### LINKED ISSUE #296: WFI Detector Field Coverage
Blocking: #294
@shanosborne  and I found that the current reference data does not cover all pixels in a given WFI detector. Each detecter has 5 field points. Four at the corners and one at the center:
WFI_Points

We interpolate between these points to determine the field dependance of the PSF at a given point. The reference data does not cover all corners which leads to an inability to interpolate aberration terms for border pixels. The reference data is indented by a 4 pixel margin:
WFI_Points_corner

When the user selects a pixel outside the area covered by the interpolation, an exception is raised:
`RuntimeError("Attempted to get aberrations for an out-of-bounds field point")`

The reasoning behind the placement of the field points can be found in this code comment:
https://github.com/spacetelescope/webbpsf/blob/master/webbpsf/wfirst.py#L197%03-L202

The question is now how to handle this error? Should we inform the user that they picked a valid pixel and that it is out of bound of the ref data? Should we extrapolate or estimate  (closest point approximation)?

robelgeda:
closes #296 
Use nearest field point's aberration terms when the user selects a field position outside of the interpolatable area.

### DIFF
--- a/webbpsf/tests/test_wfirst.py
+++ b/webbpsf/tests/test_wfirst.py
@@ -104,17 +104,10 @@ def test_WFI_limits_interpolation_range():
     wfi = wfirst.WFI()
     det = wfi._detectors['SCA01']
     det.get_aberration_terms(1.29e-6)
+
     det.field_position = (0, 0)
-    with pytest.raises(RuntimeError) as excinfo:
-        det.get_aberration_terms(1.29e-6)
-    assert 'out-of-bounds field point' in str(excinfo.value), (
-        "FieldDependentAberration did not error on out-of-bounds field point"
-    )
-    with pytest.raises(RuntimeError) as excinfo:
-        det.get_aberration_terms(1.29e-6)
-    assert 'out-of-bounds field point' in str(excinfo.value), (
-        "FieldDependentAberration did not error on out-of-bounds field point"
-    )
+    det.get_aberration_terms(1.29e-6)
+
     det.field_position = (2048, 2048)
 
     # Test the get_aberration_terms function uses approximated wavelength when
--- a/webbpsf/utils.py
+++ b/webbpsf/utils.py
@@ -741,3 +741,76 @@ def to_griddedpsfmodel(HDUlist_or_filename=None, ext=0):
     model = GriddedPSFModel(ndd)
 
     return model
+
+
+def nearest_polygon_point(x, y, target):
+    """
+    Given a set of vertices find the nearest
+    point on the nearest edge
+
+    Parameter
+    ---------
+    x, y : numpy array
+        x and y values of vertices
+    target : tuple, array
+        x and y values of target point
+
+    Returns
+    -------
+    nearest : array
+        x and y values of nearest point to target
+    """
+
+    if len(x) == 1 and len(y) == 1:
+        return np.array([x[0], y[0]])
+    elif len(x) != len(y):
+        raise RuntimeError("Input x and y arrays are not the same dimensions")
+    elif len(x) == 0 or len(y) == 0:
+        raise RuntimeError("Can not find nearest point"
+                           " because input array is empty")
+
+    # Convert to array
+    target = np.asarray(target)
+    if len(target.shape) != 1:
+        raise RuntimeError("Multiple targets passed to nearest"
+                           " point approximation function")
+
+    vertices = []
+    for i in range(len(x)):
+        for j in range(len(x[i:])):
+            k = i + j
+            if i == k:
+                continue
+            vertices.append(
+                np.array([
+                    np.array([x[i], y[i]]),
+                    np.array([x[k], y[k]])
+                ])
+            )
+
+    min_dist = np.inf
+    nearest_pnt = None
+    for i, vertex in enumerate(vertices):
+        p1, p2 = vertex
+
+        v = p2 - p1
+        u = p1 - target
+        t = - (np.dot(u, v) / np.dot(v, v))
+
+        if 0 <= t <= 1:
+            # Point lies on line segment
+            point = ((1 - t) * p1) + (t * p2)
+            dist = np.linalg.norm(point - target)
+        else:
+            # Point is one of the edge vertices
+            dist1 = np.linalg.norm(p1 - target)
+            dist2 = np.linalg.norm(p2 - target)
+
+            point = p2 if dist1 > dist2 else p1
+            dist = dist2 if dist1 > dist2 else dist1
+
+        if dist < min_dist:
+            min_dist = dist
+            nearest_pnt = point
+
+    return nearest_pnt
--- a/webbpsf/wfirst.py
+++ b/webbpsf/wfirst.py
@@ -70,9 +70,9 @@ class WavelengthDependenceInterpolator(object):
                 if isinstance(wavelength, u.Quantity):
                     wavelength = wavelength.to(u.m).value
                 wavelength_closest = np.clip(wavelength, np.min(self._wavelengths), np.max(self._wavelengths))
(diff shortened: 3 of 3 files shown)


robelgeda:
I agree with this point. Since this is an issue with the Webb part of the code, I thought I might as well write a closest point approximation function. I added a `nearest_polygon_point` which basically detects the closest point on the closest edge. Here is a demo running the function in real time:
move_demo-2
(The blue points are field points and the red one is the target)

I also added a warning to notify the user that the coefficients are approximated at a different point.

Screen Shot 2019-04-22 at 11 11 39 AM

If this looks okay, I can add this solution to the Webb section of the code along with other optimizations in another PR.

robelgeda:
Travis failing because it expected out of bound interpolation to fail. I will write new tests for this.

mperrin:
This is really nice. Potentially a better solution than the extrapolation parts in the WebbPSF PR, do you think? I can see pros and cons to each approach. 

In the figure above I was a bit surprised that the red line doesn’t look perpendicular to the blue lines. I would expect the line to the closest point inside the polygon to intersect the boundary at right angles. But then I realized this is probably just due to the aspect ratio of that plot not being the same for X and Y. If you were to remake that with `aspect=“square”`, does the red line end up at a right angle to the blue boundary segments?

robelgeda:
Yep, you are right! It was the aspect ratio:

move_demo-3

When it comes to extrapolation, it depends on the number of points we have. I think this method is very strong when the field points cover a large portion of the detector. Extrapolation may be better for field points concentrated in a small area. Another advantage of extrapolation is the speed; this algorithm runs at ~ O(n^2) since we have to construct and check every edge. If we like the idea of this implementation I can add a convex hull algorithm (finds the boundary edges) to mitigate the overhead of computing internal edges. I am good with ether.

mperrin:
@jarronL have you seen this PR? This is a potential alternative to the extrapolation of WFE that you've been doing. Pros and cons both ways. I'd be interested in your thoughts on this?

JarronL:
As you say, I can see pros and cons for the Webb code. I like that it's elegant and straightforward and would probably be fine for many of the SI FoVs. There are cases, though, where you could get the exact same or very similar results (ie., along the perpendicular), whereas the extrapolation routine always provides unique values. For instance, the WFE coverage of MIRI is fairly lacking in the upper right region of a V2/V3 plot, so I think that you would see streaky artifacts if you filled that region in with the `nearest_polygon_point` function. I like that extrapolation algorithms take into account information interior to the border of the existing measurements, so they can preserve gradients and trends that exist within WFE data.

Would it be possible to add in one more dimension here? Maybe use three (or four) of the nearest points and calculate some sort of inverse distance weighting? In the end, I would just like to see some reasonable change along that perpendicular red line.

robelgeda:
@JarronL this sounds good to me. How would you use the inverse distance weighting though? Are you thinking of using it for extrapolation?

mperrin:
It seems to me we could perform a linear extrapolation outwards along the perpendicular red line. Something like this:
 - Let's call our desired location field point F. Proceed as above to find the nearest point inside the polygon via `nearest_point_polygon`, and the red line vector to that point. Let's call this point N1. 
 - Compute WFE at that point (i.e. Zernike polynomial coefficients). Let W1 = WFE(N1)
 - Choose a point along the extension of the red line vector into the interior of the polygon. For instance, we could imagine extending the red line's length by an extra 50%, and using the point inside the polygon at that distance along the line. Let's call this point N2. 
 -  Compute WFE at that point too. Let W2 = WFE(N2). 
 - Compute the difference of the two sets of Zernike coefficients, dW = W1-W2. That gives you a measure for how the WFE could be changing outwards along that line. Then we scale this based on the relative lengths of the red line inside and outside the polygon. In this example we extended it by 50% inside so that means we'd take dW/0.5 here.
 - The WFE at our original point F can then be extrapolated as W1 + dW/0.5. 

Hopefully that pseudocode is clear enough; let me know if not. There are of course more sophisticated ways to do this. I think we should not overthink this too much - there is no way we're going to get this precisely correct, not without a lot of work using the correlated Zemax/Code V models for all the instruments which I am not sure is needed. 

As I see it the main goals we should try to achieve right now is to have a WFE that is:
1.  continuously varying (so there are not large areas of the FOV with artificially exactly the same WFE), 
2. in a way that is not too computationally intensive to compute, 
3. and which is fairly robust to the input assumptions/data about WFE, to avoid extrapolations that blow up into unrealistic values. 

I think assuming a linear dependence of the WFE near the FOV edges probably will do all that reasonably well. This presumably is not too different from the RegularGridInterpolator approach used in #283. 

Having written out the above, I'm now starting to wonder if it isn't just going to be equally complicated as using RegularGridInterpolator... There may not be a particularly elegant way to do this.

JarronL:
Ran some pretty basic benchmark tests for the interpolation and extrapolation methods as implemented in #283. 

```python
# Import WebbPSF and setup multliprocessing
import webbpsf
webbpsf.webbpsf_core.poppy.conf.use_fftw = False
webbpsf.webbpsf_core.poppy.conf.use_multiprocessing = True
ncores = 8
webbpsf.webbpsf_core.poppy.conf.n_processes = ncores 

inst = webbpsf.NIRCam()
inst.filter = "F430M"
inst.detector_position = (1024,1024)

# Baseline test: No SI WFE, no distortion
inst.include_si_wfe = False
%timeit psf = inst.calc_psf(add_distortion=False, monochromatic=4.3e-6)                                                                                
# Result: 911 ms ± 5.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit psf = inst.calc_psf(add_distortion=False, nlambda=ncores)                                                                                
# Result: 5.62 s ± 177 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Turn on SI WFE at center of detector (will use interpolation)
inst.include_si_wfe = True
%timeit psf = inst.calc_psf(add_distortion=False, monochromatic=4.3e-6)                                                                                
# Result: 1.41 s ± 12.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit psf = inst.calc_psf(add_distortion=False, nlambda=ncores)                                                                                
# Result: 6.1 s ± 96.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Extrapolation region
inst.detector_position = (0,0)
%timeit psf = inst.calc_psf(add_distortion=False, monochromatic=4.3e-6)                                                                                
# Result: 1.8 s ± 12.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit psf = inst.calc_psf(add_distortion=False, nlambda=ncores)                                                                                
# Result: 6.53 s ± 85.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
In the end, the overheads of adding SI WFE is about 0.5 sec on my machine, and another 0.4 sec if performing the extrapolation algorithm. I ran a few other tests with varying `fov_pixels`, and this overhead appears to stay constant and does not scale with PSF size (which makes sense because we're always modifying the same OPD map). These time differences are the same for NIRCam coronagraphy.

While I'm always in favor of time savings, I'm not sure how much we want to delve into saving an extra half second for PSF calculations?

robelgeda:
@mperrin @JarronL 
I that case I think it would make sense to pick the method that gives the best estimates. It sounds like the interpolation method might serve us well. I will also add some comments on possible improvements to the interpolation code that could go into #283.

robelgeda:
@mperrin just checking in, did we decide which solution to adopt?

Issue #302, opened 2019-05-10
outputs from psf_grid do not match normalization expected by photutils.GriddedPSFModel


mperrin:
This issue's been going on a journey:  
First discovered by @bhilbert4 in the course of work on https://github.com/spacetelescope/mirage/pull/134
Then filed as an issue in photutils: https://github.com/astropy/photutils/issues/811
But it actually turns out to be something we should fix here in webbpsf... 

As @larrybradley wrote in https://github.com/astropy/photutils/issues/811
>  `GriddedPSFModel` expects that the input oversampled PSFs are normalized to `oversampling**2`. This should be better documented. 

But the `psf_grid` function in `webbpsf` instead outputs oversampled PSFs that are normalized to unity. We should fix that, by simply applying the right multiplicative scale factor before outputting the grid. 

(Incidentally I completely agree with the normalization scheme adopted in `GriddedPSFModel` and based on Jay Anderson's original ePSF work. Implicitly it's "normalize to units of surface brightness independent of pixel sampling" but that has the same effect as normalizing to oversample**2. That's definitely what makes sense for the ePSF interpolation calculations.)

mperrin:
@shanosborne If you want you can fix this as owner/author of the psf_grid code, or tell me to fix it. Either way I'm pretty sure it is literally just a one line fix

shanosborne:
@mperrin since it sounds like you know exactly what needs to be done, how about you put in the fix

mperrin:
Closed by PR #311

PR #311, opened 2019-07-22
Fix issue 302, PSF normalization as expected by GriddedPSFModel

mperrin:
Simple fix for issue #302.


mperrin:
By some git weirdness this PR is showing multiple commits including older ones from March that were already merged. Not sure why. Only the latest one is new here.

PR #319, opened 2019-10-29
Add environment.yml file

shanosborne:
At the request the JWST Mission Office, INS-created community tools for JWST should meet certain standards for Cycle 1. One of these is details on the software's dependencies in a `requirements.txt` or `environment.yml` file.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,21 @@
+channels:
+  - http://ssb.stsci.edu/astroconda
+  - defaults
+dependencies:
+  - ipython=7.8.0
+  - pip=19.3.1
+  - python=3.6.9
+  - setuptools=41.4.0
+  - pip:
+    - astropy==3.2.1
+    - docutils==0.15.2
+    - jwxml==0.3.0
+    - matplotlib==3.1.1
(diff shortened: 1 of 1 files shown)


shanosborne:
I tested this file by creating an environment with it using the command `conda env create -n envname -f environment.yml` and all the `webbpsf` tests passed (provided you have the version of `webbpsf` that matches the version of `poppy` in the file, but I think that's a fair requirement and I'm not sure of a way around that)

mperrin:
As discussed in person, we should revise this to give minimum versions rather than exact versions, I think. 

We can also drop `six` from this and from the setup.py, because that's no longer needed now that we've dropped Python 2 compatibility. 

In the dependencies file, why do you list many of the packages as pip installations specifically? I think we can just use conda directly, without invoking pip, to install all of those.

mperrin:
Also, `sphinx` is not required to use webbpsf. It's just required to compile the documentation, but we don't expect users to ever do that themselves; they should just look at the page on RTD. So perhaps we don't need to include sphinx in the environment.yml file either?

mperrin:
Closing, because superseded by #321

PR #321, opened 2019-11-06
Add environment.yml file, and update minimum dependencies

mperrin:
Supersedes #319 

 - Adds environment.yml, based on @shanosborne's version but converted to >= minimum-supported rather than == exact versions. 
 - Minimum versions are updated for most of the required packages. In may cases these hadn't been updated in a looooong time. I semi-arbitrarily chose to update the minimum supported version to major releases from about 2 years ago, e.g. numpy 1.13 and scipy 1.0 and matplotlib 2.0 all from 2017. 
 - Updates setup.py requirements versions for consistency with the environment.yml file

In addition:
 - Updates the Travis setup, to (a) increase the minimum numpy version to 1.13 consistent with the above, and (b) start running tests on Python 3.7 as well as 3.6 and 3.5. (For now 3.8 is still too new, and reportedly still problematic in some parts of the scipy stack, so let's not include that yet)


mperrin:
@shanosborne do you want to review my changes on top of your PR, or are you busy enough that I should just go ahed and merge this myself (after the tests pass etc)?

shanosborne:
I can do a quick review of it- It shouldn't take too long and travis passing tells us the most important stuff

PR #328, opened 2019-11-27
WFR4+MIMF_KDP: add LookupTable optic class for Ball-delivered OPDs

mperrin:
Add a new optic class that reads OPDs from a lookup table provided by Ball, intended specifically for Wavefront Rehearsal 4 (WFR4) sim data generation in Nov/Dec 2019.  

 - Reads OPDs from the FITS file I made from Acton's OPDs. 
 - Determines the closest field point in that table to the chosen SI detector coordinates, and loads that OPD (Note, in general the field points used by Acton's ITM calc and by the webbpsf.psf_grid calc will not match exactly; this is a known source of imprecision / error in the sims currently. Good enough for now but may try to improve it eventually. 
 - Adds in an extra 0.8 microns defocus to NIRISS, per Acton's intent for WFR4 sim scenario
 - Option to remove piston, tip, tilt from the OPDs, since ITM puts in some extra tip/tilt (maybe real, maybe sim artifact? Unclear at this time.)
 - Also small tweaks to the PSFs grid display function, to add a color bar and allow adjusting display scale. 

This is not general-purpose, and more of a hack than usual in webbpsf... Not sure if I will actually merge this or not; we can just use this branch for the WFR4 sims. 

See also the notebooks in `/ifs/jwst/tel/wfr4_mirage_sims/phase_maps_from_ball/`

FYI @obi-wan76 @Skyhawk172 @shanosborne


mperrin:
Rebased onto current develop.

mperrin:
This is old code, developed specifically for WFR4 as a short-term implementation, and is not nearly as rigorous as the subsequent implementations of better ways to exchange information between ITM and WebbPSF+MIRAGE. I don't see any need to keep or merge this branch so I'm going to close this PR.

mperrin:
Looks like we may need this again after all for the MIMF KDP simulation dataset...

mperrin:
Updated for 2021 March MIMF-2 Key Decision Point practice. See also notebooks in `/ifs/jwst/tel/MIMF_KDP_Practice`

mperrin:
This functionality has been useful in enough team practices and LREs and may well be useful in the future at some point. Rather than keeping this PR open indefinitely I'm going to go ahead and merge. 

This is an optional class, the use of which requires manually selecting it in to replace the default (as is done in the LRE and MIMF KDP notebooks). Merging it will have no impact on the results of other calculations with webbpsf.

PR #335, opened 2020-01-09
NIRCam coronagraphy PSF offsets

JarronL:
Apply wavelength dependent offsets for coronagraphic PSFs. This is due to the dispersion from the optical wedge in the coronagraphic pupil mask. Primarily effects the LW channel (approximately a 0.032 mm/um dispersion). SW channel is almost a factor of 10 smaller (fairly negligible), but we include it for completeness. The PSF shift is implemented by modifying the tilt Zernike coefficient (Z3) within the `NIRCamFieldAndWavelengthDependentAberration` function. Values have been determined using the Zernike offsets as reported in the NIRCam Zemax models where the center reference positions correspond to the NIRCam target acquisition filters (3.35um for LW and 2.1um for SW). Implementation is similar to the NIRCam Zernike and focus updates from #283.


webbpsf/optics.py
```diff
@@ -465,13 +465,18 @@ def __init__(self, name='GR700XD', which='Bach',
...
+            either a scalar wavelength (meters) or a Wavefront object
         """
 
         if isinstance(wave, poppy.Wavefront):
             wavelength = wave.wavelength
         else:
-            wavelength = float(wave)
-            wave = poppy.Wavefront(wavelength=wave)
+            wave = poppy.Wavefront(wavelength=float(wave))
+            wavelength = wave.wavelength
```
mperrin: This seems a minor but benign change; I'm curious if there was any subtle bug or issue with the prior code? Or if not then is this just a stylistic change for clarity?

webbpsf/optics.py
```diff
@@ -1184,12 +1191,18 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
         self.instr_name = instrument.name
 
         # work out which name to index into the CV results with, if for NIRCam
+        is_nrc_coron = False
```
mperrin: Similarly, curious if you can explain the reasoning for moving this up here - I would think that the `is_nrc_coron` variable should be defined under the ` if instrument.name == 'NIRCam':` block?

webbpsf/optics.py
```diff
@@ -1469,19 +1478,27 @@ def __init__(self, instrument, **kwargs):
...
-        cf_scale = -1.09746e7 # convert from mm defocus to meters RMS wfe
-        sw_cf = np.array([-5.169185169, 50.62919436, -201.5444129, 415.9031962,  
-                          -465.9818413, 265.843112, -59.64330811]) / cf_scale
-        lw_cf = np.array([0.175718713, -1.100964635, 0.986462016, 1.641692934]) / cf_scale
-        self.fm_short = np.poly1d(sw_cf)
-        self.fm_long  = np.poly1d(lw_cf)
+        defocus_to_rmswfe = -1.09746e7 # convert from mm defocus to meters (WFE)
+        sw_focus_cf = np.array([-5.169185169, 50.62919436, -201.5444129, 415.9031962,  
+                                -465.9818413, 265.843112, -59.64330811]) / defocus_to_rmswfe
+        lw_focus_cf = np.array([0.175718713, -1.100964635, 0.986462016, 1.641692934]) / defocus_to_rmswfe
```
mperrin: To double check - no changes in focus model coefficients here, just reformatting a bit? (Looks the same to me by eye but I'm not great at diff'ing big piles of digits in my head. )

webbpsf/optics.py
```diff
@@ -1184,12 +1191,18 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
...
         if instrument.name == 'NIRCam':
             channel = instrument.channel[0].upper()
             lookup_name = "NIRCam{channel}W{module}".format(
                 channel=channel,
                 module=instrument.module
             )
+            # Check for coronagraphy
+            pupil_mask = self.instrument._pupil_mask
+            if (pupil_mask is not None) and ( ('LYOT' in pupil_mask.upper()) or ('MASK' in pupil_mask.upper()) ):
+                is_nrc_coron = True
```
mperrin: [optional change] 
This can be simplified to: 
`is_nrc_coron = (pupil_mask is not None) and ( ('LYOT' in pupil_mask.upper()) or ('MASK' in pupil_mask.upper()) )`

webbpsf/optics.py
```diff
@@ -1493,9 +1510,31 @@ def __init__(self, instrument, **kwargs):
...
+        
+        if isinstance(wave, poppy.Wavefront):
+            wavelength = wave.wavelength
+        else:
+            wave = poppy.Wavefront(wavelength=float(wave))
+            wavelength = wave.wavelength
+
+        # Check for coronagraphy
+        pupil_mask = self.instrument._pupil_mask
+        if (pupil_mask is not None) and ( ('LYOT' in pupil_mask.upper()) or ('MASK' in pupil_mask.upper()) ):
```
mperrin: [optional change]
Perhaps let's rename this `is_nrc_coron` for consistency with the same variable as above. And again can be defined  in a one-liner:
`is_nrc_coron = (pupil_mask is not None) and ( ('LYOT' in pupil_mask.upper()) or ('MASK' in pupil_mask.upper()) )`

webbpsf/optics.py
```diff
@@ -1469,19 +1478,27 @@ def __init__(self, instrument, **kwargs):
...
-        self.fm_short = np.poly1d(sw_cf)
-        self.fm_long  = np.poly1d(lw_cf)
+        defocus_to_rmswfe = -1.09746e7 # convert from mm defocus to meters (WFE)
+        sw_focus_cf = np.array([-5.169185169, 50.62919436, -201.5444129, 415.9031962,  
+                                -465.9818413, 265.843112, -59.64330811]) / defocus_to_rmswfe
+        lw_focus_cf = np.array([0.175718713, -1.100964635, 0.986462016, 1.641692934]) / defocus_to_rmswfe
+        self.fm_short = np.poly1d(sw_focus_cf)
+        self.fm_long  = np.poly1d(lw_focus_cf)
+        
+        # Coronagraphic tilt offset model
```
mperrin: [change]
Can you please add here a brief comment describing where these models come from, and how they were derived? The text about deriving from Zemax in the comment from the PR would suffice.

mperrin:
Hi @JarronL - are you available to work on this PR this week, or should we punt until the future (because you're working on HST proposals or otherwise busy this week?) Either way fine with me; I was just reminded about this open PR in a recent conversation and wanted to check in with you about it. 

One particular practical matter: Due to the "partial freeze" of public tools in the run up towards the Cycle 1 deadline, we're trying to leave the `master` branch unchanged on this for now, and instead working on `develop`. Can you please rebase this onto the develop branch, and change the PR to point towards develop as the base rather than master? Let me know if you need tips on how to do any of that and I can advise on the necessary git-fu.

JarronL:
@mperrin Yep, I'll take a look over the next couple of days and throw it in the develop branch instead.

webbpsf/optics.py
JarronL: If `is_nrc_coron` is defined only within the `if instrument.name == 'NIRCam':` block, then the variable is undefined for other instruments. But there are instances where it checks this variable no matter what instrument is in use.

webbpsf/optics.py
JarronL: During development, I ran into an issue with type definitions. For instance, `wavelength` is sometimes an astropy unit if `wavelength = wave.wavelength`, but the print and debug statements were expecting a float. So the change that I implemented always forces this by explicitly setting `wavelength = wave.wavelength` and modifying the print statements to use the value. I also wanted to make this consistent throughout the document, so that's why you see the change multiple times.

webbpsf/optics.py
JarronL: Correct, I'm just changing variable names to be more explicit about what the coefficients refer to.

JarronL:
Closed and moved to develop branch (PR #347) along with above requested changes and additions.

PR #339, opened 2020-02-17
improve reproducibility of thermal slew SM model. Fixes #338

### LINKED ISSUE #338: inconsistent behavior with thermal slew model when called repeatedly
The behavior of the `opd.thermal_slew` function is not well defined if you call it multiple times in a row. Is it supposed to reset back to the original parameters and perform a calculation of a single slew? Or is it supposed to add on another slew, and model that subsequent to whatever wavefront state was previously the case? This should be made clearer in the documentation. 

In most ways the code is set up to reset back to the original OPD and then calculate a slew on top of that (see, in `update_opd` it first resets back to the input OPD then computes perturbations). However the secondary mirror defocus global hexike model is additive (see the code in `update_opd` lower down, where it changes the value of `_global_hexike_coeffs` based on the SM slew model. That's a persistent change that lasts over multiple function calls. It would be more consistent if we changed that to NOT persistently change the values in that array.

mperrin:
Several small changes to the OTE linear model to fix spacetelescope/webbpsf#338 
 - Clarify in doc string of `thermal_slew` that multiple calls in a row to this function aren't cumulative, but rather it resets back to the starting point OPD each time. 
 - Fix a small bug in the `update_opd` function, which prevented the above from being completely true. See note in spacetelescope/webbpsf#338, about how the SM defocus would get repeatedly added into the global hexikes array. This resulted in different results if you called the function multiple times in a row with the same parameters. 
 - Re-enable the global zernikes option as well as the global hexikes option (so that the `move_global_zernikes` function will actually work)
 - Add unit test for the fix of the SM defocus bug.

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -1688,7 +1688,7 @@ class OTE_Linear_Model_WSS(OPD):
             self.update_opd(display=display)
 
     def move_global_zernikes(self, zvector, unit='micron',
-                             absolute=False):
+                             absolute=False, delay_update=False, display=False):
         """ Add one or more aberrations specified arbitrarily as Zernike polynomials.
         This assumes no particular physics for the mirror motions, and allows adding
         any arbitrary WFE.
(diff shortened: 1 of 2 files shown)


kjbrooks:
Looks good! The only thing that I think might be missing is a test of `_apply_global_zernikes()` since that is never tested. If you have a small test that you can write for it and to make sure it is working well with `_apply_global_hexikes()` then I think this would be more complete.

mperrin:
@kjbrooks thanks for the review. Good suggestion to add that test, but given other more urgent priorities for WTP5 I'm going to leave that for another time. I recorded that as a new issue spacetelescope/stpsf#24 and will go ahead and merge this now.

PR #343, opened 2020-03-03
Update simulated OTE notebook

### LINKED ISSUE #342: Simulated OTE Mirror Move Demo.ipynb looking for webpsf-data file that doesn't exist
@obi-wan76 notified me that the current output of the `Simulated OTE Mirror Move Demo.ipynb` shows an error that clearly has to do with my account so in trying to run a clean version of the notebook I found that under the JSC OTIS Test Configurations section, `webbpsf` is trying to find the file `jwst_pupil_JSC_OTIS_Cryo.fits` in `webbpsf-data`, the file seems to not exist in `webbpsf-data`.

kjbrooks:
Updated Simulated OTE Mirror Move Demo notebook to make sure it was a clean run with no errors and removed unnecessary JSC OTIS testing section.

Closes #342


mperrin (approved):
This looks good, thanks!

One optional suggestion - I was briefly confused why the segment PSFs were different in some of the figures, comparing this version of the notebook to the one currently in the repository. Took me a minute or two to realize/remember that the notebook is doing random deployments. We could set the random number generator seed at the start of the notebook to ensure it's consistently reproducible. I've been doing that in my other notebooks recently that use random states and it is definitely helpful in my opinion. That would just be a call to `np.random.seed()`  near the top of the notebook. What do you think, worth adding that and re-running? Your call.

kjbrooks:
I have never used `np.random.seed()` so this is new to me. I tried a test run of this with `np.random.seed(0)` at the top of the notebook and also before any command using an `np.random` call, just in case, but I did not see a change in the behavior. I still found that everything is repeatable until the last two sections of the notebook and those are still random. Since I am not exactly sure what I am supposed to be seeing/doing for this change, I don't think I am the best person to implement it.

mperrin:
OK - I'll go ahead and merge this as-is then. Just adding that call to set the random seed _should_ result in consistent behavior after that. Not sure why it isn't... But that can wait for another time!

PR #344, opened 2020-03-04
Flipped the sign on all of the influence function matrix elements to match WSS

Skyhawk172:
We noticed that the WAS-generated SUR to move segments "up" on the detector had the opposite effect in WebbPSF. Flipped the sign of the IFM elements, at least for now and until we better understand this issue.


mperrin:
(dealt with some git fuss since `master` and `develop` had diverged slightly. Merged the necessary commits into develop directly.)

mperrin (approved):

PR #347, opened 2020-03-06
NIRCam Coronagraph PSF Dispersion

JarronL:
Apply wavelength dependent offsets for NIRCam coronagraphic PSFs. This is due to the dispersion from the optical wedge in the coronagraphic pupil masks. Primarily affects the LW channel (approximately 0.015 mm/um dispersion). SW channel is almost a factor of 10 smaller (fairly negligible), but we include it for completeness. The PSF shift is implemented by modifying the tilt Zernike coefficient (Z3) within the NIRCamFieldAndWavelengthDependentAberration function. Values have been determined using the Zernike offsets as reported in the NIRCam Zemax models. Center reference positions correspond to the NIRCam target acquisition filters (3.35um for LW and 2.1um for SW). Implementation is similar to the NIRCam Zernike and focus updates from #283.

Additional minor changes to `optics.py`:
- Ensure `wave` is a `poppy.Wavefront` object and `wavelength` is generally an `astropy.unit `quantity.
- Create a convenience variable `is_nrc_coron` rather than awkwardly checking NIRCam pupil string values multiple times
- Zernike values stored in FITS header are in units of [m], not [nm]
- Added a unit test to check wavelength dependent PSF offset

This pull request replaces PR #335 in the master branch.

### DIFF
--- a/webbpsf/tests/test_nircam.py
+++ b/webbpsf/tests/test_nircam.py
@@ -421,4 +421,4 @@ def test_nircam_coron_wfe_offset(fov_pix=15, oversample=2, fit_gaussian=True):
     assert np.allclose( diff_25_33, 0.016, rtol=rtol), "PSF shift between {:.2f} and {:.2f} um of {:.3f} mm does not match expected value (~0.016 mm).".format(warr[1], warr[0], diff_25_33)
     # Difference from 3.3 to 5.0 um should be ~0.030mm
     diff_50_33 = np.abs(yloc[2] - yloc[1])
-    assert np.allclose( diff_50_33, 0.032, rtol=rtol), "PSF shift between {:.2f} and {:.2f} um of {:.3f} mm does not match expected value (~0.032 mm).".format(warr[1], warr[0], diff_25_33)
+    assert np.allclose( diff_50_33, 0.032, rtol=rtol), "PSF shift between {:.2f} and {:.2f} um of {:.3f} mm does not match expected value (~0.032 mm).".format(warr[1], warr[2], diff_50_33)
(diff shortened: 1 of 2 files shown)


webbpsf/tests/test_nircam.py
```diff
@@ -357,3 +357,48 @@ def test_ways_to_specify_weak_lenses():
...
+        fit_g = fitting.LevMarLSQFitter()
+        g = fit_g(g_init, xvals, yvals)
+        yloc.append(g.mean.value)
+    yloc = np.array(yloc)
+
+    # Difference values should be greater than 0.010mm
+    # Should be along the lines of 0.013mm between wave=3.23 and 2.5um, 
+    # and 0.031mm between wave=3.23 and 5.0um.
+    assert np.abs(yloc[0] - yloc[1]) > 0.01, "Expected PSF shift between {:.2f} and {:.2f} um is too small".format(warr[1], warr[0])
+    assert np.abs(yloc[2] - yloc[1]) > 0.01, "Expected PSF shift between {:.2f} and {:.2f} um is too small".format(warr[1], warr[2])
```
mperrin: @JarronL - I see the checks here in this unit test checks that the PSF shift is above some minimum, but doesn't test that it's below some maximum, or that it's near a desired value, etc. 

Given the comment text that gives specific values we expect, how about testing against those instead: 
```python

assert np.allclose( yloc[1] - yloc[0], 0.013, rtol=0.1), "Expected PSF shift between {:.2f} and {:.2f} um does not match expected value".format(warr[1], warr[0])

assert np.allclose( yloc[2] - yloc[1], 0.031, rtol=0.1), "Expected PSF shift between {:.2f} and {:.2f} um does not match expected value".format(warr[2], warr[1])
```
That checks that the output shifts match the expected values within 10%. Tolerance could be adjusted, but that seems like a decent starting threshold.

JarronL: Ok, added your suggestion along with the option to disable the Gaussian fitting in case astropy modules don't exist somewhere along the line of testing and installing.

mperrin: @JarronL great thanks. FYI I think the disable fitting option isn't needed - astropy is a strict requirement of poppy, so should always be present anywhere webbpsf is installed. This is fine as is; I'm just mentioning this as an FYI.

PR #351, opened 2020-04-22
Sensible NIRCam SIAF aperture selection

JarronL:
Select an appropriate NIRCam SIAF aperture for coronagraphic image masks and Lyot pupil stops. An auto-select option can be disabled using the `auto_aperturename` attribute. The aperture name gets updated any time there is a change to the detector, pupil, or image mask. Defined an `aperturename` property attribute to help facilitate this. Includes a unit test function to check that `aperturename` is updated correctly for different NIRCam modes.

Should explore moving these up to the parent class?

Also added a `module` setter mimicking the behavior of the `channel` setter.

### DIFF
--- a/webbpsf/tests/test_nircam.py
+++ b/webbpsf/tests/test_nircam.py
@@ -422,3 +422,75 @@ def test_nircam_coron_wfe_offset(fov_pix=15, oversample=2, fit_gaussian=True):
     # Difference from 3.3 to 5.0 um should be ~0.030mm
     diff_50_33 = np.abs(yloc[2] - yloc[1])
     assert np.allclose( diff_50_33, 0.032, rtol=rtol), "PSF shift between {:.2f} and {:.2f} um of {:.3f} mm does not match expected value (~0.032 mm).".format(warr[1], warr[2], diff_50_33)
+
+def test_nircam_auto_apname():
+    """
+    Test that correct apertures are chosen depending on channel, module, detector, mode, etc.
+    """
+    import pysiaf 
+
+    nc = webbpsf_core.NIRCam()
+
+    nc.filter='F200W'
+    assert nc.apname == 'NRCA1_FULL'
+
+    # auto switch to long
+    nc.filter = 'F444W'
+    assert nc.apname == 'NRCA5_FULL'
+
+    # and it can switch back to short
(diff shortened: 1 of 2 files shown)


mperrin:
> Should explore moving these up to the parent class?

I think we should at least move up the `apname` to the parent class. And perhaps expand to `aperture_name` for clarity?  

For many of the other instruments a very simple implementation would suffice - for NIRISS and FGS it's pretty trivial since we only care about one full frame aperture per detector. For NIRSpec it can key off the selected image plane mask in some instances. MIRI would want a similar implementation like NIRCam, depending on coronagraph optic selections.

mperrin:
Hi Jarron - I think the one thing this needs before being merged is a unit test to demonstrate correct functionality for a few example cases. That could look a lot like the `test_nircam_auto_pixelscale` function for example. Do you think you could put something like that together?  Once we have a test to go with this we can merge this PR, and I'll leave for later to refactor the implementation consistently across all the SIs.

JarronL:
> Hi Jarron - I think the one thing this needs before being merged is a unit test to demonstrate correct functionality for a few example cases. That could look a lot like the `test_nircam_auto_pixelscale` function for example. Do you think you could put something like that together? Once we have a test to go with this we can merge this PR, and I'll leave for later to refactor the implementation consistently across all the SIs.

Sure thing. I'll take a look at implementing this over the next week.

mperrin:
Hi @JarronL  - any chance you can give us an updated ETA for when you might be able to work on this? If it's not this week that's fine, but we'll likely go ahead and do the 0.9.1 release without this in that case. But hopefully this might not take long to just add a unit test or two, and then we can merge this? Up to you.

JarronL:
@mperrin, I will try to get an update for you tomorrow.

mperrin (approved):
@JarronL Thanks this looks great. Much appreciated!

One small thought on naming. `apname` may not be as self-explanatory and clear to others as we might like. Would you be OK if I merge this PR and then subsequently rename to `aperturename` throughout?

JarronL:
I just went ahead and did the `apname` to `aperturename` change. Been looking for an excuse to try out VSCode's built-in variable name change; worked like a charm. **Achievement Unlocked**

PR #353, opened 2020-04-23
minimum supported python version is now 3.6, not 3.5

mperrin:
As discussed in email - we're incrementing to min supported version to 3.6, consistent with current astropy and conda.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ matrix:
         - env: TOXENV='py37-astropydev-test'
 
         # Try minimum supported versions
-        - env: TOXENV='py35-legacy35-test'
         - env: TOXENV='py36-legacy36-test'
 
         # Try released POPPY
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -105,7 +105,7 @@ Software Requirements
 
 See `the environment.yml specification file <https://github.com/spacetelescope/webbpsf/blob/master/environment.yml>`_ for the required package dependencies. 
 
-**Required Python version**: WebbPSF 0.8 and above require Python 3.5 or higher.
+**Required Python version**: WebbPSF 0.9.1 and above require Python 3.6 or higher.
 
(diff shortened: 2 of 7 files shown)


shanosborne (commented):
There's a few other places that 3.5 is defined that I found that should be changed: `webbpsf/__init__.py`, `environment.yml`, and `installation.rst`. We can also delete all mentions of py35 in `tox.ini`, which would be removing all references to 'legacy35' from the `[testenv]` section and all references to 'py35' in the `envlist` definition at the top of the file.

mperrin:
ah great. I knew that seemed too easy. I'll add those fixes shortly

shanosborne:
@mperrin I've been trying and I can't seem to re-create the error that travis is showing. I see that the same error happened in the PR where you include python 3.8 in travis, but it's a different test that fails in that PR. So somehow there's occasionally an issue with getting the list of the PRD versions. In the next few days I can try to add a patch to pysiaf that does a better job of handling any problems in that part of the code.

mperrin:
@shanosborne I wonder if some of this intermittent weirdness is related to the GitHub server problems earlier today? It seems mostly working now but was down a bunch of the morning. In any case we'll see if this latest build passes and if so, hopefully no need to worry too much tracking down intermittent flakiness.

shanosborne:
Oh I did hear about that! That would make sense since the code that's was having a problem was looking at a github link. Okay so then maybe for the time being everything is fine - it looks like the previously failing test is now running and didn't fail on import

mperrin:
OK the Travis test is passing now, but for some reason the codecov coverage report is failing. Based on my examination of the codecov reports in more detail, it looks like the "change" is supposedly a loss of overage in 3 lines, in a completely unrelated part of the code which have nothing to do with this PR but may be related to PR #349 that I merged yesterday. 

As far as I can tell this is some testing system confusion / glitch, and I don't want to pay it too much attention at this time.  @shanosborne would you agree with that? I'm inclined to just merge this now as is.

shanosborne:
@mperrin I agree - I don't think it's important enough to hold up this PR, I say go ahead and merge it

PR #356, opened 2020-05-26
Fix installation issues with otelm directory

shanosborne:
Original description of the issue from @kjbrooks 

> I am trying to set up mirage and the ote notebooks require the latest on webbpsf develop. Mirage currently installs webbpsf via pip (pypi) which means that it is installing develop. Since I need develop, I uninstalled webbpsf and reinstalled it using the command: pip install git+https://github.com/spacetelescope/webbpsf@develop . This worked, except, for some reason, I don’t get the otelm directory. First question: do you know why? Secondly, Marshall said that y’all are talking about a 0.9.1 release so you might need to update the setup.py to address whatever issue I am getting, which is why I am bugging you.

The issue was that there was no `__init__.py` file in in `otelm/` directory which meant that it wasn't getting seen and installed with the rest of the module. I added the file and checked and that command now installs `webbpsf` including the `otelm/` directory.

I also added an `__init__.py` file to the `tests/surs/` directory because it was also missing one. I can remove this but figured it wouldn't hurt to include this in the module and may break things if not.


mperrin (approved):
Thanks for this fix - looks good. I think the same end goal can also be accomplished in another way by specifying package data files, or in theory by the MANIFEST.IN file in the repo top directory; but clearly that way isn't working as it should so this seems better. 

(Sorry for not reviewing this sooner; it slipped through the cracks by accident...)

PR #360, opened 2020-06-08
Generalize aperturename across all SIs

mperrin:
Refactor the `aperturename` functionality from #351 to generalize across instruments. 

- [x] Provide `inst.aperturename` attribute across all SpaceTelescopeInstrument subclasses
- [x] Implement default aperture names for all instruments as the full frame apertures (which is what we we doing before behind the scenes, but now it's explicitly visible as the attribute)
- [x] Implement `auto_aperturename` functionality for MIRI, selecting based on corona graphic masks, so roughly consistent with how it works for NIRCam.   

@JarronL would you be willing to take a look at this please? I don't seem to be able to officially tag you as a reviewer, perhaps because you're external to the spacetelescope organization on here... Let me see if I can find a work around for that...

### DIFF
--- a/docs/jwst.rst
+++ b/docs/jwst.rst
@@ -34,6 +34,11 @@ All classes share some common attributes:
    on that detector for the center location in any calculated output PSF.
    Note that the ``detector_position`` value should be
    specified using the order (X,Y).
+ * The ``aperturename`` attribute provides the `SIAF <https://pysiaf.readthedocs.io>`_ aperture name 
+   used for transforming between detector position and instrument field of view on the sky. By default
+   this will be a full-frame aperture for the currently-selected detector, but you may select any
+   subarray aperture or other aperture named in the SIAF for that instrument. For NIRCam and MIRI,
+   the aperturename can (optionally) automatically update for coronagraphic subarrays.
 
 .. warning::
 
@@ -114,7 +119,6 @@ just set the desired detector and the channel and module are inferred
 automatically.
 
 
-
 The choice of ``filter`` also impacts the channel selection: If you choose a
 long-wavelength filter such as F460M, then the detector will automatically
 switch to the long-wave detector for the current channel. For example, if the
@@ -144,6 +148,7 @@ behavior on filter selection can be disabled by setting ``nircam.auto_channel = 
     filter or detector attribute whenever you want to toggle between SW or LW channels.
 
 
+
 Coronagraph Masks
 ------------------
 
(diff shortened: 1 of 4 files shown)


mperrin:
We probably should also update the documentation to describe this functionality too...

docs/jwst.rst
```diff
@@ -162,6 +167,13 @@ can still be used, but the same masks can also be referred to as "MASKRND" and
 "MASKSWB" or "MASKLWB", the nomenclature that was eventually adopted for use in
 APT and other JWST documentation. Both ways work and will continue to do so.
 
+The NIRCam class can automatically switch its ``aperturename`` attribute when a
+coronagraphic mask is selected, to select the aperturename for the appropriate
+coronagraphic subarray.  The detector reference pixel location will also update
+to the center of the coronagraphic subarray. This behavior on image mask or
+pupil mask selection can be disabled by setting ``nircam.auto_aperturename =
+False``.
+
```
JarronL, line 178: Should also mention that the aperture name will always change if the `detector` property (attribute?) changes.

webbpsf/webbpsf_core.py
```diff
@@ -856,7 +899,7 @@ def _get_fits_header(self, result, options):
                                           "[arcmin] Det. pos. in telescope V2,V3 coord sys"), after=True)
         result[0].header.insert("DET_V2", ('DET_V3', v2v3pos[1].value,
                                            "[arcmin] Det. pos. in telescope V2,V3 coord sys"), after=True)
-        result[0].header["APERNAME"] = (self._detectors[self._detector], "SIAF aperture name")
+        hdulist[0].header["APERNAME"] = (self._aperturename, "SIAF aperture name")
```
JarronL: I updated this as part of the generalization, because the GitHub review wouldn't let me leave a comment here. I simply copy/pasted from the NIRCam class, but forgot to change the variable name to `result` rather than  `hdulist`.

webbpsf/webbpsf_core.py
```diff
@@ -1890,8 +1956,6 @@ def _get_fits_header(self, hdulist, options):
         hdulist[0].header['CHANNEL'] = ('Short' if self.channel == 'short' else 'Long', 'NIRCam channel: long or short')
         # filter, pupil added by calc_psf header code
         hdulist[0].header['PILIN'] = ('False', 'Pupil imaging lens in optical path: T/F')
-        # Update APERNAME
-        hdulist[0].header["APERNAME"] = (self._aperturename, "SIAF aperture name")
```
JarronL, line 1894: This has been moved to the JWInstrument class

JarronL (changes requested):
This looks pretty good to me. That was a quick turnaround! Just a couple minor comments, one of which was my fault.

webbpsf/webbpsf_core.py
JarronL: Nevermind about the change request. I just went ahead and fixed my own typo.

JarronL:
Sorry, not used to the review features. I modified `webbpsf_core.py` to move the APERNAME header information to the ``JWInstrument`` class, but screwed it up due to copy/paste error, but then fixed it. 

The only thing I would add is a little detail in `jwst.rst` to say that the aperture name will always change if the detector property changes. Everything else looks good to me at this point!

mperrin:
@JarronL I added a sentence in the docs as you suggested. 
As for "not used to the review features" first I would say it's actually even more helpful that you just pushed in your edits directly so thanks much for that. Secondly I'd be happy to talk you through any details of how to use the review tools on here if you want. 

(IMO the code review tools on GitHub are the actual killer app for git; that's what makes it worth it despite the sometimes pain and confusion of git's command line app and various deeper arcane compexities...)

JarronL (approved):

PR #361, opened 2020-06-09
Allow GUIDER1 as synonym for FGS1, and likewise for 2

mperrin:
Allows `fgs.detector='GUIDER1'` as a synonym for `fgs.detector='FGS1'`, and so on. 

Fairly trivial PR / low priority. I think we discussed this on Slack recently @shanosborne...  Intended so that the detector name style used in DMS is also supported in WebbPSF. For now the output FITS headers and so on are unchanged and still say e.g. "FGS1", but let me know if you think it would be better to swap those to GUIDER instead?

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -2231,6 +2231,9 @@ class FGS(JWInstrument):
 
     Not a lot to see here, folks: There are no selectable options, just a great big detector-wide bandpass
     and two detectors.
+
+    The detectors are named as FGS1, FGS2 but may synonymously also be referred to as
+    GUIDER1, GUIDER2 for compatibility with DMS convention
     """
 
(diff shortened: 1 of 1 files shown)


shanosborne (commented):
Hmm, looks like the tests are failing because something isn't passing through right with the `_detector_geom_info` attribute for FGS

mperrin:
Yeah I left out a necessary line in the subclass function when overriding the property setter. Fixed now so the tests should now pass...

shanosborne (approved):
Looks good to me. All the tests are passing and I checked that setting detector to `FGS1` and `GUIDER1` (and `FGS2` and `GUIDER2`) produces the same PSF. So I think this is set. And I agree with your original comment - I think leaving the header saying "FGS1" is fine

PR #362, opened 2020-06-09
Set origin='lower' in display_opd()

shanosborne:
Following a question from Brian Hicks, I added an `origin='lower'` argument to the `display_opd` function in order to keep the labeling/struts/segments all straight. See the before and after pictures below which were created with the lines 
```
nc, ote = webbpsf.enable_adjustable_ote(nc) 
ote.move_seg_local('A1', roc=.5) 
ote.display_opd()
```

Before the change was applied:
opd_before
After this change was applied: 
opd_after

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -346,7 +346,7 @@ class OPD(poppy.FITSOpticalElement):
         if ax is None:
             ax = plt.gca()
 
-        plot = ax.imshow(self.opd * mask * scalefact, vmin=-vmax, vmax=vmax, cmap=cmap, extent=extent)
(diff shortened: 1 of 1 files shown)


mperrin (approved):
Hi @shanosborne  - Approved, yes. (Sorry, thought I'd approved this already last week). 

Did you happen to look and see if there are any other imshow plot calls in that file which might also need to be updated? Would be worth a 30 s quick check right now just to avoid any similar situations in the future...

mperrin:
CI Test says it fail because the plot command doesn't get tested, but that's status quo for _all_ the display and plot commands. So that's no showstopper and I'm going to merge this.

PR #366, opened 2020-06-19
add automation for PyPI from Travis

mperrin:
Add CI automation for PyPI deployment on releases, following https://docs.travis-ci.com/user/deployment/pypi/


shanosborne (approved):
The travis files matches what's in poppy so that looks good. 

The "Prerequisites",  "Releasing new data packages", and "Finishing the Release" sections of release.rst seem to be in block quote format. Im not sure if that was intentional but it is different from the rest of the document.

mperrin:
Those sections aren't block quotes, those are numbered lists. Same as what was there before so I don't see what you think is different from the rest of the document? It seems to render correctly in the sphinx build in my local tests...

shanosborne:
Interesting - it renders as a block quote when shown in the github interface (i went to the changed file, clicked the 3 dots, and chose "view file"). But if it looks good with sphinx then that's fine!

mperrin:
Maybe some subtle issue having to do with whether there's any spaces before the # or not? 🤷

PR #375, opened 2020-07-14
Update thermal slew scaling

JarronL:
Previous thermal slew scaling equation did not have cumulative totals equal to 1.0 if broken down into smaller jumps (e.g., -5 to 20 deg, then 20 to 45 deg would have a total scaling greater than -5 to 45 deg). Updated equation fixes this issue by taking into account the sin of specific start and end angles rather than just delta angles.

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -1855,7 +1855,9 @@ class OTE_Linear_Model_WSS(OPD):
             segment name will fits 9 Hexikes to that segment
         """
         if not self.scaling:
-            scaling = np.sin(np.radians(self.end_angle) - np.radians(self.start_angle)) / np.sin(np.radians(45.) - np.radians(-5.))
+            num = np.sin(np.radians(self.end_angle)) - np.sin(np.radians(self.start_angle))
+            den = np.sin(np.radians(45.)) - np.sin(np.radians(-5.))
+            scaling = num / den
 
         else:
             scaling = self.scaling
--- a/webbpsf/tests/test_opds.py
+++ b/webbpsf/tests/test_opds.py
@@ -46,11 +46,12 @@ GLOBAL_FOCUS2 = [GLOBAL_FOCUS[0] * SCALING_FACTOR]
 COEFFS_A1 = np.array([-3.52633363e-09, -2.90050902e-09, 1.25432196e-09, -7.43319098e-12,
                       -5.82462948e-11, -1.27115922e-10, -1.91541104e-12, 3.64760396e-11,
                       4.97176630e-13])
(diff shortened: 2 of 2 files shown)


mperrin:
Hi @JarronL, finally after a bit of vacation I'm trying to catch up on a few things and was reminded of this PR. 

I'll take a look through the code, but first order of business is I can see the unit tests have failed. Have you had a chance to look yet into which test(s) are failing, and what might need to be done to fix that?

mperrin:
OK, looking at that equation the revision makes sense to me, and I see that it has the desired property of cumulative totals equal to 1. So I suspect we need to update one or more of the test cases. I have to admit I don't recall the details of how we set up the tests for this part of the code. It looks like we have a bunch of hard-coded numbers in the test file. @kjbrooks what do you recall of this?  I think we may have just implemented these tests to, in effect, check for any changes in the behavior of the thermal slew model, without particularly checking that the results are rigorously correct. So maybe we just declare that the new results are more correct, and update the table of numbers accordingly? 

Separately we could add a test for the property that a given total slew still gives the same total amplitude even if broken Ito smaller jumps. That would be something like 


```
    otelm = webbpsf.opds.OTE_Linear_Model_WSS()
    otelm.thermal_slew(delta_time, start_angle, end_angle, scaling, case='EOL')
    coeffs_onebigslew = otelm._get_thermal_slew_coeffs(segid=seg)

    otelm = webbpsf.opds.OTE_Linear_Model_WSS()
    otelm.thermal_slew(delta_time, start_angle, middle_angle, scaling, case='EOL')
    coeffs_slewpart1 = otelm._get_thermal_slew_coeffs(segid=seg)
    otelm.thermal_slew(delta_time, middle, middle_angle, scaling, case='EOL')
    coeffs_slewpart2 = otelm._get_thermal_slew_coeffs(segid=seg)

    assert(np.allclose(coeffs_onebigslew, coeffs_slewpart1 + coeffs_slewpart2)), "should get same total coefficients for one big slew or if broken into two parts"

```
(That test code not yet actually tested!)

kjbrooks:
I can't really remember what my thought process was on the tests. I think I was just trying to ensure we would get a result that we would expect, but I agree that that might not be the most complete approach. I also agree that the easiest way to get those test working would be just to re-run with the new code, declare that output to be a more truthful truth, and update those values. 

I think adding the test @mperrin mentioned above would be a good test for what this PR is trying to fix. 

Once the tests are updated, then this PR looks good to merge to me.

JarronL:
Okay, I can have a look at the unit tests and update. Not sure why I didn't get notified when these failed...

mperrin (approved):
Looks good to me. Thanks for adding the test @JarronL

PR #377, opened 2020-07-24
avoid reloading SIAF multiple times, for small runtime speed improvement

mperrin:
Small runtime efficiency improvement: Just load the SIAF information once, rather than reloading it each time when adjusting detector geometry information for a WebbPSF instrument instance. 

Motivation: I noticed the runtime for instantiating a WebbPSF object (e.g. just `nrc= webbpsf.NIRCam()` ) is almost entirely dominated by the time it takes to read the SIAF information. This takes up to ~0.2-0.3 s, depending on instrument (NIRCam is slower, not surprisingly). I also noticed that we read in the SIAF information multiple times. Not that it is a big deal if it takes half a second or so to create a NIRCam instance, or a quarter second every time to change to a different aperture, but it's very easy to make this more efficient by reading the SIAF just once and storing it as an object attribute. 

With this change the runtime to create any WebbPSF object instance drops by about 50%, and the time to change to a different aperture becomes << 1 ms instead of ~0.2s.

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -764,6 +764,8 @@ class JWInstrument(SpaceTelescopeInstrument):
     def __init__(self, *args, **kwargs):
         super(JWInstrument, self).__init__(*args, **kwargs)
 
+        self.siaf = pysiaf.Siaf(self.name)
+
         opd_path = os.path.join(self._datapath, 'OPD')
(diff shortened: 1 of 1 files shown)


shanosborne (approved):
Good idea! I checked that there are no `Siaf` class calls anywhere else in the file that got missed, and all the tests are passing so I'd say this is good to go!

PR #378, opened 2020-07-24
Toward better OTE field dependence modeling: use OTE linear model class by default 

mperrin:
As discussed in slack: We want to enhance the OTE wavefront model to include models of field dependence over the OTE focal plane (both the nominal as-built WFE in the case of a perfectly aligned telescope, and the potential excess WFE from MIMF aberrations if the SM is misaligned). 

- [x] As a practical note this is going to be easier to implement if all the linear models for the OTE are in one class. This PR refactors the JWST models to by default automatically cast the wavefront error read from disk (the predicted or requirements WFE files) into the OTE_Linear_Model_WSS class, instead of a plain FITSOpticalElement. This happens behind the scenes with no need for any change in user code. 
The exceptional to this is, the OTE linear model does not yet support anything other than the default npix=1024 pupil array size, so if any other pupil size is selected for the input model we fall back to the plain FITSOpticalElement. This limitation would be nice to remove but is out of scope for this PR. 
- [x] I add a placeholder method OTE_Linear_Model_WSS._apply_field_dependence_model(), which currently does nothing yet but which we can & will implement with the necessary Zernike calculations. 
- [x] I also fixed a couple unrelated minor issues that I noticed while testing this: a change in the pyfftw version string that was causing problems for a unit test, and improving a log string to show a reasonable number of significant figures. (Unrelated to the core of this PR but not worth the effort to split into a separate PR)

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -556,7 +556,7 @@ class OTE_Linear_Model_Elliott(OPD):
 
     """
 
-    def __init__(self, opd=None, opd_index=0, transmission=None, rm_ptt=False, zero=False):
+    def __init__(self, opd=None, opd_index=0, transmission=None, rm_ptt=False, rm_piston=False, zero=False):
         """
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


marcio-melendez (approved):
It looks good to me with the place holder for apply_field_dependence_model to change the OPD

PR #383, opened 2020-09-24
Updated Modes for the WFIRST/Roman CGI

kian1377:
This pull request changes the wfirst.py file such that the CGI class includes the updated SPC modes for F730 and F825. It now refers to the older modes as deprecated, but the deprecated modes are still functional. In order for the new modes to function, the file structure for webbpsf-data has been changed and the Phase-B data files for the 2 new SPC modes have been added. The webbpsf-data also includes the files for the HLC mode, however, that mode is not yet functional in the webbpsf wfirst code. The updated file structure is shown in the picture below and the full webbpsf-data directory is available on box. 

image

In addition, there are 3 new notebooks:
- Roman_deprecated_tests.ipynb: this file demonstrates the functionality of the now deprecated modes.
- Roman_F730_tests.ipynb: this file demonstrates the functionality of the new mode. 
- Roman_F825_tests.ipynb: this file demonstrates the functionality of the new mode. 

Also, two fits files have been added called proper_spcf730_result.fits and proper_spcf825_result.fits. These contain results of off-axis PSFs produced by PROPER in order to compare the Webbpsf results with those of another model. 

It should be noted that the new modes appear to be incorrect when it comes to the on-axis PSFs, whereas the comparison with the PROPER PSFs for the off-axis case show very high agreement; why this is case is still being looked into. Also, the new modes added do not yet support the implementation of OPD files. 

Lastly, credit should also be given to @Jashcraf and @biancalondra for contributing work to make the updated modes.


webbpsf/wfirst.py
```diff
@@ -557,18 +550,39 @@ def _masked_pupil_path(self):
...
+ ███████████ ██████████  █████████  ███████████     █████████    █████████  ███████████   █████ ███████████  ███████████
+░█░░░███░░░█░░███░░░░░█ ███░░░░░███░█░░░███░░░█    ███░░░░░███  ███░░░░░███░░███░░░░░███ ░░███ ░░███░░░░░███░█░░░███░░░█
+░   ░███  ░  ░███  █ ░ ░███    ░░░ ░   ░███  ░    ░███    ░░░  ███     ░░░  ░███    ░███  ░███  ░███    ░███░   ░███  ░ 
+    ░███     ░██████   ░░█████████     ░███       ░░█████████ ░███          ░██████████   ░███  ░██████████     ░███    
+    ░███     ░███░░█    ░░░░░░░░███    ░███        ░░░░░░░░███░███          ░███░░░░░███  ░███  ░███░░░░░░      ░███    
+    ░███     ░███ ░   █ ███    ░███    ░███        ███    ░███░░███     ███ ░███    ░███  ░███  ░███            ░███    
+    █████    ██████████░░█████████     █████      ░░█████████  ░░█████████  █████   █████ █████ █████           █████   
+   ░░░░░    ░░░░░░░░░░  ░░░░░░░░░     ░░░░░        ░░░░░░░░░    ░░░░░░░░░  ░░░░░   ░░░░░ ░░░░░ ░░░░░           ░░░░░    - Jaren Ashcraft
+                                                                                                                        
+"""
```
mperrin: Remove please :-)

webbpsf/wfirst.py
```diff
@@ -581,20 +595,22 @@ class CGI(WFIRSTInstrument):
...
+    fieldstop_list = ['FS_HLC']
+
+    _mode_table = {#MODE CAMERA  FILTER  APODIZER   FPM             LYOT STOP
+    	'HLC_F575':     ('IMAGER','F575','NONE','HLC_F575_FPM','LS_F575'), # Flight Mode
+        'CHARSPC_F730': ('IFS', 'F730', 'CHARSPC730', 'CHARSPC_F730_BOWTIE', 'LS_F730'), # Flight Mode
+        'DISKSPC_F825': ('IMAGER','F825','DISKSPC825','DISKSPC_F825_ANNULUS','LS_F825'), # Flight Mode
+        'CHARSPC_F660': ('IFS', 'F660', 'CHARSPC', 'CHARSPC_F660_BOWTIE', 'LS30D88'), # Depreciated Mode
+        'CHARSPC_F770': ('IFS', 'F770', 'CHARSPC', 'CHARSPC_F770_BOWTIE', 'LS30D88'), # Depreciated Mode
+        'CHARSPC_F890': ('IFS', 'F890', 'CHARSPC', 'CHARSPC_F890_BOWTIE', 'LS30D88'), # Depreciated Mode
+        'DISKSPC_F721': ('IMAGER','F721','DISKSPC', 'DISKSPC_F721_ANNULUS', 'LS30D88')} # Depreciated Mode
```
mperrin: minor: spelling "deprecated"

webbpsf/wfirst.py
```diff
@@ -643,12 +659,17 @@ def camera(self, value):
...
+                self.pixelscale = 0.040  # Nyquist at 830 nm
+        if value == 'HYBRID':
+            if not hasattr(self, 'fov_arcsec') or not self._override_fov:
+                self.fov_arcsec = 2 * 0.45
+            if not hasattr(self, 'pixelscale') or not self._override_pixelscale:
+                self.pixelscale = 0.020 # Nyquist at 825 nm
         else:  # default to 'IFS'
             if not hasattr(self, 'fov_arcsec') or not self._override_fov:
-                self.fov_arcsec = 2 * 0.82  # 2015 SDT report, Section 3.4.1.1.1:
+                self.fov_arcsec = 2 * 0.55  # 2015 SDT report, Section 3.4.1.1.1:
```
mperrin, line 648: The IFS was descoped; shouldn't we delete rather than updating this IFS section here?

webbpsf/wfirst.py
```diff
@@ -643,12 +659,17 @@ def camera(self, value):
         self._camera = value
         if value == 'IMAGER':
             if not hasattr(self, 'fov_arcsec') or not self._override_fov:
-                self.fov_arcsec = 3.2
+                self.fov_arcsec = 2 * 1.4
             if not hasattr(self, 'pixelscale') or not self._override_pixelscale:
-                self.pixelscale = 0.020  # Nyquist at 465 nm
+                self.pixelscale = 0.040  # Nyquist at 830 nm
+        if value == 'HYBRID':
```
mperrin, line 641: `elif`?

webbpsf/wfirst.py
```diff
@@ -820,19 +876,21 @@ def _get_fits_header(self, result, options):
...
-        result[0].header.set('PUPLDIAM', apodizer_hdr['PUPLDIAM'],
-                             comment='Full apodizer array size, incl padding.')
+        if self._apodizer != 'NONE':
+            result[0].header.set('APODIZER', self.apodizer, comment='Apodizer')
+            result[0].header.set('APODTRAN', os.path.basename(self._apodizer_fname),
+                                 comment='Apodizer transmission')
+            result[0].header.set('PUPLSCAL', apodizer_hdr['PUPLSCAL'],
+                                 comment='Apodizer pixel scale in m/pixel')
+            result[0].header.set('PUPLDIAM', apodizer_hdr['PUPLDIAM'],
+                                 comment='Full apodizer array size, incl padding.')
```
mperrin, line 869: The PUPLSCAL and PUPLDIAM keywords are also written below based on the lyot stop header, so these values will be overwritten. This is an oversight in the existing code, not in your code, but perhaps we should fix it now as part of these improvements. We could use for instance APODSCAL and APODDIAM here.

mperrin (changes requested):
Hi @kian1393, 

Thanks very much for this contribution! I've reviewed it and have a few suggestions and requests. 

- First, you commented about the on-axis PSFs disagreeing with the PROPER reference calculations. Do you expect to be able to investigate and resolve that sometime soon (prior to merging this PR), or do you think we should merge this now as 'good enough'? What level of disagreement do you see between the two cases? Let me know what you think the next steps should be for this, please. 
- The test notebooks need some updates. First, they contain hard-coded paths to e.g. `/groups/douglase/webbpsf-data/CGI/optics`, so they won't run for anyone else. This should be done with relative paths into webbpsf-data. Next, the notebooks start with some modifications to the FITS files on disk.  It seems that this should be done once and the updated (correct keywords etc) FITS files saved as the ones in the WebbPSF-data directory on Box. It needs to be possible to just invoke the model and perform PSF calculations without having to modify the files each time. If this is already the case, then the notebook needs to be updated. Lastly, the notebook filenames and text in them should probably state 'Roman CGI' not just Roman. 
- See below for additional comments/questions on a few parts of the code. 

I'd also like to suggest some additional areas which could benefit from updates, but may be out of scope for this PR: 
- We should consider moving some of the cross-checks of the coronagraphic PSFs to automated tests in `webbpsf/tests/test_wfirst.py` or perhaps a new `webbpsf/tests/test_roman_cgi.py`, not just having the tests in the notebooks. Are you familiar yet with writing automated software tests? If not it's a good skill to learn and I can point you in the right direction if interested. 
- We should also consider updating the documentation pages to describe the Roman CGI support, and link people to the notebooks. Would you be willing to help with that part too? See `docs/wfirst.rst` for instance. 

Thanks again - I'm really glad to see this update, and I think it's 90+% of the way there already, just needs these few bits polished which I expect to be pretty straightforward. Happy to discuss more for next steps with you, @douglase, and others.

webbpsf/wfirst.py
kian1377: This was some pixel art included by Jaren Ashcraft. It helped with scrolling to the section of code being modified and I forgot to delete it in the final uploaded version.

webbpsf/wfirst.py
kian1377: I was not aware of the IFS being descoped. I remember @douglase mentioned something about the IFS but said to leave it for now. I will speak to him for more clarification regarding this.

kian1377:
Hello,

Thank you for the comments. To begin with, I have not had time to further investigate the disagreement in the on-axis PSFs due to academics, however, I am looking to do so in the next couple of weeks. I can add a section showing the on-axis PROPER PSFs similar to how I show the comparison between the off-axis case until it is solved, so at least the disagreement can be seen. 

I can update the notebook as well in order to have the relative paths instead of the hardcoded paths. Regarding the modifications to the fits files, the files already contain the correct keywords and values because of the testing I have been doing, however, I only included that portion of the code such that users can explicitly see what was being done and what values were being set. 

As far as updating the test_wfirst.py file, I do not have experience writing automated software tests, but I am open to learning this in order to update it in a future PR. I am also open to updating the documentation pages, however I am unfamiliar with this as well so please point me in the right direction for these tasks. 

Regards,
Kian Milani

mperrin:
Hi Kian, 

Great, thanks. For now I would encourage you to focus on the investigation of the on-axis PSFs (as time permits after classes and other academic commitments, for sure). 

The testing and documentation we can sort out later. For now, in case you're curious I will point you at this page: https://docs.pytest.org/en/stable/talks.html and we can talk more later. The basic idea is simple: write a function that checks some desired fact is true about the output of your code (for instance, that the output from the CGI model is within say 5% of the output from the PROPER reference calculation), and then the pytest framework will automatically run that test after every revision of the code, and verify that the result is true. In effect, it can be the same sort of check you already do in your notebooks, but made automatic. This approach to software development and validation has proven to be an effective way to find and prevent bugs; it ensures not only does your test code in the notebook run once when you first develop it, it has to continually run correctly after any later edits to the software, or else it'll raise an alert to let you know. It's a very useful tool to have in your tool kit, particularly for reproducible and reliable scientific computing.  But we can sort all that out later, after we've got the models in better agreement for on-axis as well as off-axis calculations. Cheers!

Issue #384, opened 2020-09-28
move_sur function in opds.py needs to apply clocking move 


Skyhawk172:
The call to self.move_seg_local needs to also have the "clocking=update.moves['CLOCK']*sign" argument.

mperrin:
Closed by PR #385

PR #385, opened 2020-09-28
Added line to include clocking from SUR

Skyhawk172:


Skyhawk172:
This is to fix issue https://github.com/spacetelescope/webbpsf/issues/384

Skyhawk172:
It looks like the test is failing because it requires the quick fix to JWXML for which I've also submitted a Pull Request.

mperrin (approved):
Looks good - thanks Charles for both the fix and the improved test

PR #389, opened 2020-11-10
OTE field dependence, part 2: Field Dependence for As-Built Aligned OTE

mperrin:
Models for OTE field dependence based on the Code V models correlated with the as-built OTE. Model developed by Greg Brady, with inputs from Perrin and Telfer.


mperrin:
Ideally this pull request should have included some summary notes here on its content, along with figures demonstrating the output. There's some partial documentation included in the PR content itself (see `docs/jwst.rst`). 

Before this can be merged, we need to create an updated data file release including the additional tables with the field dependence model coefficients. 

This also could use documentation of the outputs (e.g. plots of example WFE vs field) and usage notes. 

Also the unit test code could be improved to actually test correctness of the outputs, in comparison to some expected values for a given representative field point, rather than just checking for nonzero WFE between different field points.

mperrin:
Merged develop; deconflicted.

Added a check such that if the requisite data files are not present, this will issue a warning and continue (rather than an exception that ends the program). This allows better back compatibility temporarily as we roll this out.

shanosborne:
@mperrin I believe I've come up with a fix for the failing tests which you can see in the last commit on this PR https://github.com/shanosborne/webbpsf/pull/4. I'm still playing around, but if you are okay with this fix I can apply it to this branch too.

mperrin:
@shanosborne how come you deleted this branch? This is still not merged so that seems like not at all the right thing to do. I know you made some fixes for the tests in your branch but this still has not yet been merged (mostly since I haven't had time for anything webbpsf this past month plus). I'm going to reopen this.

shanosborne:
@mperrin, shoot sorry that was a mistake. I was doing some local branch clean up and must have confused the branch names. My bad!

mperrin:
This PR is replaced by PR #505, which now has the same functionality, fixes the CI problems, has improved unit tests, and a cleaner git commit history

PR #391, opened 2020-11-16
Move missing data files message inside error

### LINKED ISSUE #380: Have a way to suppress the data files missing error message
This is probably related to #266, but I didn't see it explicitly so I'm raising it as a distinct issue:

When I do, say, the following:
```
from webbpsf import NIRSpec
instrument = NIRSpec()
```

I get an `OSError` about the data files being missing, which is quite sensible.  However, at the same time *even if I catch the exception*, I see a big ``***********  ERROR  ******  ERROR  ******  ERROR  ******  ERROR  ***********`` message.  This is problematic because it means downstream applications can't check for this error and do the download as needed.  It doesn't seem like this is at all clear because the error gets printed as `stderr.write(...)`, bypassing any possible catching mechanisms. 

Several solutions come to mind of varying complexity, but I think the easiest is simply to include this error message as part of the `OSError` instead of a separate error message.  The user will still then see the error (and it normally goes to stderr), but catching the exception will supress it.

shanosborne:
Closes #380 

Moves missing data files print out inside the `IOError`/`EnvironmentError` call so it can be caught by users if necessary

### DIFF
--- a/webbpsf/utils.py
+++ b/webbpsf/utils.py
@@ -201,14 +201,13 @@ def get_webbpsf_data_path(data_version_min=None, return_version=False):
     if path_from_config == 'from_environment_variable':
         path = os.getenv('WEBBPSF_PATH')
         if path is None:
-            raise EnvironmentError("{}\nEnvironment variable $WEBBPSF_PATH is not set!".format(
-                MISSING_WEBBPSF_DATA_MESSAGE))
+            raise EnvironmentError(f"Environment variable $WEBBPSF_PATH is not set!\n{MISSING_WEBBPSF_DATA_MESSAGE}")
     else:
         path = path_from_config
 
     # at minimum, the path must be a valid directory
     if not os.path.isdir(path):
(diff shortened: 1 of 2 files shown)


webbpsf/utils.py
```diff
@@ -201,15 +201,14 @@ def get_webbpsf_data_path(data_version_min=None, return_version=False):
     if path_from_config == 'from_environment_variable':
         path = os.getenv('WEBBPSF_PATH')
         if path is None:
-            sys.stderr.write(MISSING_WEBBPSF_DATA_MESSAGE)
-            raise EnvironmentError("Environment variable $WEBBPSF_PATH is not set!")
+            raise EnvironmentError("{}\nEnvironment variable $WEBBPSF_PATH is not set!".format(
```
mperrin: Minor:  maybe put the lengthy error message text at the end of the exception messages instead of the front? Also, now that we're only supporting recent Pythons we can do all the string formatting with f-strings, so like: 
```
raise  EnvironmentError(f"Environment variable $WEBBPSF_PATH is not set!\n{MISSING_WEBBPSF_DATA_MESSAGE}"
```

mperrin (approved):
An old PR that I still owed you a review for. Looks good to me but here is one optional suggestion.

PR #392, opened 2020-11-17
WIP: OTE Secondary Mirror IFM field-dependence implementation

Skyhawk172:
This PR is intended for the implementation of the OTE field-dependent Secondary Mirror Influence Function, based on BALL SER #2288152 as well as work by Randal Telfer.

On top of code additions/changes, this PR also relies on two input CSV files, located in the ./otelm folder. 

NOTE: SMIF relies on ote.v2v3 to determine field angle and need to determine how to best get that information from the instrument instance.  Currently implemented using the instrument instance's `_tel_coords()` function to assign the ote.v2v3 values (see webbpsf_core.calc_psf() ).


webbpsf/opds.py
```diff
@@ -1059,7 +1062,7 @@ class OTE_Linear_Model_WSS(OPD):
     """
 
     def __init__(self, name='Unnamed OPD', opd=None, opd_index=0, transmission=None, segment_mask_file='JWpupil_segments.fits',
-                 zero=False, rm_ptt=False, rm_piston=False, v2v3=None):
+                 zero=False, rm_ptt=False, rm_piston=False, v2v3=None, ctrl_pt_instr='nircam', ctrl_pt_fp='nrca3_full'):
```
mperrin: Doc string needs comments to describe the new parameters for the control points. 

Please spell out `control_point_` rather than `ctrl_pt`. (And is "control point" the right nomenclature, or should it be something like "reference point"? Not sure.  If control point is the terminology used in the WAS we should be consistent here, but we should also make sure to have comments that spell out clearly what this means and how it's used! )

webbpsf/opds.py
```diff
@@ -1347,15 +1352,98 @@ def _get_basis(*args, **kwargs):
...
+        dx/dy are the field separation from the control point, in radians;
+        Z_k are the Zernike (3: defocus, 4: 0-Degree Astig, 5: 45-Degree Astig);
+        alpha and beta are taken from the SMIF_Matrix/hexike.csv calculated 
+        by Randal Telfer.
+        
+        See Also Ball SER 2288152 "JWST WFSC MIMF Control Algorithm Design".
+
+        Parameters
+        ----------
+        dx, dy = field angles, in radians, from the OTE control point (nominally, NRCA3_FULL).
```
mperrin, line 1396: Needs parameter description for what the `zernike` input parameter does.

webbpsf/opds.py
```diff
@@ -1347,15 +1352,98 @@ def _get_basis(*args, **kwargs):
...
+        # GET SM POSE:
+        # RE-ORDER SUCH THAT (1) X-TRANS, (2) Y-TRANS, (3) X-TILT, (4) Y-TILT.
+        sm_errors = [self.segment_state[-1, 2],     # X-TRANS
+                         self.segment_state[-1, 3], # Y-TRANS
+                         self.segment_state[-1, 0], # X-TILT
+                         self.segment_state[-1, 1], # Y-TILT
+                         self.segment_state[-1, 4]] # PISTON; Ignore for consistency with BALL SER
+        
+        # GET SM INFLUENCE MATRIX AND CLEAN UP THE ARRAY A BIT:
+        zernike = zernike
```
mperrin: this line does nothing

webbpsf/opds.py
```diff
@@ -1347,15 +1352,98 @@ def _get_basis(*args, **kwargs):
...
+        zernike = zernike    
+        if zernike:
+            #ZERNIKE PROJECTED ONTO EXIT PUPIL:
+            smif = np.genfromtxt(os.path.join(__location__, 'otelm', "SMIF_zernike.csv"), delimiter=",", skip_header=1, usecols=[0, 1, 2, 3, 4, 5, 6])
+        else:
+            # HEXIKE PROJECTED ONTO ENTRANCE PUPIL (WHAT WEBBPSF NEEDS):
+            smif = np.genfromtxt(os.path.join(__location__, 'otelm', "SMIF_hexike.csv"), delimiter=",", skip_header=1, usecols=[0, 1, 2, 3, 4, 5, 6])
+
+        # DELETE THE FIRST ROW OF EACH SM MODE SINCE WE'RE NOT USING IT (RIGHT?)
+        smif = np.delete(smif, [0, 3, 6, 9, 12], axis=0)
```
mperrin: I find this part confusing to follow - what's in the first row? This may need some more comments to explain what the expected format and layout of the table is, etc. It's not obvious how the 0,3,6,9 etc maps into deleting the first row of each mode...

webbpsf/opds.py
```diff
@@ -1347,15 +1352,98 @@ def _get_basis(*args, **kwargs):
...
+
+        # REMOVE FIRST ELEMENT FROM EACH ROW (i.e. THE "MODE" COLUMN):
+        smif = smif[:,1:]
+    
+        alphas = smif[0::2] # EVEN ROWS ONLY
+        betas  = smif[1::2] # ODD ROWS ONLY
+
```
mperrin: This all seems a little more confusing than ideal, and relies on meta-knowledge about the table that is not in the table itself. There's nothing in the table or its header that explains that every other row is the 'alpha' or 'beta' coefficients... Ideally this should be added in as another table column (maybe converting this into a real astropy table rather than just numpy array, so we can have metadata?) or else add some commented header lines starting with `#` to the start of that file to explain what's going on. I worry this is all complicated enough that if we don't write down clearly how it works, we will really struggle if we have to edit this code again at some point in the future...

webbpsf/opds.py
```diff
@@ -1347,15 +1352,98 @@ def _get_basis(*args, **kwargs):
...
+        Updates self.opd based on V2V3 coordinates and amplitude of SM pose.
         """
-        # To be written! Probably will look a lot like _apply_global_zernikes, after
-        # computing the Zernike coefficients based on the multifield model
-        pass
 
+        if self.v2v3 is None:
+            return
+        else:
+            dx =-(self.v2v3[0] - self.ote_ctrl_pt[0]).to(u.rad).value # NEGATIVE SIGN B/C TELFER'S FIELD ANGLE COORD. SYSTEM IS X/Y = -V2/V3
```
mperrin: This comment confused me, in this part: 
```
TELFER'S FIELD ANGLE COORD. SYSTEM IS X/Y = -V2/V3
```

I first read those as ratios, like it was dividing the X and Y coordinates together. Could not understand why that was the field angle. I think this would be more clear if written as 
```
TELFER'S FIELD ANGLE COORD. SYSTEM IS (X,Y) = (-V2, V3)
```
which I think is what you actually mean here?

webbpsf/opds.py
```diff
@@ -1347,15 +1352,98 @@ def _get_basis(*args, **kwargs):
...
+
+            z_coeffs = self._get_zernike_coeffs_from_smif(dx, dy, **kwargs)
+
+            perturbation =  poppy.zernike.opd_from_zernikes(z_coeffs, npix=1024, 
+                                                    basis=poppy.zernike.hexike_basis_wss, aperture=self.amplitude)
+
+            perturbation[~np.isfinite(perturbation)] = 0.0
+            
+            self.opd += perturbation*1e-6
+            self.opd_header['SM_IFM'] = 'Applied'
```
mperrin: This header is OK but not super informative. We could instead  add in either a record of the Zernike coeffs used, or (at some small cost in increased computation time) the overall RMS of the perturbation.

webbpsf/tests/test_opds.py
```diff
@@ -273,3 +274,37 @@ def test_single_seg_psf(segmentid=1):
...
+    """ 
+    Test that the OTE SM Influence function returns expected Hexike coefficients.
+    """
+    
+    # Create an instance of the OTE linear model
+    otelm = webbpsf.opds.OTE_Linear_Model_WSS()
+
+    # Case 1: otelm.v2v3 is None, should return None
+    otelm._apply_sm_field_dependence_model()
+    assert ( otelm._apply_sm_field_dependence_model() is None)
```
mperrin, line 304: Not sure this test case is as rigorous as it could be; doesn't  function always return None? Even if there are v2v3 coordinates it still doesn't return anything, I think.

webbpsf/tests/test_opds.py
```diff
@@ -273,3 +274,37 @@ def test_single_seg_psf(segmentid=1):
...
+
+    # Case 2: check coefficient at control point; should return zeros.
+    assert( np.allclose(otelm._get_zernike_coeffs_from_smif(0., 0.), np.asarray([0.]*9) ))
+
+    # Case 3: dx=1, dy=1, SM Poses all equal to 1 um
+    telfer_zern = [-0.055279643, -0.037571947, -0.80840763, -0.035680581, -0.0036747300, 0.0033910640] # Taken from Telfer's tool, after some tweaks
```
mperrin: after some tweaks? Sort of ambiguous :-)

webbpsf/tests/test_opds.py
```diff
@@ -273,3 +274,37 @@ def test_single_seg_psf(segmentid=1):
...
+    otelm.segment_state[-1, :] = 1.0
+    assert (otelm._get_zernike_coeffs_from_smif(1.0, 1.0, zernike=True), telfer_zern) # Telfer's tool uses exit pupil Zernike a priori
```
mperrin: This is a little tricky, because the test case checks the Zernike version, but the actual usage in WebbPSF will use the hexike version. Does testing the Zernike version here validate the correct results in the hexike version? I don't think so, at least not directly... Can we come up with a test case that more directly validates the code as used to modify the OPDs for PSF calculations?

webbpsf/opds.py
Skyhawk172: Removed.

webbpsf/opds.py
Skyhawk172: The WAS indeed refers to it as the control point, so I've spelled it out and added to the doc string, as suggested.

webbpsf/opds.py
Skyhawk172: Yes, that's correct. I fixed it.

webbpsf/opds.py
Skyhawk172: Good point - I've added the coefficients of all the Hexikes that we're adding in.

webbpsf/opds.py
Skyhawk172: Okay, I decided to have the CSV files be more explicit about what each row is. The manipulation as an astropy Table is also cleaner, so hopefully it's easier to follow when looking at the CSV files themselves and the doc string notes.

webbpsf/opds.py
Skyhawk172: see comment below about reformatting the CSV files to be more explicit.

webbpsf/opds.py
```diff
@@ -1084,6 +1087,15 @@ def __init__(self, name='Unnamed OPD', opd=None, opd_index=0, transmission=None,
...
             the master chief ray location between the two NIRCam modules.
+        control_point_instr: str
+            Name of the instrument where the OTE control point is located. Default: 'nircam'.
+            The OTE control point is the field point to which the OTE has been aligned and defines the field angles
+            for the field-dependent SM pose aberrations.
+        control_point_fieldpoint: str
+            Name of the field point where the OTE control point is located, on instrument defined by "control_point_instr". 
+            Default: 'nrca3_full'.
+            The OTE control point is the field point to which the OTE has been aligned and defines the field angles
+            for the field-dependent SM pose aberrations.
```
mperrin, line 1094: Under what scenarios, if any, would we want to change these control point parameters? And can't the control point instrument be inferred from the control point field point name?

webbpsf/opds.py
```diff
@@ -1349,15 +1369,100 @@ def _get_basis(*args, **kwargs):
...
+            z_coeffs = self._get_zernike_coeffs_from_smif(dx, dy, **kwargs)
+
+            perturbation =  poppy.zernike.opd_from_zernikes(z_coeffs, npix=1024, 
+                                                    basis=poppy.zernike.hexike_basis_wss, aperture=self.amplitude)
+
+            perturbation[~np.isfinite(perturbation)] = 0.0
+
+            self.opd += -perturbation*1e-6
+            
+            self.opd_header['SM_IFM'] = 'Applied SMIF field-dependent aberrations'
```
mperrin: minor, this one is probably better suited to be a HISTORY keyword. We might also consider including here in the header some of the numerical values for dx, dy, control point, and so on. (low priority)

webbpsf/otelm/SMIF_hexike.csv
```diff
@@ -0,0 +1,16 @@
+Type,SM Mode,H3 (45-astig),H4 (Focus),H5 (0-astig),H6 (X coma),H7 (Y coma),H8 (3rd Sphr)
```
mperrin, line 6: Ideally we should add a header line in these files with a `# ` to say what they are.

webbpsf/opds.py
```diff
@@ -1349,15 +1369,100 @@ def _get_basis(*args, **kwargs):
...
+        if self.v2v3 is None:
+            return
+        else:
+            dx =-(self.v2v3[0] - self.ote_control_point[0]).to(u.rad).value # NEGATIVE SIGN B/C TELFER'S FIELD ANGLE COORD. SYSTEM IS (X,Y) = (-V2,V3)
+            dy = (self.v2v3[1] - self.ote_control_point[1]).to(u.rad).value
+
+            z_coeffs = self._get_zernike_coeffs_from_smif(dx, dy, **kwargs)
+
+            perturbation =  poppy.zernike.opd_from_zernikes(z_coeffs, npix=1024, 
+                                                    basis=poppy.zernike.hexike_basis_wss, aperture=self.amplitude)
```
mperrin, line 1447: The basis function used here appears to always be the `hexike_basis_wss`; therefore there's no way to use the Zernike basis here - which seems to be inconsistent with the existence of a Zernike basis. Am I missing something? 

Anyway I sort of think we should just remove the Zernike option entirely, since the hexike basis option is the one we're actually using. Better to simplify and have less code, if there is not a particular use case for having multiple code paths.

webbpsf/tests/test_opds.py
mperrin: As I noted before, I don't really like this test case: we only ever use the `zernike=True` option here in this test case, not in actual optical calculations. And this doesn't test the correctness of the `hexike` path we actually do use. We should rework this. Any recommendations on how? It seems to me we should be able to come up with an equivalent test case using the hexike code, just by converting one basis into the other via the OPD decomposition codes in poppy.

webbpsf/webbpsf_core.py
```diff
@@ -1017,7 +1017,15 @@ def calc_psf(self, outfile=None, source=None, nlambda=None, monochromatic=None,
...
         self.options['crop_psf'] = crop_psf
 
-        # Run poppy calc_psf
+        # UPDATE THE OPD V2V3 BASED ON DETECTOR POSITION, IN ORDER TO CALCULATE SM FIELD-DEPENDENT WFE.
+        # SEE opds._apply_sm_field_dependence_model()
+        #
+        # v2v3 attribute exists only if using the linear model, so check first:
+        if hasattr(self.pupil, 'v2v3'):
+            self.pupil.v2v3 = self._tel_coords().to(units.arcsec)
+            self.pupil.update_opd()
```
mperrin: We should make this more clever so as to not re-invoke the update_opd machinery for multiple calculations at the same field point. 

```
if hasattr(self.pupil, 'v2v3') and self.pupil.v2v3 != self._tel_coords().to(units.arcsec):
```

mperrin (changes requested):
Hi Charles, we should get this merged into develop sometime soon but I think there are still a few pieces that aren't clear to me. I don't want to over-polish this because clearly it has worked through the recent WFTPs after all. But see below comments and let me know what you think.  Hopefully the below is small stuff we can mostly deal with fairly quickly within the next week or so.

webbpsf/opds.py
Skyhawk172: I've wondered about whether NRCA3_FULL or NRCA3_FP1 should be the control point, but regardless, I favor having the flexibility to adjust it if we ever need it.

As for the "control_point_instrument", yes, we can infer it from the field point. I'll make the change.

webbpsf/opds.py
Skyhawk172: Done.

webbpsf/otelm/SMIF_hexike.csv
Skyhawk172: Done.

webbpsf/opds.py
Skyhawk172: Agreed.

webbpsf/webbpsf_core.py
Skyhawk172: Done

Skyhawk172: Done.

mperrin (approved):
Looks good and I'm going to go ahead and merge this now. Thanks @Skyhawk172

PR #395, opened 2021-01-06
update syntax to remove unnecessary underscores after poppy PR 386

mperrin:
Update to remove now-unnecessary underscores before the `[_]get_optical_system` function.

Needed to fix failing tests after https://github.com/spacetelescope/poppy/pull/386

### DIFF
--- a/webbpsf/tests/test_nircam.py
+++ b/webbpsf/tests/test_nircam.py
@@ -354,7 +354,7 @@ def test_ways_to_specify_weak_lenses():
         nrc.pupil_mask = pup
         if filt is not None: nrc.filter = filt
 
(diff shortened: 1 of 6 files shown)


shanosborne (commented):
It looks like most the the tests have passed. However, the py37-stable test (an allowed failure) did fail though in the same way that the other test envs in 391 are failing

mperrin:
> However, the py37-stable test (an allowed failure) did fail though in the same way that the other test envs in 391 are failing

That makes sense: the stable test cases uses the released stable `poppy`, so it's inconsistent with the fix here for dev poppy. That test case is an allowed failure precisely for cases like this.

mperrin:
@shanosborne Are you able & ready to approve this so we can merge? Given that the one allowed failure case in CI is indeed allowed.

shanosborne (approved):

PR #397, opened 2021-01-12
Updates for sign convention and poppy 1.0

mperrin:
Update webbpsf to check for poppy versions >= 1.0, and if so, forgo the sign convention flip needed in earlier versions. See https://github.com/spacetelescope/poppy/pull/384 for details and discussion

Other supporting changes: 
* Update to pupil segment geometry used in the OTE linear model, to the RevW geometry, for more precise consistency with the OTE OPD at segment boundaries. 
* Options to display OTE OPDs in either exit or entrance pupil coordinates 
* New segmented-Zernike-type basis `JWST_WAS_Full_Basis` which is a 6DOF basis per each segment. This extends the functionality in `JWST_WAS_PTT_Basis` for piston, tip, tilt terms. 
* Updates to coronagraphic unit tests, in particular for better runtime performance


mperrin:
FYI @grbrady

Skyhawk172 (approved):
I have tested this PR by tilting different segments and confirming that the displacements are as expected.

PR #398, opened 2021-01-20
Gridded PSF library generation should copy hexike P/T/T keywords, if present

mperrin:
Fixes an oversight in which the extra FITS keywords that record segment-level hexike piston, tip, tilt were not getting copied to gridded PSF library output files.  

The keywords `S##PISTN`, `S##XTILT`, and `S##YTILT` (where `##` is some segment number) are added during PSF calculations if and only if using the OTE linear model with the `remove_piston_tip_tilt` option set to `True`. This is in order to record the segment tip and tilt values for later use, before removing them in certain calculations such as when generating individual segment PSFs. For this to be used in MIRAGE, we've gotta propagate these values into the gridded PSF library FITS headers too. (which are derived from the base PSF headers, but not all keywords are copied). 

This PR checks if such keywords exist, and if so then it copies 'em all over to the gridded PSF library.

### DIFF
--- a/webbpsf/gridded_library.py
+++ b/webbpsf/gridded_library.py
@@ -343,6 +343,16 @@ class CreatePSFLibrary:
             meta["NORMALIZ"] = (psf[ext].header["NORMALIZ"], "PSF normalization method")
             meta["TEL_WFE"] = (psf[ext].header["TEL_WFE"], "[nm] Telescope pupil RMS wavefront error")
 
+            # Copy values for per-segment Zernike {piston, tip, tilt}, if present
+            # (these are only present or used in incoherent segment PSF generation with the
+            # remove_piston_tip_tilt value set to True):
+            if 'S01PISTN' in psf[ext].header:
+                # if we have one such keyword, assume we have them for all segments and all types
+                for i in range(1,19):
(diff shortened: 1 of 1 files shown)


mperrin:
This enables https://github.com/spacetelescope/mirage/pull/607

mperrin:
Without this fix, here's the telescope state in OTE-16, computed (incorrectly!) using the segment psf options in MIRAGE, the segment PSF arrangement is a mess (due to un-modeled radial translation to optical tilt cross terms): 

Screen Shot 2021-01-19 at 10 17 42 PM

With this fix and https://github.com/spacetelescope/mirage/pull/607, the segment pattern is as it should be:

Unknown-2

mperrin:
Note, due to time-urgency, a copy of this fix was directly dropped right into https://github.com/spacetelescope/webbpsf/pull/392 too

mperrin:
Code coverage test fails, but that is not surprising since the unit tests don't exercise the`remove_piston_tip_tilt` option. That's not used in regular WebbPSF scenarios, and we aren't trying to have 100% coverage for the more esoteric OTE commissioning usage cases by just our team. This is sufficient so I'm going to go ahead and merge.

PR #400, opened 2021-01-25
WebbPSF: Rename WFIRST -> Roman

robelgeda:
This PR aims to replace the term ``WFIRST`` with ``Roman`` in the WebbPSF codebase as well as docs.


webbpsf/roman.py
```diff
@@ -1,6 +1,6 @@
 """
 ==============================
-WFIRST Instruments (pre-alpha)
+Roman Instruments (pre-alpha)
```
mperrin: We should delete the comment "(pre-alpha)" here, which is no longer true (and hasn't been for a long while! I hadn't realized this was still stated like that...)

robelgeda:
@mperrin We could provide backwards compatibility by adding

``import .roman as wfirst``

In `__init__.py` if you think this is a good idea. It would allow people to use the code they already have without having to rename all their imports, and sometimes api calls.

webbpsf/roman.py
```diff
@@ -466,7 +466,7 @@ def __init__(self):
             Set to True or False to force using or not using the cold pupil mask,
             or to None for the automatic behavior.
         """
-        pixelscale = 110e-3  # arcsec/px, WFIRST-AFTA SDT report final version (p. 91)
+        pixelscale = 110e-3  # arcsec/px, Roman-AFTA SDT report final version (p. 91)
```
mperrin: super minor, but we can also cite https://roman.ipac.caltech.edu/sims/Param_db.html for this (it's the same value, but more current reference than the old WFIRST-AFTA SDT report.)

mperrin:
> import .roman as wfirst

@robelgeda yes, I do think that's a good idea, yes please. In the long run I don't think we should maintain this name alias forever, but it does ease things in the short run for downstream users. And it's super simple to do.

robelgeda:
Sounds good to me, I have made the update and it seems to work on my local machine.

docs/roman.rst
```diff
@@ -27,18 +27,18 @@ To work with the WFI model, import and instantiate it just like any of the JWST
...
 
 .. note:: 
 
-   *Tutorial notebook for WFIRST*
+   *Tutorial notebook for Roman*
 
-   This documentation is complemented by an `IPython Notebook tutorial for WFIRST PSFs <http://nbviewer.ipython.org/github/spacetelescope/webbpsf/blob/stable/notebooks/WebbPSF-WFIRST_Tutorial.ipynb>`_. Downloading and run that notebook to use a beta notebook GUI for the WFI model, and to explore code samples for common tasks interactively.
+   This documentation is complemented by an `IPython Notebook tutorial for Roman PSFs <http://nbviewer.ipython.org/github/spacetelescope/webbpsf/blob/stable/notebooks/WebbPSF-WFIRST_Tutorial.ipynb>`_. Downloading and run that notebook to use a beta notebook GUI for the WFI model, and to explore code samples for common tasks interactively.
```
mperrin: What about this Jupyter notebook? Seems to me this should also be updated, or at least checked?

mperrin (changes requested):
Looks good overall and I think this is 99% of the way there. The one requested change I would like is to update also the Jupyter notebook `webbpsf/notebooks/WebbPSF-WFIRST_Tutorial.ipynb` as well, please.

mperrin (approved):
Looks good, thanks Robel.

PR #404, opened 2021-02-05
Updating to meet INS-JWST Software Standards

shanosborne:
This PR updates the repo to adhere to INS-JWST community software standards outlined in https://github.com/spacetelescope/ins-jwst-community-software/

We were mostly complient, but I added the following updates:
- Expand the contributing guide
- Replace the environment.yaml file with a requirements.txt file because dependabot works with requirements files and I thought it best not to have duplicate files
- Update the dependencies in the requirements.txt file, except for 1 (numpy) in order to test depedabot (which I'm setting up next)


mperrin:
One question; I see the requirements are all specified as "exactly equal to", rather than "greater than or equal to".  I think we probably want those to be >=, yes?

shanosborne:
Aha yes, that's the main difference between requirements.txt files and environment.yaml files. Requirements files should (according to my research) specify exact versions (its just how they're supposed to work). And that's what makes them useful for automated dependency controllers, since they'll always contain the newest version of the dependencies that work with the software. However, the setup.py/cfg file still has "<=" so that when users actually install the package, they can still have a range of version options.

mperrin (approved):

PR #409, opened 2021-02-16
Model SUR quantization for actuator step size

mperrin:
Fine moves are not infinitely fine; there's a minimum step size achievable by the actuators on JWST. Let's include that as a (small but nonzero) limiting factor. 

To do: 

- [ ] Unit test
- [ ] Documentation (succinct)
- [ ] Cross check with an actual hexapod kinematics model? This may want different scale factors per degree of freedom... but that may be best left out of scope for this PR.


mperrin:
Not going to worry about this for now. Overtaken by events / not particularly necessary for current purposes.

PR #410, opened 2021-02-17
bug fix: avoid matplotlib deprecation of changing colormaps

mperrin:
Copy colormap before modifying it. Avoids causing a warning about modifying a global colormap, like the following: 
```
2021-02-17 09:03:18,762 - stpipe - WARNING - /home/mperrin/software/webbpsf/webbpsf/opds.py:340: 
MatplotlibDeprecationWarning: You are modifying the state of a globally registered colormap. 
In future versions, you will not be able to modify a registered colormap in-place. To remove this 
warning, you can make a copy of the colormap first. 
cmap = copy.copy(mpl.cm.get_cmap("RdBu_r"))
```

Same issue as https://github.com/spacetelescope/poppy/pull/379, just in a different set of places in the code.

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -37,6 +37,7 @@ import astropy.io.fits as fits
 import astropy.units as u
 import logging
 from collections import OrderedDict
+import copy
 
 import poppy
 import poppy.zernike as zernike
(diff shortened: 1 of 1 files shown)


shanosborne (approved):
Looks good to me!

PR #382, opened 2020-09-14
Proof of Concept: override_aberrations function for WFIRST (Roman)

robelgeda:
There was an urgent request to make overriding the WFIRST/Roman detector aberrations data possible. This was initially planned to be a part of the WFIRST/Roman prism and grsim implementation. The current plan is to make this functionality available via @robelgeda's `override_detector` branch until the WFIRST/Roman prism and grsim implementation is complete. This PR  gives an overview of what is being provided in that branch which is a temporary means to satisfy the feature request. This PR may be closed when the initial prism and grism PR is opened. 

This branch will be closed once the prism and grsim PR is opened.

PR #416, opened 2021-02-24
Roman Prism and Grism 

robelgeda:
This PR introduces the Prism and Grism modes for the Roman WFI based on cycle 8 design. It also introduces the ability to user override the aberrations files used to construct detectors. 

This PR means #382 can be closed (any time).

# Updates 

## WebbPSF 
- **New** Grism and Prism filters: 
    - `GRISM_FILTER = 'G150'`
    - `PRISM_FILTER = 'P120'`
- Changing filters to `G150` or  `P120` changes the mode of the WFI and the aberrations files (unless there is a user aberrations override)
- **New** `WFI._aberrations_files`: Dictionary with all default aberrations files paths. Dictionary key is the WFI mode.
- **New** `WFI._load_detector_aberrations(path)`: Class function to load aberrations files at `path`
- **New** `WFI. _get_filter_mode (wfi_filter)`: given a filter (string), `wfi_filter`, return the WFI mode. WFI modes are:
    -  Imaging
    -  Grism
    -  Prisim 
- **New** `WFI.mode`: Class property that returns the current mode of the WFI instance by passing the current filter to `WFI. _get_filter_mode`
- **New** `WFI.override_aberrations(aberrations_path)`: Overrides and locks the current aberrations with aberrations at `aberrations_path`. Lock means changing the filter/mode has no effect on the aberrations. 
- **New** `WFI.reset_override_aberrations()`: Releases `WFI.override_aberrations` lock and start using the default aberrations.
- **New** Tests for mode and filter switching. 
- **New** Field point nearest point approximation (extrapolation). 

## WebbPSF Data 
- **New** `wim_zernikes_cycle8_prism.csv` and `wim_zernikes_cycle8_grism.csv` aberrations files.
- **Updated** `webbpsf-data/WFI/filters.tsv` to add  'G150' and  'P120' (`nlambda` set to 20)
- **New** Filter throughput files for `G150` and  `P120` from Pandeia 

## Sample PSFs

Screen Shot 2021-03-01 at 2 37 33 PM
Screen Shot 2021-03-01 at 2 37 23 PM


webbpsf/roman.py
```diff
@@ -473,22 +475,40 @@ def __init__(self):
...
         super(WFI, self).__init__("WFI", pixelscale=pixelscale)
 
         self._pupil_controller.set_base_path(self._datapath)
 
         self.pupil_mask_list = self._pupil_controller.pupil_mask_list
 
+        # Define defualt aberration files for WFI modes
+        self._aberrations_files = {
+            'imaging': os.path.join(self._datapath, 'wim_zernikes_cycle8.csv'),
+            'prisim': os.path.join(self._datapath, 'wim_zernikes_cycle8_prism.csv'),
```
mperrin: Typo: 'prism' spelled wrong in the key here

webbpsf/roman.py
```diff
@@ -473,22 +475,40 @@ def __init__(self):
...
-        assert len(self._detectors.keys()) > 0
+        self._load_detector_aberrations(self._aberrations_files[self.mode])
         self.detector = 'SCA01'
 
         self.opd_list = [
             os.path.join(self._WebbPSF_basepath, 'upscaled_HST_OPD.fits'),
         ]
         self.pupilopd = self.opd_list[-1]
 
+    def _load_detector_aberrations(self, path):
```
mperrin, line 535: Please add a doc string. This is a simple function so it doesn't need to be long but it should exist, please :-)

webbpsf/roman.py
```diff
@@ -558,6 +570,89 @@ def _unmasked_pupil_path(self):
...
+        """Current WFI mode"""
+        return self._get_filter_mode(self.filter)
+
+    @mode.setter
+    def mode(self, value):
+        """Mode is set by changing filters"""
+        raise AttributeError("WFI mode cannot be directly specified; WFI mode is set by changing filters.")
+
+    def override_aberrations(self, aberrations_path):
+        """Override and lock detector aberrations"""
```
mperrin: The "override aberrations" functionality is a little complicated and non-obvious, so I think this is a case where the doc string should be longer and more explanatory. Under what circumstances would a user want to override the aberrations? What contents need to be in the file at the provided path? Maybe there should be some example code showing how to use this, either here or in the Roman tutorial notebook?

webbpsf/roman.py
```diff
@@ -558,6 +570,89 @@ def _unmasked_pupil_path(self):
...
+
+        if value not in self.filter_list:
+            raise ValueError("Instrument %s doesn't have a filter called %s." % (self.name, value))
+
+        self._filter = value
+
+        # Update Aberrations
+        # ------------------
+        # Check if _aberrations_files has been initiated (not empty) and if aberrations are locked by user
+        if self._aberrations_files and not self._is_custom_aberrations:
```
mperrin, line 718: Why the check for if the aberrations files are empty? Under what circumstances would that be the case?

webbpsf/tests/test_roman.py
```diff
@@ -190,8 +193,51 @@ def _test_filter_pupil(filter_name, expected_pupil):
...
+def test_swapping_modes(wfi=None):
+
+    if wfi is None:
+        wfi = roman.WFI()
+
+    tests = [
+        # [filter, mode, pupil_file]
+        ['F062', 'imaging',  wfi._unmasked_pupil_path],
+        ['F184', 'imaging', wfi._masked_pupil_path],
+        [PRISM_FILTER, 'prisim', wfi._unmasked_pupil_path],
```
mperrin: This has the same typo for "prism" spelled wrong!

mperrin (changes requested):
Looks good overall. See below for a few small requests for revisions but nothing big. 

Note, I have not reviewed any of the data values within the new files, either for the aberration coefficients or the filter transmissions. I assume you have double checked those values are are they should be.

webbpsf/tests/test_roman.py
robelgeda: Thanks for catching this!

webbpsf/roman.py
robelgeda: When the class is initialized, with `super(WFI, self).__init__("WFI", pixelscale=pixelscale)`, the filter is set to a new value in that function. We can not fill in ` self._aberrations_files` before the `super.__init__` call because the `self._datapath` attribute is not yet initialized. So we have to initialize ` self._aberrations_files` to be empty and check for when `super.__init__` sets the filter, so the code does not crash.

If we define `self._datapath` before `super.__init__`, this will not be needed. But I thought this was the safest way forward.

webbpsf/roman.py
robelgeda: Sounds good

webbpsf/roman.py
robelgeda: I added a bit of comments, please let me know how they look

robelgeda:
From @Skyhawk172 :
> looking at the prism and grism spreadsheets, I noticed the “local_x” and “local_y” values are different than what we have in the “normal filter” spreadsheet.
> For example, for our regular filters, these values range from -20.44 to +20.44, identically for all SCAs, whereas in the prism/grism spreadsheets they vary non-uniformally from SCA to SCA and go from about -25 to 17.  The net effect of these differences is that the derived pixel value for some of these “local” coordinate values is negative and/or can be many tens of pixels from the edge of the detector, leading to an error “Could not get aberrations for input field point” when trying to get the aberration close to the edge of the detector.
> This local-to-pixel conversion is done in _load_wfi_detector_aberrations.build_detector_from_table().
> We make the assumption “to place (-20.44, -20.44) at (4, 4)“, but that means the prism/grism interpolator won’t cover all of the SCA pixels.

webbpsf/roman.py
```diff
@@ -11,15 +11,21 @@
 import os.path
 import poppy
 import numpy as np
-from . import webbpsf_core
-from scipy.interpolate import griddata
+
+from scipy.interpolate import griddata, RegularGridInterpolator
```
Skyhawk172, line 15: RegularGridInterpolator is not needed anymore now that we've simplified the handling of the extrapolation case.

webbpsf/roman.py
```diff
@@ -11,15 +11,21 @@
...
 from astropy.io import fits
 import astropy.units as u
 import logging
 
+from . import webbpsf_core
+from .optics import _fix_zgrid_NaNs
```
Skyhawk172, line 21: _fix_zgrid_NaNs is not needed anymore now that we've simplified the handling of the extrapolation case.

Skyhawk172 (approved):
I've done a series of tests and all my results indicate that the implementation is correct. We will be examining the results closer to understand some interesting features  (e.g. FWHM), but it does appear those features are the result of the reference data provided to us by GSFC.

robelgeda:
Added data to v1.0.0rc2 and tests are passing

mperrin (approved):

PR #417, opened 2021-02-24
Update MIRI detector cross artifact model; fixes #415

### LINKED ISSUE #415: WebbPSF model for MIRI "cross artifact" appears low compared to CDP
The MIRI cross artifact (detector internal scattering effect) model in WebbPSF appears to be substantially lower than what's shown in the CDP PSFs. See plots below.  Found this in the course of modifying MIRISim to work with WebbPSF for OTE commissioning sims. 

This is mostly relevant for the shorter wavelength PSFs (so, less relevant for coronagraphic science with MIRI but very relevant to OTE commissioning uses of MIRI which are mostly in F560W where this effect is strongest). 

The [code in webbpsf](https://github.com/spacetelescope/webbpsf/blob/develop/webbpsf/distortion.py#L306) is based on Alistair's tech report MIRI-TN-00076-ATC_Imager_PSF_Issue_4.pdf, which gives an IDL model for this effect, which we ported to Python. I can't remember now what cross checking we did at the time. (@shanosborne, any chance you have any recollection or notes on this?).  This doesn't use the [much higher fidelity GPU model](https://github.com/merope82/GimMIRI) by @merope82. 

Seems surprising to me that no one on the MIRI team has noticed or commented on this before... surely someone has done this comparison before me?  @JarronL, @merope82, @swolff9 - have any of you perhaps looked at this, or know of others who might have? 

Is Alistair's tech report still the best reference for this? Is there some other documentation of what was assumed in the CDP PSFs? Updating the model parameters in webbpsf would be easy, if we can get guidance as to what the appropriate values are. Or shall I just empirically adjust the amplitude until it matches the CDP PSFs? 

Unknown
Unknown-1

mperrin:
See discussion at #415: the model for the MIRI cross artifact (detector internal diffraction) turned out to be low in webbpsf compared to the MIRI CDP PSFs. I am not sure how this was not caught before. I derived empirical scaling factors that bring them into agreement. 

Relatively simple PR in terms of the code; all the effort was in figuring out the appropriate scale factors and verifying things are now working as intended. 

This PR assumes that the MIRI CDP PSFs can be taken as a reliable reference for the cross artifact amplitudes vs wavelength. 

**Before (using original amplitudes):**

See difference column at right, then compare to the "after" version of this plot, below.. 

Unknown-4

**After (using rescaling factors):**

Same plot scale as above. 

Unknown-7

Horizontal cuts across the PSFs show excellent quantitative agreement now (unlike the prior plots visible in #415): 

Unknown-9


mperrin:
@merope82 would you be so kind as to take a look at the plots on this update to WebbPSF, please? It's pretty quick and dirty and ad hoc, but I think suffices for current purposes.

merope82:
Hi @mperrin! They do look significantly better; possibly even better at
approximating the artifact than the model. Obviously scaling is scaling. As
you can see, the diffraction also widens the PSF features (basically yields
a halo around all higher flux regions), which the scaling does not subtract.

shanosborne (approved):
Looks good to me! I'm glad to see we got some more accurate numbers

mperrin:
FYI to @merope82  and @mlibralato

PR #418, opened 2021-02-24
Add notebook to help test sign convention

Skyhawk172:
In preparation for v1.0 release, here's a notebook to help debug/check that we get our signs correct.

PR #419, opened 2021-02-26
add unit test for sign convention; adapted from notebook version in PR #418

mperrin:
Supersedes #418 by @Skyhawk172. Same test, now in unit test form rather than notebook so it will run as part of the CI infrastructure.


mperrin:
This is @Skyhawk172's code (from #418), which I adapted into unit test format rather than notebook format. I think that can count more or less as a "review" for this, and I'd be fine with merging now. But first I'll check if any of @Skyhawk172 or @shanosborne or @obi-wan76 want to look this over today and provide a quick review via the GitHub PR interface, just for the record?

PR #424, opened 2021-03-01
Adding synphot

shanosborne:
This PR aims to work with https://github.com/spacetelescope/poppy/pull/388 to replace `pysynphot` with `stsynphot`.

After discussion with @mperrin, we decided that completely replacing `pysynphot` in webbpsf would cause problems for our dev users (especially ote sim mirage users) if they have mismatched webbpsf and poppy versions. So instead, we decided to add a check for the user's poppy version inside webbpsf. If poppy > 0.9.2 (the latest release), the code would use `stsynphot`, if poppy <= 0.9.2, the code would use `pysynphot`. We will remove this check for the next release of webbpsf, but this will help users in the meantime.

Testing Note: Because of the version check above, since the poppy PR with the `stsynphot` changes isn't merged yet, in order to test this you'll need to change the `>` to `<` in line 57 in `webbpsf_core.py`. That or we'll need to rebase the poppy PR to appear that the change is after 0.9.2


shanosborne:
@pllim I've tried running this branch with your branch in poppy and I'm getting some errors from the `stsynphot` calls in `poppy`. Would you be able to look this over and see if you understand the errors? Thanks!

dev_utils/compute_psf_library.py
```diff
@@ -6,13 +6,19 @@
...
+from packaging import version as package_version
+import poppy
+if not os.environ.get('PYSYN_CDBS'):
+    os.environ['PYSYN_CDBS'] = '/grp/hst/cdbs'
+assert exists(os.environ['PYSYN_CDBS']), "Can't load synthetic photometry files!"
+poppy_ver = poppy.__version__
+if package_version.parse(poppy_ver) > package_version.parse("0.9.2"):
+    import stsynphot
+else:
+    import pysynphot
```
pllim: I wouldn't do this: This looks like a maintenance nightmare. Much easier to cut the next major release and not use pysynphot at all.

pllim (commented):
pysynphot and stsynphot are not 100% interchangable, especially when it comes to Quantity support and underlying machinery.

dev_utils/compute_psf_library.py
mperrin: Agreed in the long run, yes. This is intentional and needed in the short term transitional period however. We plan to take it out after this next release.

pllim:
The tests passed here. I don't understand where are the errors you need help with. Can you please clarify? Thanks!

shanosborne:
@pllim, the errors come from running this branch of webbpsf with your `pysyn-stsyn-port` branch of poppy, in an environment where `stsynphot` is installed

pllim:
@shanosborne , can you please have this branch call that branch here, so we can all see the log in CI? Thanks!

pllim:
I waited a while for https://travis-ci.org/github/spacetelescope/webbpsf/jobs/760994800 to run but it passed. 🤷  (I cancelled the irrelevant jobs.)

shanosborne:
Oh shoot you're right. I messed up the test and didn't add an `stsynphot` installation to it, so it passed because it never called any of the `stsynphot` code (which is optional and is only called if you have the package installed). I've just updated the test and hopefully it actually tests what we want it to now. Sorry about that!

pllim:
Seems like Travis CI has a very long queue time now. Still waiting for it to start. 🤞

shanosborne:
Right, so now we have the case where the fact that the poppy PR isn't merged yet has become a problem. So I'm going to add a commit to flip the sign of the version check in `webbpsf_core.py` which we will need to undo before merging the PR, but is necessary for testing.

mperrin:
@shanosborne for the version check - how about instead of looking at the version number we try looking at just what poppy is using directly. Something like this: 

```
try:
   import poppy.instrument.stsynphot
   # poppy is using sysynphot so webbpsf should too
   import stsynphot
except ImportError:
   import pysynphot
```

And, optional, but I wonder if it's worth consolidating these if/else statements into one place in the code, so there's less to undo later when we can take out all the pysynphot refs eventually. 

This all would indeed be easier if we could afford a case of temporary breakage of dependencies if we don't merge both PRs at the exact same time, but alas from past experience that way lies confusion and hassle...

pllim:
I still think it is less maintenance burden to branch out 0.9.x that only uses pysynphot and only support stsynphot going forward. People who want to use pysynphot can use your 0.9.x releases and you can backport things into it from develop as needed. But I am not the one who has to maintain this package.

pllim:
I guess we finally got it? Are the errors in https://travis-ci.org/github/spacetelescope/webbpsf/jobs/761016058 what you also see locally, @shanosborne ?

pllim:
Most of the failures are caused by your source and bandpass "waveset" not overlapping completely or at all. This usually means either one of them have wavelength in existing values but now in wrong unit or some assumptions you made with PySynphot no longer makes sense. I'll see if I can sniff it out with your test cases.

There is one error where somehow you are passing in zero or negative for wavelength, which is not allowed.

pllim:
Looking at [check_overlap()](https://synphot.readthedocs.io/en/latest/api/synphot.spectrum.SpectralElement.html#synphot.spectrum.SpectralElement.check_overlap) and the logic at https://github.com/spacetelescope/synphot_refactor/blob/3254fd40cb07017288febed3144aa1bdb486f0e7/synphot/spectrum.py#L1522-L1523

```python
            if self.waveset is None:
                return 'partial_notmost'
```

my hunch is that your bandpass is analytical, which would have undefined `waveset` (because you can sample it to infinity). In PySynphot, there was no real analytic model.

An example:

```python
>>> from synphot import SpectralElement, SourceSpectrum, Observation
>>> from astropy.modeling.models import Const1D
>>> source = SourceSpectrum.from_vega()  # Some empirical source
>>> bandpass = SpectralElement(Const1D, amplitude=0.8)  # Purely analytical bandpass
>>> obs = Observation(source, bandpass)
PartialOverlap: Source spectrum and bandpass do not fully overlap. You may use force=[extrap|taper] to force this Observation anyway.
```

Now you have to think about how you wish to sample the observation. Does extrapolation makes sense or you want to taper at the edges of `source.waveset`?

From my experience, `taper` is the safer way to go, as blind extrapolation might cause overestimation of observed flux, but it really depends on your use case.

pllim:
After discussions with @shanosborne , I have created [this gist](https://gist.github.com/pllim/baa5e38525a5df38c282d40f25447763) to explain why you are seeing failures, and provide workarounds, where possible. Many thanks to Shannon who patiently provided me with the necessary info.

**tl;dr**

1. No overlap and partial overlap -- Your bandpass is too far from the blackbody peak, resulting in useless observation. You can force the observation to happen in `stsynphot` by resampling the blackbody (as shown in the gist above) but don't expect the observation to give you anything useful.
2. ~Negative wavelength -- Your bandpass is unphysical. Do not create such a bandpass.~ After the bandpass typo is fixed, this is also a partial overlap case, as above.

mperrin:
@shanosborne FYI the PR in poppy is now merged, so here we can take out the temporary parts (extra tox test case etc) and hopefully merge this shortly. Can you take care of the few updates needed in this, please? I'm not sure if anything's needed besides just the changes in the test setup?

docs/references.rst
```diff
@@ -133,9 +133,9 @@ Instrument + Filter Throughputs
 ---------------------------------
 
 Where possible, instrumental relative spectral responses were derived from the
-Pysynphot CDBS files used for the development version of the JWST Exposure Time Calculators (ETCs),
+Stsynphot CDBS files used for the development version of the JWST Exposure Time Calculators (ETCs),
 normalized to peak transmission = 1.0 (because absolute throughput is not
-relevant for PSF calculations). Not all filters are yet supported in Pysynphot,
+relevant for PSF calculations). Not all filters are yet supported in Stsynphot,
 however.
```
mperrin: Minor: This text can be updated, as it's outdated in ways besides just the py/stsynphot change. Let's say the following:

Instrumental relative spectral responses are derived from the `reference data files used by the Pandeia engine  <https://jwst-docs.stsci.edu/jwst-exposure-time-calculator-overview/jwst-etc-pandeia-engine-tutorial/installing-pandeia#InstallingPandeia-DataFiles>`_  for the `JWST Exposure Time Calculator<https://jwst.etc.stsci.edu>`_, normalized to peak transmission=1.0 (because absolute throughput is not relevant for PSF calculations).

mperrin (changes requested):
As discussed this should have the temporary test case revised out before we merge. 

Looking at this now I'm starting to have second thoughts about the back compatibility approach, but probably not enough to revise. I still want to make sure we take zero chance of breaking any dependencies in any of the team practices etc; but I'm not sure we want all the switching code around in the long run. We probably should consider pysynphot deprecated and remove this before some later release I suppose. 

One minor change requested, plus the test case stuff, and then this should be good to merge. Thanks!

pllim:
Re: tox -- You still want a new job to test against stsynphot (unless you want to swap it out with pysynphot in existing jobs) but maybe not call it "poppydev".

mperrin:
Travis passes, codecov is "failing" but irrelevant (-0.01% delta), I'm going ahead with this since it's needed with the current dev poppy.

PR #429, opened 2021-03-23
Replace griddata with RegularGridInterpolator for distortions

### LINKED ISSUE #426: WebbPSF's distortion model is really slow to compute - we should improve it
It's always been the case since we added the distortion model that it's a slow part of the calculation, but I've only just recently realized _how_ slow. For typical use cases it's 75-95% of the computation time, totally dominating over the actual optical propagation. 

I find myself often disabling the distortion model (add_distortion=False) in most of my own calculations because it's so painfully slow... I'd like to improve on this. 

Timing results first, then ideas on what to improve. 

<HR>

**_Timing results_:**

Some detailed timing results using `line_profiler` in Jupyter: 
```
%load_ext line_profiler

nrc = webbpsf.NIRCam()

%lprun -f webbpsf.webbpsf_core.SpaceTelescopeInstrument.calc_psf psf = nrc.calc_psf(nlambda=1)
```
That will output a detailed printout of the runtime for calc_psf, including the fraction of time in each line of that function. The relevant one to look for is `self._calc_psf_format_output(result, local_options)`; the distortion model is called within that to compute the images for the additional output extensions. 

In the above case of a monochromatic calculation, that line takes **95% of the time**,  6.7 seconds out of total runtime 7 seconds. The actual optical propagation is just 0.3 s in comparison! 

Compare to without the distortion model:
```
%lprun -f webbpsf.webbpsf_core.SpaceTelescopeInstrument.calc_psf psf = nrc.calc_psf(nlambda=1, add_distortion=False)
```
It takes 32% (.17 s) out of a total runtime of 0.5 s to format the output (in this case it's just binning down to the detector pixel scale, without preparing the distortion model).

Similar for a broadband calculation with 20 wavelengths:
```
%lprun -f webbpsf.webbpsf_core.SpaceTelescopeInstrument.calc_psf psf = nrc.calc_psf()
```
The distortion model again takes 6.8 s out of a total runtime of 9.3 s. 

Using the call profiler `%prun` confirms that it's the ndgriddata function call that's taking up nearly all that time. (See e.g. output of `%prun psf = nrc.calc_psf(nlambda=1)`

<HR>

**_What might we do to improve this_:**

The current distortion model is almost certainly overkill: we use the SIAF distortion polynomials to work out the coordinate transforms on a pixel-by-pixel basis from the undistorted 'ideal' frame to the detector-coordinates 'science' frame, then use scipy.interpolate.griddata to warp the PSF into the distorted frame.

Doing this on a pixel-by-pixel basis is extremely generalized but likely not necessary for the modest size of PSF postage stamps. I expect we could get very, very close to the same results just using a simple linear transformation (e.g rotation + skew matrix) which can be implemented in a much more performant way using for instance the scipy affine_transform function.  

Given the relatively small geometric distortion in the JWST instruments, I expect this linear approximation model will be adequate for our purposes, and would remove what is currently the most substantial bottleneck in JWST PSF calculations. I don't think it makes sense to continue spending 70+% of calculation runtime on what is a relatively minor effect for most use cases of PSF modeling. (though still important enough we do want to include it at some level)

@shanosborne @grbrady @obi-wan76 @Skyhawk172 I'm curious to hear if you have any thoughts on the above?

JarronL:
Previously, the `add_distortion` function used scipy's `griddata` to transform from an irregularly gridded 'sci' coordinate array to a regularly gridded 'sci' coordinate array. However, the input 'sci' coordinates are derived from a regularly gridded 'idl' array. Since, it would be much faster to use RegularGridInterpolator (RGI), we could instead start with these 'idl' coordinates and interpolate onto some set of irregular 'idl' points, which map to our desired 'sci' coord values.

I've written a generalized distortion function that uses this concept to transform an image between 'idl' coordinates to some other `pysiaf` frame ('sci', 'tel', 'det'). This is useful for other purposes, like distorting model images (e.g, disks) into the detector frame before convolving with PSFs.

Benchmarks on my laptop for `fov_pixels=512` and `oversample=4` for a single PSF:
1. without distortion: ~11 sec
2. with original distortion (griddata): ~75 sec
3. with updated distortion (RGI): ~13 sec

While the original implementation added a factor of ~6x to the PSF calculation time, the RGI only takes a 10-20% overhead hit. The resulting distortions calculations are also very similar:

Above is the difference of the corner PSFs relative to the PSF at the center of the detector (for A5, F405N). I've set `include_si_wfe=False` so that the only change would be from geometric distortions rather than WFE differences.

So, I think solves the excessive distortion overhead in #426.

### DIFF
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -2,13 +2,10 @@ import copy
 
 import astropy.convolution
 import astropy.io.fits as fits
-from astropy.io.fits.util import fill
 import numpy as np
-from poppy.optics import NgonAperture
 import pysiaf
 from scipy.interpolate import RegularGridInterpolator
 from scipy.ndimage.interpolation import rotate
-
 
 def _get_default_siaf(instrument, aper_name):
     """
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -91,6 +91,11 @@ def distort_image(hdulist_or_filename, ext=0, to_frame='sci', fill_value=0,
         will return the full set of new coordinates. Output will then
         be (psf_new, xnew, ynew), where all three array have the same
         shape.
+    aper : None or :mod:`pysiaf.Aperture`
+        Option to pass the SIAF aperture if it is already known or
+        specified to save time on generating a new one. If set to None,
+        then automatically determines a new `pysiaf` aperture based on
+        information stored in the header.
     """
 
     # Read in input PSF
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -2,9 +2,11 @@ import copy
 
 import astropy.convolution
 import astropy.io.fits as fits
+from astropy.io.fits.util import fill
 import numpy as np
+from poppy.optics import NgonAperture
 import pysiaf
-from scipy.interpolate import griddata
+from scipy.interpolate import RegularGridInterpolator
 from scipy.ndimage.interpolation import rotate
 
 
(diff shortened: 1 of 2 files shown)


webbpsf/distortion.py
```diff
@@ -2,9 +2,11 @@
 
 import astropy.convolution
 import astropy.io.fits as fits
+from astropy.io.fits.util import fill
 import numpy as np
+from poppy.optics import NgonAperture
```
mperrin: Is this used anywhere?

webbpsf/distortion.py
```diff
@@ -2,9 +2,11 @@
 
 import astropy.convolution
 import astropy.io.fits as fits
+from astropy.io.fits.util import fill
```
mperrin: Is this used anywhere?

webbpsf/distortion.py
```diff
@@ -40,8 +42,150 @@ def _get_default_siaf(instrument, aper_name):
...
+        is same size as input image, and coordinate grid spans the min
+        and max values of siaf_ap.convert(xidl,yidl,'idl',to_frame). 
+    ynew_coords : None or ndarray
+        Array of y-values in new coordinate frame to interpolate onto.
+        Can be a 1-dimensional array of unique values, in which case 
+        the final image will be of size (ny_new, nx_new). Or a 2d array 
+        that corresponds to full regular grid and has same shape as 
+        `xnew_coords` (ny_new, nx_new). If set to None, then final image
+        is same size as input image, and coordinate grid spans the min
+        and max values of siaf_ap.convert(xidl,yidl,'idl',to_frame).
```
mperrin, line 88: What are the use cases for passing in coordinate arrays here? I can see this adds flexibility, but also complexity. I'm curious if you have particular needs or scenarios in mind that will use this.

webbpsf/distortion.py
```diff
@@ -40,8 +42,150 @@ def _get_default_siaf(instrument, aper_name):
...
+        that corresponds to full regular grid and has same shape as 
+        `xnew_coords` (ny_new, nx_new). If set to None, then final image
+        is same size as input image, and coordinate grid spans the min
+        and max values of siaf_ap.convert(xidl,yidl,'idl',to_frame). 
+    return_coords : bool
+        In addition to returning the final image, setting this to True
+        will return the full set of new coordinates. Output will then
+        be (psf_new, xnew, ynew), where all three array have the same
+        shape.
+    """
```
mperrin, line 99: doc string is missing aper parameter

mperrin (changes requested):
This looks great! I have a  few very minor change requests; it looks like there's a couple extra unused imports, and one parameter missing in the doc string. 

Thanks very much for doing this.

webbpsf/distortion.py
JarronL: Weird. No idea how this and the other import calls got in there, as I don't see them in the develop branch. Maybe I accidentally cloned an older version of the branch?

webbpsf/distortion.py
JarronL: Odd occurrence. Will remove.

webbpsf/distortion.py
JarronL: Useful for plotting purposes onto custom grid spacing. I am currently using this for visualizing model images onto already existing V2/V3 coordinate grids. If you want, I can remove these options here, since I've actually already made  a slightly different version of this function for `webbpsf_ext`.

webbpsf/distortion.py
JarronL: Added.

webbpsf/distortion.py
mperrin: No, definitely don't remove. I could tell you had put in the option for some reason, and wanted to understand what it was.

mperrin (approved):
Looks good! 

I checked the timing and results of a 3x3 PSF grid calculation using NIRISS. 
- With current develop, took 46.6 s.  
- With this PR, takes 25.5 s. 
The resulting PSFs are visually indistinguishable in the view from display_psf_grid.

The test calculation was a 3x3 grid with default properties per each PSF (using 21 wavelengths and a 101x101 PSF 4x oversampled). Hence the distortion part is only part of the total runtime. In any case, this suffices to verify a nice speedup from this PR. Thanks again @JarronL

PR #430, opened 2021-03-23
JWST Detector Position handling 

JarronL:
While updating the distortion grid, I realized that WebbPSF always converts the specified detector positions to ints and restricts boundaries within the detector pixels box. However, since these values correspond to SIAF aperture 'sci' coords and are primarily used to convert to 'tel' (V2/V3), converting to ints prevents precise transformations between the 'idl' and 'tel' coordinate systems. Indeed, the automatic SIAF aperture look-up many time specifies fractional numbers, that then get converted to ints. This makes it confusing when applying the header information to SIAF transformations, especially if one wants to track the small sub-pixel shifts for coronagraphic masks.

In addition, there's very little reason to constrain to the box of detector pixels, because it is possible to simulate PSFs outside this region (ie., SI WFE extrapolation). Instead of raising a ValueError, I just print a warning to alert the user in case they are unaware.

JWST instrument tests have been updated to account for fractional PSFs.

### DIFF
--- a/webbpsf/tests/test_miri.py
+++ b/webbpsf/tests/test_miri.py
@@ -68,6 +68,6 @@ def test_miri_aperturename():
     ref_tel_coords = miri._tel_coords()
 
     miri.aperturename = 'MIRIM_SUB256'
-    assert miri.detector_position == (128, 128), "Changing to a subarray aperture didn't change the " \
+    assert miri.detector_position == (128.5, 128.5), "Changing to a subarray aperture didn't change the " \
                                                  "reference pixel coords as expected"
     assert np.any( miri._tel_coords() != ref_tel_coords), "Changing to a subarray aperture didn't change the V2V3 coords as expected."
--- a/webbpsf/tests/test_niriss.py
+++ b/webbpsf/tests/test_niriss.py
@@ -65,7 +65,7 @@ def test_niriss_aperturename():
     ref_tel_coords = niriss._tel_coords()
 
     niriss.aperturename = 'NIS_SUB128'
-    assert niriss.detector_position == (64, 64), "Changing to a subarray aperture didn't change the " \
+    assert niriss.detector_position == (64.5, 64.5), "Changing to a subarray aperture didn't change the " \
                                                  "reference pixel coords as expected"
     assert np.any(
         niriss._tel_coords() != ref_tel_coords), "Changing to a subarray aperture didn't change the V2V3 coords " \
--- a/webbpsf/tests/test_psfgrid.py
+++ b/webbpsf/tests/test_psfgrid.py
@@ -32,8 +32,7 @@ def test_compare_to_calc_psf_oversampled():
     # Pull one of the PSFs out of the grid
     psfnum = 1
     loc = grid.meta["grid_xypos"][psfnum]
-    locy = int(float(loc[1]) - 0.5)
-    locx = int(float(loc[0]) - 0.5)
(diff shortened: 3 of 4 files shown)


JarronL:
Okay, thinking about this more, it seems like this has a fundamental conflict with odd/even PSFs (centered in the middle of a pixel versus at the pixel corners), which means it doesn't make sense to track shifts finer than 0.5 pixels. I will need to think more about how V2/V3 information is tracked, specifically for coronagraphic PSFs where I'm shifting the masks to emulate a PSF shift.

mperrin:
Part of the issue here is I think we've collectively never really nailed down the exact semantics of what we want the detector position variable to mean. I think we should try to nail down exactly what the intent is, if we're going to change any of this. 

Part of my original thinking - and a big part of why this rounds to integer pixels - was that the detector_pixel value is most useful for mapping the subarray produced by webbpsf into the full array detector coordinates. Now this was never really explicit. But in some sense this is the distinction between _detector_ position (where is this piece of the detector we are simulating?) and _source_ position (where is the star itself located with respect to that detector?). The two are implicitly related, but not identical!

The detector position (coordinates of the subarray being simulated) naturally has a straightforward interpretation, is always an integer offset, and is independent of factors such as distortion.  It avoids half pixel offsets for odd/even arrays. If you generate a set of coronagraphic simulations and offset the target around, the detector position stays constant; it's independent of the star position, and fixed relative to a SIAF aperture for instance. So that all seems nice, and relatively simple. 

Now on the other hand this becomes muddled in practice, because we use the detector position to look up the SI wavefront error model. That's a shortcut and an approximation, which holds under the assumption that most PSF calculations have no source offset away from the detector position, and half pixel or less offsets are small compared to the spatial scale on which wavefront varies significantly. 

As you say, the imprecision with respect to the SIAF field points is also a bit of an inconsistency. But would we really want PSF generation for some specified SIAF aperture point with a non-integer reference position return a PSF that is not centered within the output array by default, but instead is offset by whatever the fractional pixel amount is? I think that would not be the least surprising behavior. 

This all gets even more tricky when shifting the coronagraph masks instead, yes. 

I can think of a couple ways to proceed, if we're going to change any of this. 
 - Simply define some additional header keywords for additional coordinate reference information, computed however needed based on the detector position and any supplied additional offsets of either the source or the coronagraph mask. This could include writing separate keywords for the source apparent position distinct from the detector position. 
 - Make explicit the interpretation described above, that the detector_position keywords are intended to represent a detector subarray coordinate mapping. Set keywords for the subarray position within the full array (ideally, consistent with those keywords used in JWST pipeline data products for subarray data)
 - Even more ambitiously, create an add a WCS header giving the coordinate mapping into the full frame. (Is it possible to specify a WCS header not into Ra,Dec units but just into another pixel coordinate system? I think so but would have to consult the WCS spec for the exact details of how to set it up. Not sure this would be useful enough to be worth the effort; WCS information seems most useful in the full output products from MIRAGE or pyNRC, more so than for just a PSF in a subarray. 

Definitely happy to discuss more. There's almost certainly a more clever way to do this than what I cobbled together over the years.

mperrin:
Hi @JarronL, what do you want to do about this old PR draft from this past spring, which you marked as invalid? Is there any reason to keep this PR open at this time? I'm inclined to close it without merging, if that's OK with you.

JarronL:
Yeah, closing this issue is fine with me.

PR #436, opened 2021-04-01
Update dependencies min versions

shanosborne:
Updating the minimum versions to match `poppy`


mperrin (approved):

mperrin:
Travis passes; the codecov test report "failure" for this has to be confused results from codecov again, because we're just changing dependency versions slightly and there's no way that changes any of the code coverage within webbpsf itself. Ignoring codecov and merging.

PR #440, opened 2021-04-02
Update the OPD linear model so that we can adjust the pixel sampling

kjbrooks:
Up until now, the OPD linear model has been limited to 1024 pixels. This created some limitations, for example when trying to create FGS simulated imaging without Fourier aliasing in coherent mode. 

This PR will:

- [x] add a `npix` parameter to `opds.OTE_Linear_Model_WSS`
- [x] use that `npix` parameter instead of the hardcoded 1024s
- [x] include a new `JWpupil_segments.fits` file for a higher number of `npix`
- [x] Update the linear model to load the right `JWpupil_segments` depending on npix
- [x] Make sure to think about other parts of webbpsf that might use `JWpupil_segments.fits` if that filename gets updated
- [x] Update tests!
- [x] Test in creation of MIRAGE images for FGS


kjbrooks:
@obi-wan76 FYI

webbpsf/opds.py
```diff
@@ -2443,18 +2455,14 @@ def random_unstack(ote, radius=1, verbose=False):
 
 # --------------------------------------------------------------------------------
 
-def segment_primary(infile='JWpupil.fits'):
+def segment_primary(infile='JWpupil.fits', npix=1024):
```
mperrin: This segment_primary function is old and I think not used anymore. 

This has been superseded by `webbpsf.optics.WebbPrimaryAperture()` which can be called with `label_segments=True`.

I think this whole function can just be deleted. (I'd forgotten it was here... Gotta admit it's strange to now be finding code I've written more than a decade ago. 🤦)

mperrin: @kjbrooks First, thanks for working on this! Did you see my comment on this PR earlier today?  Apologies if not.  The `segment_primary` function doesn't need to be updated to work for larger sizes, since there's another better way to set up the segment map using webbpsf.optics.WebbPrimaryAperture()

kjbrooks: Oops.

kjbrooks: This is what happens when I get into a coding mode! I will go ahead and stop worrying about this then!!

mperrin:
@kjbrooks out of curiosity, have you calculated what level of sampling we need to avoid Fourier aliasing in the FGS PSFs? Is it sufficient if we go up by 2x to 2048? Or do we need 4096? I've created versions of the segment map file for both, which I've attached here: 

[JWpupil_segments_RevW_npix2048.fits.gz](https://github.com/spacetelescope/webbpsf/files/6259809/JWpupil_segments_RevW_npix2048.fits.gz)
[JWpupil_segments_RevW_npix4096.fits.gz](https://github.com/spacetelescope/webbpsf/files/6259811/JWpupil_segments_RevW_npix4096.fits.gz)

kjbrooks:
No, I haven't gotten that far yet. I think this is about ready for testing (except for the consistency of the file naming of the JWpupil_segments images) so that would be the next step. I am not sure how to calculate the level of sampling we would need (at least, I haven't thought about it very hard), other than to just start running tests. What do you recommend?

mperrin:
You can do a simple Fourier optics calculation: The aliasing will occur beyond an angle of `npix/2 `_λ/D_. So take the minimum λ for FGS, and the maximum angular size we want to compute a PSF for without aliasing, and solve for `npix`.

kjbrooks:
@mperrin just getting back to this. To clarify, would the "maximum angular size" here be the full size of the large array? I am specifically thinking about image stacking where we have PSFs at the maximum distance while also stacking PSFs in the center.

mperrin:
> @mperrin just getting back to this. To clarify, would the "maximum angular size" here be the full size of the large array? I am specifically thinking about image stacking where we have PSFs at the maximum distance while also stacking PSFs in the center.

Yes that's correct.

mperrin:
Or rather - I should listen to my own advice and never use the word "size" because it's ambiguous. Should say "maximum angular radius" so it's the radius to the maximum outer segment PSF (rather than the diameter of the array)

Worked example: 

1. The large array for JWST PSFs is 800 NIRCam pixels radius = 400 FGS pixels radius = approx 25 arcseconds radius. 
2. Let's take λ_min as 0.6 microns. Therefore λ_min/D = 0.018 arcsec. 
3. 25 arcseconds / 0.018 arcsec per λ_min/D = 1400 λ_min/D

Therefore (as expected), a 1024 pixel OPD is insufficient because npix/2 is only 512. Even 2048 is not sufficient since npix/2 is still less than 1400.

kjbrooks:
With this PR, if we expect anyone to want to run the code for npix larger than 1024 (i.e. 2048, 4096), both the `JWpupil_segments` and `jwst_pupil` files are required for these larger sizes, to run this code successfully. Right now `webbpsf-data` only has both of these files for npix=1024, and is missing the `jwst_pupil` image for npix=4096 (see attached) and the `JWpupil_segments` files for npix=2048 and npix=4096 (see files attache by @mperrin above). I will need to defer to the WebbPSF dev team about if we want to include these files in the next release of `webbpsf-dev` or if we assume that their use will be limited enough that that won't be necessary. This will also impact the tests that I write. @mperrin or @shanosborne any thoughts on this?

[jwst_pupil_RevW_npix4096.fits.gz](https://github.com/spacetelescope/webbpsf/files/6401398/jwst_pupil_RevW_npix4096.fits.gz)

mperrin:
For sure let’s add these in to the data files in the next release! No reason not to make this broadly available, and it will make CI testing easier to also include them, agreed.

mperrin:
To follow up about the data files: @kjbrooks can you please put copies of all the new 2048/4096 data files into the central store folder under tel/share/webbpsf/webbpsf-data-source

We can then make new versions of the data zip files, including the one for use by the CI system on here (we need to do that  anyway for a couple other PRs so it'll be convenient to do all at once)

kjbrooks:
@mperrin I have moved the 3 files mentioned to central store, so that should be ready to go.

webbpsf/opds.py
```diff
@@ -126,8 +130,19 @@ def __init__(self, name='unnamed OPD', opd=None, opd_index=0, transmission=None,
 
         self.segnames = np.asarray([a[0:2] for a in constants.SEGNAMES_WSS_ORDER])
 
-        full_seg_mask_file = os.path.join(utils.get_webbpsf_data_path(), segment_mask_file)
+        if segment_mask_file == 'JWpupil_segments.fits' and npix != 1024: # Be backwards compatible with file naming
```
mperrin: > Be backwards compatible with file naming

FWIW I don't think this is particularly important to keep exactly backwards compatible. It would simplify things to rename the existing file to have a 1024 in the name, so shall we just do that so they're all consistent?

kjbrooks: Sure, I can make that change in the code. Do you want me to change the name of the one in tel/share/webbpsf/webbpsf-data-source?

mperrin: Sure. 

I'm trying to recall why I did not change the filename before. I think in part I was trying to minimize the need for new data files for use by the CI runner - switching the expected name will immediate make the CI system fail since it can't find the new name in the existing data files. But we need to update that now anyway as part of this so I think that's fine.

kjbrooks:
As a note for how the tests are working to show if the aliasing is no longer a problem, here is an example of a simulated FGS image using the default npix=1024, showing evidence of aliasing, and below a newly created FGS simulated image of the same scene, using an npix of 4096. These were created using `mirage`, which utilizes `webbpsf`.
fgs_simulated_image_w_aliasing
fgs_simulated_image_wo_aliasing

webbpsf/opds.py
mperrin: Sure. 

I'm trying to recall why I did not change the filename before. I think in part I was trying to minimize the need for new data files for use by the CI runner - switching the expected name will immediate make the CI system fail since it can't find the new name in the existing data files. But we need to update that now anyway as part of this so I think that's fine.

mperrin:
CI working and updated to include the new 2048 npix files, as well as the renamed 1024 pixel ones.

kjbrooks:
The above stated work has now been completed. 

There are a couple things that need some attention in the review:

1. For the test that I wrote could use some input on how much change we will tolerate between the PSFs that have been created with two different sizes (below is a screen shot of the difference image between the two methods - note the colorbar, units are fractional intensity / pixel)
Screen Shot 2021-05-07 at 1 20 03 PM
2. How much testing coverage would you like to include here? There are tests that I can write for mismatch in file and npix, but I haven't yet included those
3. I have one test that is failing for me locally, `test_opds.test_segment_tilt_signs`, however, upon looking at this test, there are some typos and inconsistencies in the code so while the failure could be because of the code I wrote, I am not yet convinced until I get a confirmation that that test is written as expected.
4. Tests are failing here and it's not clear to me why though it seems to be with the travis build?

mperrin:
FYI @kjbrooks there's some broader issue with Travis that is causing tests to fail across the board. @shanosborne was looking into this but as of a day or so ago it wasn't yet clear what was going on. Broader issue than just this PR, for sure.

webbpsf/webbpsf_core.py
```diff
@@ -885,25 +885,15 @@ def _get_telescope_pupil_and_aberrations(self):
...
-            else:
-                _log.warning("Nonstandard resolution pupil, so linear model for OTE mirror moves is not supported")
-                pupil_optic = poppy.FITSOpticalElement(
-                    name='{} Entrance Pupil'.format(self.telescope),
-                    transmission=pupil_transmission,
-                    opd=opd_map,
-                    planetype=poppy.poppy_core.PlaneType.pupil
-                )
+            # get npix from pupil_transmission
+            npix = int(pupil_transmission.split('npix')[-1].split('.')[0])
```
mperrin: I think this may need generalization since there are cases in which the pupil_transmission variable may already be a FITS HDUList. We may need to set npix above in the options  block for pupil intensity (lines 863-885), so it can be set differently as appropriate depending on what the input is.

webbpsf/webbpsf_core.py
```diff
@@ -885,25 +885,15 @@ def _get_telescope_pupil_and_aberrations(self):
...
-            if 'npix1024' in pupil_transmission:
-                # The linear model is limited to require 1024 pixels right now, so in this case (the default)
-                # we can use that:
-                pupil_optic = opds.OTE_Linear_Model_WSS(
-                    name='{} Entrance Pupil'.format(self.telescope),
-                    transmission=pupil_transmission,
-                    opd=opd_map,
-                    v2v3=self._tel_coords()
-                )
-            else:
```
mperrin, line 899: Yay for getting rid of this if/else case for not being able to support other sizes of OTE model!

webbpsf/opds.py
```diff
@@ -1118,15 +1133,23 @@ def __init__(self, name='Unnamed OPD', opd=None, opd_index=0, transmission=None,
...
             for the field-dependent SM pose aberrations.
+        npix : int
+            Size of OPD: npix x npix
 
         """
 
-        OPD.__init__(self, name=name, opd=opd, opd_index=opd_index, transmission=transmission, segment_mask_file=segment_mask_file)
+        OPD.__init__(self, name=name, opd=opd, opd_index=opd_index, transmission=transmission, segment_mask_file=segment_mask_file, npix=npix)
         self.v2v3 = v2v3
+        self.npix = npix
```
mperrin: I think setting `self.npix` here is redundant since that also gets set in the superclass init.

webbpsf/opds.py
```diff
@@ -1118,15 +1133,23 @@ def __init__(self, name='Unnamed OPD', opd=None, opd_index=0, transmission=None,
...
+
+        # Check that the shape of the OPD that has been passed, matches the npix parameters
+        if self.opd is not None:
+            if not self.opd.shape == (self.npix, self.npix):
+                raise ValueError(f"npix value of {self.npix} does not match shape of OPD provided: {self.opd.shape}")
```
mperrin: Optiona, but maybe this also moves up into the superclass init?  (That said, the other OPD linear model class doesn't get used at all any more and hasn't gotten used since IV&V analyses post JSC, so this class hierarchy is in some ways not as critical to maintain now, and it's equally effective to have this check here)

webbpsf/opds.py
```diff
@@ -1336,7 +1359,7 @@ def _apply_hexikes_to_seg(self, segment, hexike_coeffs, debug=False):
         apmask = np.ones_like(Xc)  # by construction, we're only evaluating this for the good pixels
 
         hexikes = zernike.hexike_basis_wss(x=Xc, y=Yc, nterms=len(hexike_coeffs),
-                                           aperture=apmask)
+                                           aperture=apmask, npix=self.npix) #FIXME: KJB do we need to pass npix here?
```
mperrin: I think not  needed to pass `npix` in this case, because Xc, Yc, and apmask all encode the number of pixels sufficiently. The `hexike_basis_wss` function will use the X and Y coordinates supplied so that sets the number of pixels in the output.

mperrin (commented):
Looks good so far after a review of the code. I'll need to run some tests locally since the Travis setup is bugged out right now.

kjbrooks:
@mperrin, let me know what you are thoughts are on the tests and what kind of tolerance we want to allow on the PSFs produced by the different sized OPDs/pupil masks.

mperrin:
Hi @kjbrooks, here's some test code I can suggest for the test_changing_npix function. You should be able to just add these lines to the end of that function. Tested locally on my machine here, this works.

```

    # Let's check individual pixel values, at least where the PSF is not too dim.
    # Check all pixels which have > 1e-6 of the total flux (we can safely ignore pixels with very low intensity)
    mask = psf_1024[0].data>1e-6
    assert np.allclose(psf_1024[0].data[mask], psf_2048[0].data[mask], rtol=0.01), 'Pixel values differ by more than 1%'

    # Let's check that the total flux in the PSF does not change much.
    #  (A small amount is acceptable and not surprising, since higher resolution improves the fidelity at which
    #   we model light that is scattered by segment edges to very wide angles outside of the simulated PSF FOV)
    assert np.isclose(psf_1024[0].data.sum(), psf_2048[0].data.sum(), rtol=0.005), "PSF total flux should not change much"

    # Let's also check a derived property of the whole PSF: the FWHM.
    # The FWHM should be very close to identical for the two PSFs.
    assert np.isclose(webbpsf.measure_fwhm(psf_1024), webbpsf.measure_fwhm(psf_2048), rtol=0.0001), "PSF FWHM should not vary for different npix"
```

kjbrooks:
@mperrin thanks for the suggestion! I am still getting one failed test locally, `test_segment_tilt_signs`. Are you seeing this as well?

mperrin:
No, all tests passing for me. What version of poppy are you using? My guess is you're using one that's not the most recent. 

One other thing I want to do for the tests is make sure we test the LOM itself (i.e. moving segments) not just the generation of a PSF with no WFE as in the current test. I'm working on code for that here now and will try to add that shortly

kjbrooks:
I have the latest commit of develop installed for poppy. I have uninstalled, checked my branch and reinstalled, and I am still failing that test.

mperrin:
I pushed some additions to the `test_segment_tilt_signs` to run it for 2048 pixels as well as 1024. This runs here locally for me, with the modified versions of the files that I just put in central store. If it still errors can you put a copy of the error log here please?

kjbrooks:
I am still getting errors. And I am also failing the new test that you added, but that seems to be because I don't have the latest webbpsf-data files that you had updated. Error log for the recurring error:
Screen Shot 2021-05-11 at 9 02 26 AM

I do want to point out that even though this error is on my end, this test is not doing things exactly as expected. There is a mismatch between the segment name and assertion message on lines 399 and 429 on this branch. There is also an undefined variable on line 430 which is why I am puzzled as to why this test is passing in CI.

(For reference, my poppy version is 0.9.3.dev123+g8266798)

mperrin:
> There is a mismatch between the segment name and assertion message on lines 399 and 429 on this branch. There is also an undefined variable on line 430 which is why I am puzzled as to why this test is passing in CI.

Ah good catch. The segment name mismatch on line 429 was causing the next couple lines to never get ran, so the undefined variable on line 430 didn't matter. Those lines got accidentally skipped to check the B segment results. I just fixed this and pushed the change. 

> (For reference, my poppy version is 0.9.3.dev123+g8266798)

This is a little surprising to me - I would expect something > 1.0.   We haven't actually made a release 1 yet but I incremented the initial number for development to reflect the sign change, so I would expect something like 1.1dev<blah>. I have 1.1.dev109+g71f23fd for instance.

mperrin (approved):

mperrin:
Tests pass (in CI and locally for me) and this looks good. Thanks again @kjbrooks, this will be very useful

PR #442, opened 2021-04-07
specify jitter arcseconds per axis

robelgeda:


mperrin:
Are there any other places this should be clarified, for instance any of the function doc strings?

robelgeda:
I think this is the only place we mention it. The rest is in Poppy. I will make a quick PR for Poppy now.

robelgeda:
I have added some pointers in the underlying doc strings:

https://github.com/spacetelescope/poppy/pull/427

mperrin (approved):

PR #454, opened 2021-05-03
update data files release script

mperrin:
FYI @shanosborne @kjbrooks 

Update data files release script to 

1. Include the 1024 and 2048-sized pupil files in the minimal data zip we use for Travis CI, while still exclude the larger files
2. Update the master script for where the files should go after being created.


shanosborne (approved):
Looks good to me!

PR #455, opened 2021-05-10
Move CI to github actions, replacing Travis

shanosborne:
PR to replace Travis CI with Github Actions. A few things to note:
- Results of CI can be seen here (they won't show up on the main repo until it's been merged): https://github.com/shanosborne/webbpsf/actions
- Info on PyPi release method can be seen here: https://github.com/spacetelescope/action-publish_to_pypi 
- Note that continue-on-error is not quite the same as allow-failure. There is no one for one replacement of that functionality (yet), but details about the differences can be seen here: https://github.com/actions/runner/issues/2347
- Emails are automatically sent out to everyone subscribed to the repository when a build fails or is canceled.


mperrin (approved):
Looks good, as far as I can tell, and tests are working in the forked version so that seems good enough to go

PR #456, opened 2021-05-10
update MIRI detector scale params to match current SIAF

mperrin:
Minor but significant: Update MIRI pixel scale and rotation to match current SIAF. 

No change in values for other SIs (but I update the comment text to state the SIAF version)


mperrin:
ping @shanosborne for a review on this one (should be very quick!)

shanosborne (approved):
looks good!

PR #464, opened 2021-06-02
Undefined variables in optics.py

JarronL:
Found a couple fairly benign bugs related to variable changes in optics.py (courtesy of MS Visual Studio's PyLance code checking functionality). 

Also converted a print statement at the beginning of NIRCam class init function to a `_log.info()`.


mperrin (approved):
Thanks @JarronL  👍

PR #364, opened 2020-06-17
Include First steps of WFIRST -> Roman renaming

shanosborne:
I changed the repo's README to use Roman rather than WFIRST in all instances and then changed the first instances of "WFIRST" in the docs, but only for the introduction and index pages since those are usually the ones people go to first.

Let me know if you want any more changes to go in now vs when we wait for the full docs/code update.

PR #471, opened 2021-07-21
Release 0.9.2

robelgeda:
This is a soft release to provide the Roman Prism and Grism code.


mperrin:
Hi @robelgeda I think this will also need the GitHub Actions CI migration from PR #455

mperrin:
Since we know the poppy 1.0 release is not back compatible, we should also update the dependency requirements for this release to keep poppy < 1.0, so nobody accidentally ends up with this webbpsf version on top of poppy 1.0 once that comes out. 
I think that can most easily be done by modifying setup.cfg to have `poppy==0.9.2` instead of `poppy >= 0.9.1`. I think alternatively `poppy >=0.9.1, <1.0.0` would also work. 

Also I don't think this comment is needed in the readme: 
> This is a one-off release and is not meant to be built on (the next release will be a major release v1.0.0).

Let's say instead "This release only improves a subset of Roman functionality; additional improvements to both Roman and JWST models will be at the upcoming 1.0.0 major release".

robelgeda:
Sounds good to me @mperrin , we will add these changes. 

It looks like the last data is `v0.9` should the new data release be `0.9.2` (same as the package version) or  `0.9.1`?

mperrin:
New data file version should match package version, yes, so 0.9.2 in this case.

shanosborne:
@ojustino that wasn't a typo in the tox.ini file, so that last commit should be reverted. The issue is that the requirements.txt file is new and needs to be included in this PR. That, or you need to remove the test that calls it (so removing py38-latest-test from the CI yml file)

mperrin (commented):
Couple quick questions, mostly related to making sure the right things are in this branch or not:

- I see there are a lot of removed docs files like "docs/wfirst_figures/fig_cgi_spc_f770.png" - is that all intentional? Those seem to be fully removed rather than moved/renamed to something with Roman in the name?
- I also see removals of the notebooks: `notebooks/WebbPSF-WFIRST_Tutorial.ipynb` and `notebooks/wfirst_cgi_demo.ipynb`. Those too seem to be deleted entirely, which seems incorrect to me. 
- Can you explain also the deletion of the test `test_WFI_includes_aberrations`?

robelgeda:
> Couple quick questions, mostly related to making sure the right things are in this branch or not:
> 
> * I see there are a lot of removed docs files like "docs/wfirst_figures/fig_cgi_spc_f770.png" - is that all intentional? Those seem to be fully removed rather than moved/renamed to something with Roman in the name?
> * I also see removals of the notebooks: `notebooks/WebbPSF-WFIRST_Tutorial.ipynb` and `notebooks/wfirst_cgi_demo.ipynb`. Those too seem to be deleted entirely, which seems incorrect to me.
> * Can you explain also the deletion of the test `test_WFI_includes_aberrations`?

Something went wrong with the git history where those files did not get copied over. I am going to re-do this step by hand to make sure we get it right. Thank you for checking that!

robelgeda:
This PR is re-structured to only include the Prism and Grism code. It also includes any supporting code from v1.0.0. The following PRs have been included in this release:
- #416 
- #455 

Note that this PR does not include #364, thus the Roman code remains WFIRST.

docs/relnotes.rst
```diff
@@ -30,6 +30,33 @@ Road Map for Future Releases
...
+    -  Grism
+    -  Prisim
+- New `WFI.override_aberrations(aberrations_path)`: Overrides and locks the current aberrations with aberrations at `aberrations_path`. Lock means changing the filter/mode has no effect on the aberrations. [:pr:`416`, :pr:`471`, :user:`robelgeda`]
+- New `WFI.reset_override_aberrations()`: Releases `WFI.override_aberrations` lock and start using the default aberrations. [:pr:`416`, :pr:`471`, :user:`robelgeda`]
+- New Tests for mode and filter switching. [:pr:`416`, :pr:`471`, :user:`robelgeda`]
+- New Field point nearest point approximation (extrapolation). [:pr:`416`, :pr:`471`, :user:`robelgeda`]
+
+**Software and Package Infrastructure Updates:**
+
+- This minor release uses CircleCI and removes TravisCI. [:pr:`455`, :user:`shanosborne`, :pr:`471`, :user:`robelgeda`]
```
mperrin: Shouldn't this say "Github Actions" not CircleCI? I don't think we are using CircleCI, are we?

mperrin: I am just going to fix this directly myself since it's a small thing

mperrin (approved):
Looks good (except for one small doc text issue that I fixed myself already)

PR #472, opened 2021-07-23
Merge and deconflict from stable 0.9.2 to develop

mperrin:
Needed after PR #471 (Release 0.9.2) to get those commits into the develop branch, such that future merges of develop to stable can be done cleanly. 

No actual functionality changes intended here in `develop` - this PR is purely for git-deconflicting changes which we now have multiple copies of.


mperrin:
Cancelling this PR - for git reasons it'll work better if I do this from the `release-0.9.2` branch rather than from `stable`.

PR #473, opened 2021-07-23
Merge and deconflict from develop-0.9.2 to develop

mperrin:
Needed after PR #471 (Release 0.9.2) to get those commits from `develop-0.9.2` into the `develop` branch, such that future merges of develop to `stable` can be done cleanly.

No actual functionality changes intended here in develop - this PR is purely for git-deconflicting changes which we now have multiple copies of.


mperrin:
Note, this replaces #472 . I realized it was better to merge from `develop-0.9.2` to `develop` rather than from `stable` since trying to merge from stable would end up needing some deconflicting merge commit to go onto `stable` which we want to avoid...

mperrin:
Question: who's going to take responsibility for this branch? It's not going to be @robelgeda since he's headed off to grad school. @ojustino would you be able to take a look at what needs to be done here?

ojustino:
@mperrin: @robelgeda and I discussed that I would inherit responsibility for this pull request. I will look into what needs to be done here.

mperrin:
@ojustino hi, I wanted to check in on whether you had a plan or timeline for when you might be able to work on this branch and get it merged in, please. Any chance we could get to this within the next few weeks?  This will eventually become a blocker for the 1.0 release of webbpsf, so we should not let this PR hang open indefinitely. Thanks!

ojustino:
Hi @mperrin, this is still on my list, but the work on Cycle 9 and v0.9.2 earlier this summer took me a couple of months ahead of schedule for allotted to Roman. I'll check with my manager and let you know, but the earliest is probably the week after next.

mperrin (commented):
This would be an approval of this PR, except since I originally created the PR, the Github UI won't let me approve it myself. But consider it approved :-)

Thanks @ojustino!

PR #474, opened 2021-07-29
LRE4 and MIMF KDP Practice 2: add options to LookupTableOPD for importing Ball-delivered OPDs

mperrin:
Another variant on the input data for LookupTableOPD, as used in LRE4 for OTE 2.52 and OTE 26. 

These should have been pushed and merged in #328 but weren't, by accident on my part. 

This isn't needed in any way for LRE5, but I'm going to go ahead and push in case it's ever needed again, perhaps for the next MIMF KDP. 

Edit: Now also adding support for the new OPD files for us in the next MIMF KDP practice, too.

### DIFF
--- a/webbpsf/optics.py
+++ b/webbpsf/optics.py
@@ -1698,7 +1698,7 @@ class LookupTableFieldDependentAberration(poppy.OpticalElement):
     """
 
     def __init__(self, instrument, field_points_file=None, phasemap_file=None,
-                 which_exercise='MIMF_KDP',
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Changes look good to me. Is there a similar capability for OTE-25.2 from the MIMF-KDF_Sept2021? I see the path for LRE4

mperrin:
Yes this includes both the paths for the LRE4 and the MIMF KDP Practice, in different parts of the if statement

PR #480, opened 2021-08-24
Update CI

shanosborne:
2 part PR to fix failing CI

1. Explicitly define the `detector_oversample` key in the `options` dictionary when building PSF objects by hand for these tests. This is now needed because of the poppy change here: https://github.com/spacetelescope/poppy/pull/453/

2. Update the CI yaml file to correctly pull the requirements file in our latest-dependencies test. (Before, it was failing because for some reason, the CI was now looking for the requirements.txt file in the tmp/ directory. So I fixed it by copying the file into the tmp/ directory first).

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -65,8 +65,14 @@ jobs:
     - name: Check conda info
       run: conda info
 
+    - name: Run tests with requirements file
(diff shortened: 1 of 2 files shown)


mperrin (approved):
Looks good - thanks @shanosborne for figuring out these fixes

PR #482, opened 2021-08-25
Extend MIRI Lyot occulter to edge of FoV

### LINKED ISSUE #481: MIRI Lyot focal plane mask seems incomplete
I've tried to make an attenuation map of all MIRI coronagraphs by moving a source position in each and every pixels of the detector for each coronagraphs.

For the Lyot Coronagraph, I ended up with a weird map, the "vertical bar" (for lack of a better word) was not going from top to bottom, but rather a small box around the center. 

Maybe I'm missing something in my input parameter that show there, though.
I've attached my end result to show you what I mean.

ds9

JarronL:
Default height of bar occulter object is 10". MIRI Lyot mask has a full field of view of 30". Extend bar occulter slightly beyond FoV (31").


mperrin (approved):

mperrin:
rebased onto current develop, to get the CI fixes into this branch

PR #488, opened 2021-09-24
Added distortion/crop arguments to Roman calc_psf

### LINKED ISSUE #315: WFIRST instruments' calc_psf functions don't support distortion
The WFIRST instruments crash if you try to pass the distortion parameter to them in calc_psf.

For ease of programming to the API, I'd prefer to have them at least accept a distortion keyword, even if it does nothing or needs to be set to False.

ojustino:
Would resolve #315.

This pull request gives the `RomanInstrument` class its own version of `calc_psf()`, like in `JWInstrument`. It introduces the `add_distortion` and `crop_psf` arguments to Roman's `calc_psf()`, though they raise attribute errors for the moment as we figure out the current WFI implementation's support for distortion.

### DIFF
--- a/webbpsf/roman.py
+++ b/webbpsf/roman.py
@@ -288,13 +288,11 @@ class RomanInstrument(webbpsf_core.SpaceTelescopeInstrument):
         Parameters
         ----------
         add_distortion : bool
-            If True, will add 2 new extensions to the PSF HDUlist object. The 2nd extension
-            will be a distorted version of the over-sampled PSF and the 3rd extension will
-            be a distorted version of the detector-sampled PSF.
+            Included for API compatibility with the JWST instrument classes, but has no
+            effect on the results for Roman WFI PSF calculations.
         crop_psf : bool
-            If True, when the PSF is rotated to match the detector's rotation in the focal
-            plane, the PSF will be cropped so the shape of the distorted PSF will match it's
-            undistorted counterpart. This will only be used for NIRCam, NIRISS, and FGS PSFs.
+            Included for API compatibility with the JWST instrument classes, but has no
+            effect on the results for Roman WFI PSF calculations.
 
         """
 
--- a/webbpsf/roman.py
+++ b/webbpsf/roman.py
@@ -17,6 +17,7 @@ from astropy.io import fits
 import astropy.units as u
 import logging
 
+from . import utils
 from . import webbpsf_core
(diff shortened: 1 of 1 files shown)


ojustino:
Leaving a reminder to check whether introducing `utils.combine_docstrings` to `RomanInstrument` so `calc_psf()` can inherit the rest of its docstring from `poppy` has unintended side effects on the class' other attributes/methods.

mperrin:
> though they raise attribute errors for the moment as we figure out the current WFI implementation's support for distortion.

In some ways the WFI implementation is already sophisticated enough to include much of this already. In particular the use of different pupil files per each detector means the pupil model _already includes_ the field-dependent pupil distortion. (At least roughly in one discrete step per detector, though not tracking the smooth variations with field within a single detector. Potentially that could be done by interpolating or morphing between the provided pupil files in a smooth way).   This is not the case for the JWST models, for which we always use an idealized JWST OTE entrance pupil; for those the distortion terms have to be modeled separately. 

Makes sense for this PR to focus on ensuring API compatibility between the JWST and Roman classes. But I don't think the actual numerical implementation of distortion models from SIAF as used in the JWST classes will ever need to be ported to Roman.

webbpsf/roman.py
```diff
@@ -276,6 +278,42 @@ def __init__(self, *args, **kwargs):
...
+        Parameters
+        ----------
+        add_distortion : bool
+            If True, will add 2 new extensions to the PSF HDUlist object. The 2nd extension
+            will be a distorted version of the over-sampled PSF and the 3rd extension will
+            be a distorted version of the detector-sampled PSF.
+        crop_psf : bool
+            If True, when the PSF is rotated to match the detector's rotation in the focal
+            plane, the PSF will be cropped so the shape of the distorted PSF will match it's
+            undistorted counterpart. This will only be used for NIRCam, NIRISS, and FGS PSFs.
```
mperrin: Let's modify the doc string text here to say "These parameters are included for API compatibility with the JWST instrument classes, but have no effect on the results for WFI PSF calculations."

mperrin (approved):
Hi @ojustino, I see this PR has been open for a while and no one was tagged yet to review it. Looks fine to me. I suggest some wording changes in the doc string text. I can add that if you want, or you can, and then let's go ahead and get this merged.

PR #493, opened 2021-10-18
Use "==" instead of "is" for string comparisons

JarronL:
Comparisons of some variables to strings were using `is` rather than `==`, likely due to changes from `None` to empty strings `""`.


mperrin (approved):
Thanks! Good catch.

PR #497, opened 2021-11-18
bug fix: avoid filling OPD outer regions with NaNs when using the global zernike option

mperrin:
Trivial one-line fix.

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -1395,6 +1395,7 @@ class OTE_Linear_Model_WSS(OPD):
 
         perturbation = poppy.zernike.opd_from_zernikes(self._global_zernike_coeffs,
                                                        npix=self.npix,
+                                                       outside=0,
                                                        basis=poppy.zernike.zernike_basis_faster)
         # Add perturbation to the opd
         self.opd += perturbation


mperrin:
@shanosborne this one's even more minor :-)

shanosborne:
Looks like pysiaf updated their requirement for astropy. Not sure why its an exact version though, and not a `>=`. I'll ask Mees about it.

>  The conflict is caused by:
    The user requested astropy 5.1.dev155+gde53bdc16 (from git+git://github.com/astropy/astropy)
    poppy 1.1.dev211+g1951496 depends on astropy>=4.0.0
    pysiaf 0.13.0 depends on astropy==4.3.1

Update: Mees is looking into the pysiaf requirement (he set it fixed to solve a different problem, so now he's going back and looking again).

shanosborne:
@mperrin I did notice another issue with the tests. The released poppy & pysiaf test (which is allowed to fail), actually failed in the defocus PR and I missed it, so it's also failing here. It's not showing up as a red x because that's how GHA works ("allowed failure" tests show a green check even when they fail, there's an open GHA ticket for it and people are mad about it).

So the `test_defocus` test needs to be updated (either here or in another PR) - not sure if the issue is the newest pysiaf or the newest poppy.

shanosborne:
The pysiaf problem should be fixed - I re-triggered the tests to check

mperrin:
I believe the failure of the defocus test (caused by the other PR, not this one) is indeed an allowed failure and we should ignore it. It's a consequence of the sign convention differences between poppy 0.9.x and 1.0.  OK to ignore, since we know poppy 1.0 will come out before webbpsf 1.0.

shanosborne (approved):
With all the tests finally passing, this is good to go!

PR #499, opened 2021-11-18
Rename 'P750L LRS grating' -> 'P750L', per #477

### LINKED ISSUE #477: LRS Grating -> Prism
This may be irrelevant to the code, but I wanted to flag that the MIRI LRS does not have gratings, it uses a double prism for dispersion. I'm only just digging into the code to see how it works exactly for LRS, so maybe this is just a vocabulary issue that has no impact on the code (as you don't model the dispersion), but I wanted to flag it in case you had made any grating-specific assumptions.

mperrin:
Fix #477, in which @skendrew pointed out that the MIRI LRS dispersing element is not a grating, so webbpsf shouldn't label it as such.

### DIFF
--- a/docs/jwst.rst
+++ b/docs/jwst.rst
@@ -469,7 +469,7 @@ LRS Spectroscopy
 ----------------
 
 WebbPSF includes models for the LRS slit and the subsequent pupil stop on the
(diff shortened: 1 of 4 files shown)


shanosborne (approved):
Looks good. The only thing would be to also update the P750L LRS grating info in the docs. I see it mentioned in `more_examples.rst`, `jwst.rst`, and `references.rst`.

PR #509, opened 2021-12-08
Updating bar mask offsets directly from pySIAF

kammerje:
Based on discussions with Marshall Perrin and Mario Gennaro we noticed that WebbPSF uses outdated bar mask offsets for the NIRCam bar mask coronagraphs. We have now updated WebbPSF to grab the most recent offsets directly from pySIAF.

### DIFF
--- a/webbpsf/optics.py
+++ b/webbpsf/optics.py
@@ -708,32 +708,6 @@ class NIRCam_BandLimitedCoron(poppy.BandLimitedCoron):
     allowable_kinds = ['nircamcircular', 'nircamwedge']
     """ Allowable types of BLC supported by this class"""
 
-    # Offsets along the bar occulters are based on
-    # outputs from John Stansberry's filt_baroffset.pro
-    # See https://github.com/mperrin/webbpsf/issues/63
-    # EDIT: updated on 8 Dec 2021 to grab offsets directly from pySIAF
-    # offset_swb = {"F182M": -1.856,
-    #               "F187N": -1.571,
-    #               "F210M": -0.071,
-    #               "F212N": 0.143,
-    #               "F200W": 0.232,
-    #               'narrow': -8.00}
-    """ Offsets per filter along SWB occulter for module A, in arcsec"""
-
-    # offset_lwb = {"F250M": 6.846,
-    #               "F300M": 5.249,
-    #               "F277W": 5.078,
-    #               "F335M": 4.075,
-    #               "F360M": 3.195,
-    #               "F356W": 2.455,
-    #               "F410M": 1.663,
-    #               "F430M": 1.043,
-    #               "F460M": -0.098,
-    #               "F480M": -0.619,
-    #               "F444W": -0.768,
-    #               'narrow': 8.0}
-    """ Offsets per filter along LWB occulter for module A, in arcsec"""
-
     def __init__(self, name="unnamed BLC", kind='nircamcircular', module='A', nd_squares=True,
                  bar_offset=None, auto_offset=None, **kwargs):
         super(NIRCam_BandLimitedCoron, self).__init__(name=name, kind=kind, **kwargs)
@@ -761,24 +735,11 @@ class NIRCam_BandLimitedCoron(poppy.BandLimitedCoron):
             raise NotImplementedError("invalid name for NIRCam occulter: " + self.name)
 
         # EDIT: updated on 8 Dec 2021 to grab offsets directly from pySIAF
-        self.offset_swb = {"F182M": self.get_baroff_from_siaf("F182M", channel='SW'),
-                           "F187N": self.get_baroff_from_siaf("F187N", channel='SW'),
-                           "F210M": self.get_baroff_from_siaf("F210M", channel='SW'),
-                           "F212N": self.get_baroff_from_siaf("F212N", channel='SW'),
-                           "F200W": self.get_baroff_from_siaf("F200W", channel='SW'),
-                           'narrow': self.get_baroff_from_siaf('narrow', channel='SW')}
-        self.offset_lwb = {"F250M": self.get_baroff_from_siaf("F250M", channel='LW'),
-                           "F300M": self.get_baroff_from_siaf("F300M", channel='LW'),
-                           "F277W": self.get_baroff_from_siaf("F277W", channel='LW'),
-                           "F335M": self.get_baroff_from_siaf("F335M", channel='LW'),
-                           "F360M": self.get_baroff_from_siaf("F360M", channel='LW'),
-                           "F356W": self.get_baroff_from_siaf("F356W", channel='LW'),
-                           "F410M": self.get_baroff_from_siaf("F410M", channel='LW'),
-                           "F430M": self.get_baroff_from_siaf("F430M", channel='LW'),
-                           "F460M": self.get_baroff_from_siaf("F460M", channel='LW'),
-                           "F480M": self.get_baroff_from_siaf("F480M", channel='LW'),
-                           "F444W": self.get_baroff_from_siaf("F444W", channel='LW'),
-                           'narrow': self.get_baroff_from_siaf('narrow', channel='LW')}
+        self.siaf = pysiaf.Siaf('NIRCAM')
+        self.offset_swb = {filt: self.get_bar_offset_from_siaf(filt, channel='SW')
+                           for filt in ["F182M", "F187N", "F210M", "F212N", "F200W", 'narrow']}
+        self.offset_lwb = {filt: self.get_bar_offset_from_siaf(filt, channel='LW')
+                           for filt in ["F250M", "F300M", "F277W", "F335M", "F360M", "F356W", "F410M", "F430M", "F460M", "F480M", "F444W", 'narrow']}
 
         if bar_offset is None and auto_offset is not None:
             offsets = self.offset_swb if self.name.lower() == 'maskswb' else self.offset_lwb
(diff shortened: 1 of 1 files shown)


webbpsf/optics.py
```diff
@@ -709,26 +711,27 @@ class NIRCam_BandLimitedCoron(poppy.BandLimitedCoron):
     # Offsets along the bar occulters are based on
     # outputs from John Stansberry's filt_baroffset.pro
     # See https://github.com/mperrin/webbpsf/issues/63
-    offset_swb = {"F182M": -1.856,
-                  "F187N": -1.571,
-                  "F210M": -0.071,
-                  "F212N": 0.143,
-                  "F200W": 0.232,
-                  'narrow': -8.00}
+    # EDIT: updated on 8 Dec 2021 to grab offsets directly from pySIAF
```
mperrin: Let's delete rather than comment out these lines, please.

webbpsf/optics.py
```diff
@@ -757,6 +760,26 @@ def __init__(self, name="unnamed BLC", kind='nircamcircular', module='A', nd_squ
         else:
             raise NotImplementedError("invalid name for NIRCam occulter: " + self.name)
 
+        # EDIT: updated on 8 Dec 2021 to grab offsets directly from pySIAF
```
mperrin, line 737: We can turn this repetitive code into a dict comprehension: 

```
self.offset_swb = {filt: self.get_baroff_from_siaf(filt, channel='SW') 
                   for filt in ["F182M","F187N""F210M","F200W", 'narrow']}
```

webbpsf/optics.py
```diff
@@ -773,6 +796,23 @@ def __init__(self, name="unnamed BLC", kind='nircamcircular', module='A', nd_squ
         else:
             self.bar_offset = None
 
+    def get_baroff_from_siaf(self, filt, channel='LW'):
```
mperrin: minor: `get_bar_offset_from_siaf` would be slightly more clear as a function name, since it's all real words (except for "siaf" of course)

webbpsf/optics.py
```diff
@@ -773,6 +796,23 @@ def __init__(self, name="unnamed BLC", kind='nircamcircular', module='A', nd_squ
...
+        """ Get bar offset directly from SIAF.
+        
+        """
+        
+        siaf = pysiaf.Siaf('NIRCAM')
```
JarronL: This effectively initializes the NIRCam SIAF every time `get_baroff_from_siaf` is called, which produces a significant overhead (approximately 230 msec per filter call on my laptop). Instead, place this in the `__init__` function so it's only called once, `self.siaf = pysiaf.Siaf('NIRCAM')`, then refer to `self.siaf` in `get_baroff_from_siaf`.

webbpsf/optics.py
```diff
@@ -709,26 +711,27 @@ class NIRCam_BandLimitedCoron(poppy.BandLimitedCoron):
...
-                  "F212N": 0.143,
-                  "F200W": 0.232,
-                  'narrow': -8.00}
+    # EDIT: updated on 8 Dec 2021 to grab offsets directly from pySIAF
+    # offset_swb = {"F182M": -1.856,
+    #               "F187N": -1.571,
+    #               "F210M": -0.071,
+    #               "F212N": 0.143,
+    #               "F200W": 0.232,
+    #               'narrow': -8.00}
```
JarronL: Add a comment of new values from current pysiaf version. For instance, v0.13.0 gives:
```python
{'F182M': -1.742781239401549,
 'F187N': -1.5437104345990476,
 'F210M': -0.03369518613985725,
 'F212N': 0.14397080350195612,
 'F200W': 0.19604610088955918,
 'narrow': -8.05266403068413}
```

webbpsf/optics.py
```diff
@@ -709,26 +711,27 @@ class NIRCam_BandLimitedCoron(poppy.BandLimitedCoron):
...
+    #               "F277W": 5.078,
+    #               "F335M": 4.075,
+    #               "F360M": 3.195,
+    #               "F356W": 2.455,
+    #               "F410M": 1.663,
+    #               "F430M": 1.043,
+    #               "F460M": -0.098,
+    #               "F480M": -0.619,
+    #               "F444W": -0.768,
+    #               'narrow': 8.0}
```
JarronL: Similar to previous comment, but for LW:
```python
{'F250M': 6.565362470429261,
 'F300M': 5.041863295787578,
 'F277W': 4.917006143111985,
 'F335M': 3.8745694991966504,
 'F360M': 3.0569949990904455,
 'F356W': 2.326902185492543,
 'F410M': 1.6218640501805301,
 'F430M': 0.9980127928784266,
 'F460M': -0.09355467488052485,
 'F480M': -0.8544070167999559,
 'F444W': -0.723448244191664,
 'narrow': 8.301714886773997}
```

JarronL (commented):
This looks pretty straightforward.  A few suggestions, including a couple comments comparing new values as well as a way of improving performance.

mperrin (changes requested):
Looks very good, thanks @kammerje! Some relatively minor suggestions below.

webbpsf/optics.py
JarronL: Obviously, @mperrin's suggestions override mine, here.

webbpsf/optics.py
JarronL: Same as above...

webbpsf/optics.py
mperrin: oh good catch, that's a significant performance hit when it all adds up. 

We could go a step further, since the siaf is read and parsed when the instrument class is set up. There's a `.siaf` attribute on any `NIRCam()` class instance, which could potentially be passed in to this. But that may be out of scope for this PR. Getting it down to a single call however is definitely worth doing.

webbpsf/optics.py
mperrin: I guess I'm curious what value you think it would have to include the numerical values as a comment text? Just for reference later?  If anything, I'd say it's worth putting a table of these into JDox, but that's out of scope for this PR :-)

webbpsf/optics.py
mperrin: and similar for the below though I didn't write that in my comment

webbpsf/optics.py
JarronL: My thinking was that if the old ones were going to be commented out, we should add updated ones (with a version attachment) for reference as a comparison for clarity on how things changed.

JarronL: But removing all reference to the numbers is fine with me. A JDox table is the appropriate place to track this.

mperrin:
@JarronL, not directly related to this PR, but at some point maybe you could help enlighten us on how the wide filter positions are calculated? I don't understand how the wavelengths are weighted for the broadband. It seems to me the W filters are all systematically shifted towards the wider edge of the mask compared to the M filters, and that's not obvious to me why that should be the case. In particular, I'd expect broadband filter averages to be tilted toward shorter wavelengths, because there's more stellar photons at those wavelengths. 

In any case we should use the values in the SIAF; it's just for my own understanding I'd like to better follow the logic for how these are derived.

webbpsf/optics.py
JarronL: Missing a comma between F187N and F210M. 🙂

JarronL:
Hmm... this is more a John Stansberry questions, but I'll give it a shot...

According to the NIRCam Coronagraphic Ops document (NIRCam_CoronagraphOps_V4.pdf), the wide band filters are treated a little differently compared to the medium band filters. Medium band filter locations take into account targets with stellar spectral types. But, the wide band "equations must be modified to account for the longer wavelengths they admit relative to those admitted by the F210M and F460M filters. For this purpose, the reference wavelength of interest is defined by the long half-transmission point, lambda_1/2, of the F210M and F460M filters, which are 2.196 and 4.747 um respectively."

For instance, the long-half-power point for F444W is 4.992, so the location was calculated as -3.16 * (4.992 - 4.747) = -0.7742. This is from 2016, though, so things have obviously changed somewhat given the slight differences relative to the pysiaf values.

In short, I guess they wanted to be sure that the longer wavelength contribution from the filters were fully occulted even if that meant the mask position is oversized for the shorter wavelength portions of the filter.

kammerje:
Thanks, I added all of your suggested changes.

JAStansberry:
The use of a different (longer) effective wavelength for calculating the offsets for the W filters came from John Krist, based on modeling he did.
It would actually be useful to use the current version of WebbPSF to model the contrast as a function of displacement along the bar. John was using a pretty notional OPD map, I think. 

Also, looking back at my script for calculating the offsets, the were based on pixel scales of 32 and 65 mas. My guess is that the difference due to that probably can't be measured, but we may as well use the right values. Even better, use the scale appropriate to the location of the bar masks and taking into account the new distortion model for the coronagraph FOV.

mperrin:
Thank you @kammerje!  Much appreciated. I'll go ahead and merge this now. 

@JAStansberry and @JarronL, we could indeed do some contrast calculations for different MASKLWB offsets as you propose. (I'm about to finally update the OPD maps for as-built latest & greatest, too, so these would be much better than earlier notional OPDs). That's probably better done as a task by Jarron or someone else on the NIRCam IDT rather than any of us in the webbpsf or telescope teams - I'd be happy to advise on how to do the modeling though. It sounds like the choices John Krist made may have been motivated by getting as deep as possible contrast, at a cost of inner working angle. Which may well be the right call but it's something to clarify and convey to users. 

Tagging @juliengirard too - FYI see the conversation on this PR about MASKLWB offsets.

mperrin (approved):

JarronL:
Sounds good. @mperrin, if you could make a new PR and tag me as the responsible, I can look into it ::waves hand vaguely:: sometime in the near future. Would like to understand how the most recent filter-dependent bar mask reference positions as defined in SIAF were calculated, though, which probably requires some combination of @mgennaro, @JAStansberry, @juliengirard.

mperrin:
Done - see spacetelescope/stpsf#64, @JarronL

PR #510, opened 2021-12-08
Fix spacing issue in docs

shanosborne:
Docs had problem where in the API section there were no spaces in between the variable name and the type. This PR adds some CSS to build on the stsci rtd theme to add more helpful formatting in the form of "variable : type"


mperrin (approved):
👍

PR #512, opened 2021-12-09
JWST OTE OPD updates for Release 1.0.0

mperrin:
Updates to JWST OTE OPDs

- [x] some small enhancements to configuration options and tests for the OTE field dependence model
- [x] Bug fix to allow use with the JWST OTE linear model of datacube OPD files and specifying a particular slice within the cube
- [x] Update OTE optical budget information for consistency with Revision AA, the last pre-launch revision of the budget and predicted performance.  
- [x] Update OPD files with new OPDs for the as-built JWST primary mirror, derived from the PSR2020 integrated modeling cycle (Predicted Stability Requirements 2020. Internal data reference at STScI: see files in `tel/share/PSR2020`). In particular we use a model file made by Joe Howard and provided to Chris Stark for the coronagraphy contrast performance WG.
- [x] Update JWST classes to use the new OPD files
- [x] Update some unit tests to use the prior OPD files, for consistency with past test results. 
- [x] Make new data zip files, and update CI build to use them.


mperrin:
This OPD change fixes the long-standing issue in which SI WFE was double-booked within the "OTE" OPDs, due to a miscommunication some years ago. This was recognized a while ago but until now didn't have time to update. 

Comparison of before/after of models for NIRCam using the prior 'predicted' and new 'prelaunch_predicted' OPDs. Look in particular at the upper left overall OPD panels, and the OTE total static WFE just below. The revised OPDs are noticeably more consistent with the budgeted predicted performance. 


**Prior Version:** 

Unknown-6

**This version:**

Unknown-7

(For this particular field point the total system WFE is shown as better than the budgeted WFE; that is expected because as-built NIRCam has significantly lower WFE at this field point than was budgeted for SI WFE. There are many instances like this. Recall the optical budget is a statistical document averaged over field points, whereas webbpsf models specific field points.) 


Comparisons of encircled energy in the resulting PSFs show a small but significant improvements in EE: 

Unknown-9
Unknown-8

mperrin:
FYI - I tagged lots of folks as "reviewers" on this, for general awareness since this is a big change in the OPD models used in webbpsf. If you're @grbrady @obi-wan76 @Skyhawk172 or @JarronL, consider this just an FYI to take a look and be aware of. @shanosborne I'm hoping you will be able to provide an actual code review. :-)

mperrin:
Right now I expect all tests will fail, until I update the data files... Standby. 

edit: Oh actually no, I was more clever than that. It all just falls back and uses the prior OPD rather than the new one for running the tests.

shanosborne (approved):
Looks good- I downloaded the new data fine and it pulled the right OPD file, so that's all good. 

The only thing I found was that `OTE_Linear_Model_Elliott`'s docstring still references `slice` instead of `opd_index` even though the parameter in `__init__` has already been switched to `opd_index`, so the docstring should be updated. 

Also, do we want to update the tests so they use the newer OPDs? Is there a benefit to doing that?

mperrin:
Thanks Shannon. I updated that one doc string. (with [ci skip], which is why the tests did not run again afterwards. But they did pass before!)

The tests do in fact use the new OPDs. I had thought I would have to modify the code to make some of them use the old OPDs for consistency, particularly the coronagraph tests which depend sensitively on the WFE. But it turns out we solved that problem years ago by having the coronagraph tests all use zero WFE (pupilopd=None). So, no action needed for that after all. The other tests that aren't as sensitive to WFE all do use the default OPDs, which now are these new ones. All set. 

Thanks!

PR #516, opened 2021-12-10
Drop CGI references in Roman code

### LINKED ISSUE #445: Roman Coronagraph Instrument
Rename CGI to "Roman Coronagraph Instrument" (Roman team request)

ojustino:
Would resolve #445.

References to the CGI have been changed to "Coronagraph" (including the `CGI` instrument class) or "Coronagraph Instrument." There are some images and outputs that still reference the CGI because we'd need an analogous pull request to `poppy` to change them.


webbpsf/roman.py
```diff
@@ -871,7 +871,7 @@ def unlock_pupil_mask(self):
         self._update_pupil() # reset pupil mask
 
 
-class CGI(RomanInstrument):
+class Coronagraph(RomanInstrument):
```
mperrin: My one hesitancy is just calling this class `Coronagraph` is maybe potentially confusing or ambiguous, as there are also coronagraphs on JWST. I'd suggest going with `RomanCoronagraph` to be explicit and unambiguous, even though it's a big longer.

webbpsf/roman.py
```diff
@@ -911,7 +911,7 @@ class CGI(RomanInstrument):
         'DISKSPC_F721': ('IMAGER', 'F721', 'DISKSPC', 'DISKSPC_F721_ANNULUS', 'LS30D88')}
 
     def __init__(self, mode=None, pixelscale=None, fov_arcsec=None, apply_static_opd=False):
-        super(CGI, self).__init__("CGI", pixelscale=pixelscale)
+        super(Coronagraph, self).__init__("CGI", pixelscale=pixelscale)
```
mperrin: This could be just `super().__init__` now. The older syntax is leftover from Py2.7. [optional change, not a big deal either way]

mperrin (approved):
Thanks for doing this so swiftly, @ojustino, for an issue we'd completely forgotten about.  Looks good to me, though I think I suggest we go with `RomanCoronagraph` instead of just `Coronagraph` to be unambiguous. 

> There are some images and outputs that still reference the CGI because we'd need an analogous pull request to poppy to change them.

I'm not sure I follow. There's nothing in poppy about any specific instrument or telescope at all; that package has the generalized optical physics models. Can you explain more what parts you think might need to be changed in poppy?

webbpsf/roman.py
ojustino: That makes sense on further thought. I followed through with `RomanCoronagraph` in a new commit.

webbpsf/roman.py
ojustino: I made this change throughout the file for consistency's sake.

ojustino:
When I run `calc_psf()` for the Coronagraph Instrument with logging on, I see messages attributed to "[poppy]" that reference it, and resulting images do, too. (`notebooks/roman_coronagraph_demo.ipynb` has some examples.) On further thought, of course `poppy` wouldn't have CGI references.

The closest guess I have at this moment is that there's somewhere in `poppy` where `BaseOpticalSystem` or a descendant has its `name` attribute has been set to "Roman+CGI", possibly because the directory it's reading from in the data files hasn't been renamed (`webbpsf-data-source/CGI`).

mperrin:
> The closest guess I have at this moment is that there's somewhere in poppy where BaseOpticalSystem or a descendant has its name attribute has been set to "Roman+CGI",

Yes, that was it! Good catch.  We were passing in that string as the name when creating the OpticalSystem instance (webbpsf_core line 410). That in turn was coming from the name attribute on the RomanCoronagraph instance, which was still being set to the string "CGI". I just added one more commit to this branch to fix that.

mperrin:
Ah, shoot. It's not that easy. Changing that string changes the folder path for where it looks on disk to get the data files. I had hoped to let that slide. But I'll go ahead and update that now for the v1.0.0 data files.

PR #517, opened 2021-12-10
Roman documentation and coronagraph demo changes

ojustino:
These are changes that bring our Roman Coronagraph docs and example notebook up to date with the changes made in #516.

The bottom-middle panel of `docs/roman_figures/fig_coronagraph_spc_f770.png` comes out looking different than before for reasons I haven't found yet.


mperrin (approved):
Excellent, thank you. Looks good. No need to worry about the slight display changes in that one intermediate plane. I think that may be related to the internal phase sign convention change in poppy 1.0 vs previous. In any case not something we need to look into further now. But thank you for being thorough and pointing it out!

PR #530, opened 2022-02-01
Add pytest-astropy-header to conftest.py

shanosborne:
Update `conftest.py` to work with `astropy` > 5.0 - this requires calling `pytest-astropy-header`. This was causing the CI in the astropy development build to fail

I also added a setting in `pyproject.toml` to make sure that when `pytest` runs, it only searches the `tests/` directory for test files. This used to work on its own, but recently has been searching the wrong folders.


mperrin (commented):
👍 Thanks @shanosborne

mperrin (approved):

PR #537, opened 2022-03-09
Fix bug in display_psf_grid

shanosborne:
Fixes an error raised in a help desk question. Moves `vmin` and `vmax` definitions inside the normalization object

> Hi, I'm receiving the following error (below) for the following command:
> webbpsf.gridded_library.display_psf_grid(psf_webbpsf_grid, zoom_in=False, figsize=(14, 14))
> 
> It looks like an error arising from a function buried deep in the code that has changed. Hopefully an easy solution.
> 
> Thanks,
> Ori 
> 
> Error: ValueError: Passing parameters norm and vmin/vmax simultaneously is not supported. Please pass vmin/vmax directly to the norm when creating it.


shanosborne:
Note, tests are failing because #530 hasn't been merged yet.

mperrin (approved):
👍

PR #540, opened 2022-03-19
Fix Failing CI

shanosborne:
The CI is failing for 2 different reasons:
- The Astropy development CI is failing because the way we were installing it in `tox.ini` is no longer allowed. So I updated the install command.
- The `test_get_zernike_coeffs_from_smif` test was failing in some builds because it calls NIRCam MIMF FP data locations, which were updated in the most recent version of pysiaf. So I updated the test to call a pinned version of pysiaf so this won't happen again if any NIRCam or MIRI numbers change in the future. (And I checked that this shouldn't be an issue for any other tests)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,7 @@ deps =
     legacy37: numpy==1.18.0
     legacy37: pysiaf==0.11.0
     legacy37: astropy==4.0.0
-    astropydev: git+git://github.com/astropy/astropy
+    astropydev: git+https://github.com/astropy/astropy
     poppydev: synphot
     latest: -rrequirements.txt
     stable: poppy
--- a/webbpsf/tests/test_opds.py
+++ b/webbpsf/tests/test_opds.py
@@ -13,6 +13,11 @@ import pytest
 import webbpsf
 import matplotlib.pyplot as plt
 
+# Set up a pinned pysiaf version so as not to break tests with any pysiaf value updates
+prd_data_dir = pysiaf.constants.JWST_PRD_DATA_ROOT.rsplit('PRD', 1)[0]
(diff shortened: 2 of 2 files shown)


shanosborne:
@mperrin do you have some time to review this PR?

mperrin:
@shanosborne can you explain a bit more about the pinning to a specific version of pysiaf? I suppose this is probably OK but I haven't fully thought it through yet. My one concern is that this makes the test not check that reasonable values are obtained in the default situation using the current pysiaf.  I see the issue with slight changes in the field point coordinates making the results slightly different... would it suffice to change the `rtol` value for that test instead? I'm happy to be convinced this is the most robust way to make the test - you've surely thought about it more than I have.

shanosborne:
From how I read the test (which could be wrong), it looks to me like it tests that the pysiaf numbers + a webbpsf function match Randal’s hardcoded values. I would expect that if we were to re-run Randal’s tool with the new telescope information, we would find that these hardcoded values would have changed and would still match up with the new pysiaf values. If that’s the case, I think it makes the most sense to update Randal’s values one time after commissioning is done and things have settled with the SIAF. And then in the meantime use the old pysiaf numbers to continue the comparison and preserve the test.

We could also change the tolerance, but I personally like to avoid widening the tolerance if it can be avoided, especially in cases where it’s already on the larger end to begin with (in this case, it’d be going to 1e-2) just because it can let other bugs pass through by accident. And so I’d think once the pysiaf numbers are done with the commissioning changes, we’d want to change the tolerance back up it it’s original value anyway. So in some ways, it’s kind of the same fix just a difference approach.

mperrin (approved):
👍

PR #549, opened 2022-05-25
Remove python 3.7 and add python 3.10 

shanosborne:
Remove support for Python 3.7, increase all CI tests by 1 Python version and now include Python 3.10

Also adding this here: updating the publish-to-pypi.yml file to work with new git security changes. This fix comes from  @zanecodes and was also implemented here https://github.com/spacetelescope/pysiaf/pull/240


shanosborne:
Tests that are currently failing are due to issues in poppy, and should pass once https://github.com/spacetelescope/poppy/pull/503 is merged

PR #556, opened 2022-06-02
Using measured OPDs from MAST 

mperrin:
Adds support for retrieving measured OPDs from MAST and using for webbpsf calculations. 

**Checklist:** 

- [x] Handle FITS formatting, resizing, header keywords, and other such book-keeping
- [x] Back out SI WFE at the sensing field point to estimate OTE-only WFE, and set that OPD into an instrument model for use in calculations.
- [x] Retrieve from MAST an OPD file specified by name
- [x] Retrieve from MAST an OPD file specified by datetime
- [x] Also back out the OTE field dependence term from the sensing field point, to avoid double counting that when webbpsf adds that term back in. 
- [x] Rework the SI WFE backout code to handle differences in which SI is used for sensing and for PSF calculation, e.g. right now you can't use an OPD measured in NIRCam into a webbpsf instance for NIRSpec. 
- [x] Implement other options for selecting which OPD to use, in particular a 'closest' option to use whichever of before/after is closer in time, and an 'interpolate' option to use some linear interpolated estimate. 

**Example:**

*Select by filename:*
```
nrc = webbpsf.NIRCam()
nrc.load_wss_opd('R2022052701-NRCA3_FP1-2.fits')
```

*Select by date:*
```
nrc.load_wss_opd_by_date('2022-06-01T12:00:00', choice='after')
```

Automatic downloads requires having a MAST API token stored in an environment variable. See https://auth.mast.stsci.edu/info


JarronL:
SI WFE extraction appears to fail when coronagraphic options are enabled. I imagine this is only a problem for NIRCam Lyot Stops because switching to coronagraphic mode selects a different reference table to determine the field-dependent WFE map.

JarronL:
I noticed that this works:
```
nrc.load_wss_opd_by_date('2022-06-01T12:00:00', choice='after')
iterating query, tdelta=3.0

MAST OPD query around UTC: 2022-06-01T12:00:00.000
                        MJD: 59731.5

OPD immediately preceding the given datetime:
	URI:	 mast:JWST/product/R2022060101-NRCA3_FP1-2.fits
	Date (MJD):	 59728.5991
	Delta time:	 -2.9009 days

OPD immediately following the given datetime:
	URI:	 mast:JWST/product/R2022060103-NRCA3_FP1-1.fits
	Date (MJD):	 59731.7095
	Delta time:	 0.2095 days
User requested choosing OPD after date 2022-06-01T12:00:00.000, which is R2022060103-NRCA3_FP1-1.fits
Importing and format-converting OPD from /Users/jarron/NIRCam/webbpsf-data/MAST_JWST_WSS_OPDs/R2022060103-NRCA3_FP1-1.fits
Backing out SI WFE from the WF sensing field point
```


 yet the following fails for some reason:
```
nrc.load_wss_opd_by_date('2022-06-02T12:00:00', choice='before')
iterating query, tdelta=3.0
iterating query, tdelta=6.0
iterating query, tdelta=12.0
iterating query, tdelta=24.0
Insufficient JWST OPD data can be found within 1 week of the specified date. This date is likely outside of the valid range
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-25-b3daf3bf328a> in <module>
----> 1 nrc.load_wss_opd_by_date('2022-06-02T12:00:00', choice='before')

~/GitHub/webbpsf/webbpsf/webbpsf_core.py in load_wss_opd_by_date(self, date, choice, verbose, **kwargs)
   1463 
   1464         """
-> 1465         opd_fn = webbpsf.mast_wss.get_opd_at_time(date, verbose=verbose, choice=choice)
   1466         self.load_wss_opd(opd_fn, verbose=verbose, **kwargs)
   1467 

~/GitHub/webbpsf/webbpsf/mast_wss.py in get_opd_at_time(date, choice, verbose)
    164         date = astropy.time.Time(date, format='isot')
    165 
--> 166     prev_opd_fn, post_opd_fn = mast_wss_opds_around_date_query(date, verbose=verbose)
    167 
    168     if choice== 'before':

~/GitHub/webbpsf/webbpsf/mast_wss.py in mast_wss_opds_around_date_query(date, verbose)
    104         if tdelta > 14:
    105             print("Insufficient JWST OPD data can be found within 1 week of the specified date. This date is likely outside of the valid range")
--> 106             raise RuntimeError("Cannot find OPDs in MAST before/after that date. Date is likely outside the range of valid data.")
    107         obs_table = mast_wss_date_query(date, tdelta)
    108 

RuntimeError: Cannot find OPDs in MAST before/after that date. Date is likely outside the range of valid data.

```

mperrin:
@JarronL please try the latest version of this, pushed just now, which should fix the issue that you saw when coronagraphic field points were selected. Now it creates an entirely separate instrument instance to retrieve the SI WFE model from.  This also handles the case where the sensing instrument is different from the current instrument model (e.g. using NRC WFS OPDs to update NRS calculations.)

JarronL:
Looks good! Maybe set `plot=False` by default?

mperrin:
Implemented the feature to find the closest OPD (either before or after, depending on which is closest). I made this the default now that it works.

mperrin:
@JarronL you can try again on the date queries that were giving you trouble before. I've made it smarter about handling edge cases (literally, when at the edge of the range of available dates)

mperrin:
Added in handling of OTE field dependence as well. Plots updated to show the effects of this. 

Note, if saving an OPD for use in other code (as requested by NIRSpec team), this now exports the total OTE WFE at the selected field point (i.e. what's saved out includes the OTE WFE as estimated at the selected field point, added on top of the inferred global WFE at the master chief ray.

mperrin:
In lieu of code review - @JarronL @juliengirard you both agree this code is working fine and ready to merge?

PR #561, opened 2022-07-15
Reordered installation recommendations

ojustino:
This is a continuation on the theme of updating the documentation. I moved AstroConda installations from most to least recommended and added a warning about expected compatibility issues with future versions of WebbPSF.


mperrin (approved):
👍

PR #563, opened 2022-07-29
output_path passed to mast_retrieve_opd() when default choice=='closest'

rcooper295:
Fix to minor issue where the output_path kwarg is not carried through to mast_retrieve_opd() from get_opd_at_time() when choice=='closest', which is the default. This causes a permission error when the user doesn't have write permissions to $WEBBPSF_PATH.


mperrin (approved):
👍 
Thanks @rcooper295

PR #578, opened 2022-09-09
add unit test for coron_shift

mperrin:
Adds a unit test that `coron_shift_{x/y}` and `source_offset_{x/y}` yield consistent results.


marcio-melendez (approved):
Looks good, it capture the agreement between both methods
image

PR #571, opened 2022-08-19
Optical model updates in flight continued: more WF measurement plotting

mperrin:
- Add wavefront trending plots and initial documentation for those. Merging in a lot of code that was developed in notebooks outside of webbpsf. 
- Misc updates to utility functions etc. More plot flexibility in gridded PSF library display. 
- Add docs page showing plots of wavefronts over time early in cycle 1

PR #582, opened 2022-09-16
enforce image origin = lower everywhere, for consistency

mperrin:
from @obi-wan76 in #571:

> @mperrin I just noticed that the pupil orientation from my test is different than yours, I am not sure where are you setting the origin of the figure, webbpsf.trending.show_opd_image? Your example has the entrance pupil orientation so I am not sure if this is a change that already exist and will get merge as part of other PR. This is just a very minor detail, mostly for consistency.


marcio-melendez (approved):
Yes, it looks good. Thanks!!
image

Issue #583, opened 2022-09-17
minor: docs rendering issue


mperrin:
Something changed in the CSS or other style display for the documentation, such that bullet point lists no longer show any bullet symbols. For instance see https://webbpsf.readthedocs.io/en/latest/relnotes.html 

This is just cosmetic but it does seem to make it harder to read.

ojustino:
I believe this is because of a problem with the `stsci_rtd_theme` package that's used to control CSS settings. `jwst`'s docs have [~the same issue~](https://jwst-pipeline.readthedocs.io/en/latest/jwst/introduction.html#output-file) ([updated link](https://jwst-pipeline.readthedocs.io/en/latest/jwst/user_documentation/input_output_file_conventions.html)).

I reported this issue at spacetelescope/stsci_rtd_theme#19 and proposed a fix at spacetelescope/stsci_rtd_theme#20, but I'm not sure what the turnaround time is for a review. A temporary solution is to manually replace each `*` with a bullet character (`\•`), but it's tedious.

ojustino:
The maintainers of `stsci_rtd_theme`accepted the fix in spacetelescope/stsci_rtd_theme#20 several months ago. They haven't done another release since the PR was merged, so we may need to specifically install the version on `main` in order for the `webbpsf` docs to render properly again.

ojustino:
Now that version 1.0.1 of `stsci_rtd_theme` is out, this issue should be moot. Bullets are functional again in the documentation.

PR #584, opened 2022-09-17
Release 1.1 prep

mperrin:
- Updates data files for release 1.1. In particular this includes setting the default OPD to an in-flight measured OPD which has been added to the data files tarball. 
- Updates CI to use the newer data files
- Release notes text (not yet final, may be missing some things. See also #583)
- Update/refresh the WebbPSF_tutorial.ipynb notebook, to fix some uses of stale deprecated syntax there.


marcio-melendez (approved):
Looks good to me. For this review:

1. I ran the improved WebbPSF_tutorial.ipynb in its entirely with no issues.  

2. Created a couple of examples to double check the correct OPD is loaded, locally (STSCI) at /grp/jwst/ote/webbpsf-data/JWST_OTE_OPD_cycle1_example_2022-07-30.fits

3. I reviewed the content on the default OPD file and I have a couple minor comments: 
a) inside the header there is a local path, "Input file is: /Users/mperrin/software/webbpsf-data/MAST_JWST_WSS_OPDs
HISTORY /O2022073001-NRCA3_FP1-1.fits ". This is not an issue at all but just wanted to mention this in case you want to hide local paths to your installations.  
b) "See FITS extension SENSING_OTE_FD_WFE for the WFE model used.    " I could not find this extension. The current default OPD has only one extension and the original OPD has many extensions but none with that name. 
hd.info()
Filename: /grp/jwst/ote/webbpsf-data/JWST_OTE_OPD_cycle1_example_2022-07-30.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU      89   (1024, 1024)   float64

mperrin:
Thanks @obi-wan76. I updated the header slightly and recreated the zip files. Purely cosmetic and informational, no functional change.  But thanks for pointing that out - this reminded me that I had meant to also include the target phase map as a file in the data release. I had made that file, but had forgotten to copy it to the data files source location. I did that now.

(The actual WSS target phase map is close to and consistent with the model WebbPSF already had, from the NIRCam SI internal WFE model, but the version the WSS uses has higher spatial frequencies, included more Zernikes from ISIM CV than the WebbPSF SI model did.  WebbPSF already has code in the load_wss_opd parts to use the exact WSS target phase map from this file, if present, for better consistency when computing OTE-only vs observatory WFE values)

FYI, those extension names referred to in the header comments are made internally inside `load_wss_opd`, but I didn't save them all when exporting the file to disk for inclusion in webbpsf data.

ojustino (approved):
We will resolve the code coverage issues post-release. For now, on to 1.1.0.

PR #590, opened 2022-09-21
Adjusted install/release instructions post-conda

ojustino:
On conda, WebbPSF is only available through the AstroConda channel, which is only compatible with Python <=3.7. Since WebbPSF v1.1.0 will raise the Python requirement to 3.8, `conda install` can't work. I adjusted the installation and release documentation pages to reflect the change.


marcio-melendez (approved):
Looks good.

PR #593, opened 2022-09-23
Changed date in release notes

ojustino:
Today should be the day now that it looks like we've figured out some procedural issues.


mperrin (approved):
Today’s the day :-)

ojustino:
`codecov` is somehow registering a decrease in coverage here... I will assume it's spurious and move on.

PR #595, opened 2022-10-08
bug fix for include_ote_field_dependence setting

mperrin:
The JWST instrument classes have an attribute `.include_ote_field_dependence` which is supposed to let you toggle off the OTE field dependence model. (Generally you don't want to do this but it can be useful in some contexts). That setting didn't actually work. This one-line fix makes it work. 

**example test code:**

```
for test_val in [True, False]:

    nrc.include_ote_field_dependence = test_val

    nrc.detector_position=(1024, 1024)
    opd1 = nrc.get_wfe('ote')

    nrc.detector_position=(0, 512)
    opd2 = nrc.get_wfe('ote')

    plt.figure()
    plt.imshow(opd1-opd2)
    plt.title(f"Delta OPD between 2 positions\nwith include field dep = {nrc.include_ote_field_dependence}",
             fontweight='bold')

    print(f"With OTE Field Dep={nrc.include_ote_field_dependence}, do OPDs vary with position? "
          f"{not np.allclose(opd1, opd2)}")
```

**without this fix, regardless of the setting, the OPD still changes with detector position:**

Screen Shot 2022-10-08 at 5 37 19 PM

**With this fix, the setting lets you turn off the field dependence:**

Screen Shot 2022-10-08 at 5 37 52 PM

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -920,7 +920,8 @@ class JWInstrument(SpaceTelescopeInstrument):
                 transmission=pupil_transmission,
                 opd=opd_map,
                 opd_index=opd_index,
-                v2v3=self._tel_coords(), npix=npix
+                v2v3=self._tel_coords(), npix=npix,
+                include_nominal_field_dependence=self.include_ote_field_dependence
             )
 
         return pupil_optic


mperrin:
(not urgent for over the weekend, but I wanted to put this fix in while I'm thinking about this. Discovered and fixed while looking at PSFs in some science data :-) )

JarronL:
Also just discovered this while comparing empirical PSFs with WebbPSF models. The recommended fix looks to work on my end. Maybe good to merge the commit?

mperrin:
Oh yes. Thanks for the reminder. This must have slipped through the cracks for @obi-wan76 to review. I'm going to merge this now. 

Separately @JarronL we should talk about comparisons with empirical PSFs - I have been working on that as well (as time permits, in between a zillion other things), have found a few interesting things, and a few to fix and improve on. We should compare notes!

PR #598, opened 2022-10-20
596 - full_path parameter added to trending.wavefront_time_series_plot

kulpster85:


webbpsf/trending.py
```diff
@@ -56,12 +60,18 @@ def wavefront_time_series_plot(opdtable, start_date=None, end_date=None, label_v
...
+
     dates = astropy.time.Time(opdtable['date'], format='isot')
     pre_or_post = []
 
     for row in opdtable:
-        full_file_path = os.path.join(webbpsf.utils.get_webbpsf_data_path(), 'MAST_JWST_WSS_OPDs', row['fileName'])
-        rmses.append(fits.getheader(full_file_path, ext=1)['RMS_WFE'])
+        if full_path is False:
+            full_file_path = os.path.join(webbpsf.utils.get_webbpsf_data_path(), 'MAST_JWST_WSS_OPDs', row['fileName'])
+            rmses.append(fits.getheader(full_file_path, ext=1)['RMS_WFE'])
```
mperrin, line 73: Did you mean to have the `rmses.append` line inside the `if` statement here? I don't see how this works if `full_path` is set to true?

webbpsf/trending.py
```diff
@@ -45,6 +45,10 @@ def wavefront_time_series_plot(opdtable, start_date=None, end_date=None, label_v
         Start and end dates for the plot time range. Default is March 2022 to present.
     label_visits : bool
         Label program_id:visit_id for each WFS visit.
+    label_events : bool
+        Label events along the x-axis in time defined in the events {}
+    full_path : bool
+        Whether the fileName column already contains a full file path or not
```
mperrin: Question. We might be able to avoid needing this parameter by making the code simply check if the path in the OPD table is a full path, via `os.path.isfile()`. If the filename provided is a valid full path to an existing file, use it as-is. If not, try appending the webbpsf data path on the front end, etc, to make an absolute path.

webbpsf/trending.py
```diff
@@ -177,15 +203,21 @@ def wfe_histogram_plot(opdtable, start_date=None, end_date=None, thresh=None):
...
-    webbpsf.mast_wss.download_all_opds(opdtable1)
+    if download_opds:
+        webbpsf.mast_wss.download_all_opds(opdtable1)
 
     # Retrieve all RMSes, from the FITS headers. 
     # These are observatory WFE (OTE + NIRCam), at the WFS sensing field point
     rmses=[]
+
+    if download_opds is False:
+        rmses = opdtable['rms_wfe']
```
mperrin: This seems potentially confusing, since now the download_opds switch is controlling something completely unrelated to downloading. Maybe this should be a separate switch? Or a check on the contents of the table to see if the `rms_wfe` column exists, or something like that?

mperrin (commented):
Hi @kulpster85 - thanks for the PR. A couple questions...

webbpsf/trending.py
kulpster85: Yeah I can clean this up to check if the column already exists in the opdtable.  TAS has the ability to provide the RMS WFE already in the generation of the table to this method.  But if grabbing from MAST then WebbPSF would open each FITS file to extract the value.

webbpsf/trending.py
kulpster85: I like that I will switch to that implementation as I agree its much cleaner.

webbpsf/trending.py
kulpster85: Yeah I can change the condition to see if the column already exists in the opdtable.

webbpsf/trending.py
```diff
@@ -56,12 +58,19 @@ def wavefront_time_series_plot(opdtable, start_date=None, end_date=None, label_v
...
 
     for row in opdtable:
-        full_file_path = os.path.join(webbpsf.utils.get_webbpsf_data_path(), 'MAST_JWST_WSS_OPDs', row['fileName'])
-        rmses.append(fits.getheader(full_file_path, ext=1)['RMS_WFE'])
+        if os.path.isfile(row['fileName']) is False:
+            full_file_path = os.path.join(webbpsf.utils.get_webbpsf_data_path(), 'MAST_JWST_WSS_OPDs', row['fileName'])
+        else:
+            full_file_path = row['fileName']
+        if len(rmses) == 0:
+            rmses.append(fits.getheader(full_file_path, ext=1)['RMS_WFE'])
```
mperrin: I think this part may be problematic in the `len(rmses) == 0` check. Won't that make it just append the calculated RMS for the first row only, and then after that the length isn't 0, so it won't append for any of the rest of the lines in the table. 

I suggest instead `if 'rms_wfe' not in opdtable.colnames` so it's consistent with  but opposite from the check above, and like the other version of this check in line 219.

mperrin (changes requested):

webbpsf/trending.py
kulpster85: Agreed I will make this change.  Thanks for the suggestion.

mperrin (approved):
👍

PR #600, opened 2022-10-24
597 - add opdtable as positional param to monthly_trending_plot

kulpster85:


marcio-melendez (commented):
This looks good. Just adding the option for the user to provide the OPD table. However, how do you check that the table provide by the user has the same fields as those created by get_opdtable_for_month? e.g.,i'is_pre_correction'  , 'is_post_correction' etc. which are fields that  the monthly trending plot function and get_corrections are expecting to find. Maybe a check that all the keys in this OPD table are the same as the expectations.

kulpster85:
> 

I can add that logic to the changes - that's a good suggestion for robustness.

kulpster85:
> This looks good. Just adding the option for the user to provide the OPD table. However, how do you check that the table provide by the user has the same fields as those created by get_opdtable_for_month? e.g.,i'is_pre_correction' , 'is_post_correction' etc. which are fields that the monthly trending plot function and get_corrections are expecting to find. Maybe a check that all the keys in this OPD table are the same as the expectations.

I have pushed some commits to address your comments @obi-wan76

mperrin:
@obi-wan76 can you re-review or approve this after the changes by @kulpster85 in response to your prior review?  (Like a month ago... I think this is ready to merge but will defer to you to re-review)

mperrin:
This is an old PR, from November, which it looks like @kulpster85 is still waiting for a review from @obi-wan76 for. Trey is this still a relevant/desired PR? If so, reminder to Marcio. :-)

kulpster85:
This would be an efficiency improvement that TAS could leverage to provide the opdtable as an input rather than download the OPDs from MAST.

marcio-melendez (approved):
I think at that time I reviewed this new changes but never submitted/approve

mperrin:
@obi-wan76 FYI looks like you approved this back in March but never merged it...

PR #601, opened 2022-10-25
Phase decomposition updates

mperrin:
(Some code I wrote back in the crazy blur of this past spring, but never PRed at the time...) 

- Refactor `SUR` class to allow creating an empty SUR in Python, adding values, and then writing it out to disk. (Before this, the code was able to read in SURs but you couldn't just create one. While we don't expect or intend to create any SURs for flight usage this way, in some simulation cases this can be useful).
- Update `move_sur` function to allow passing in a SUR object, as well as a filename on disk for a SUR file. 
- Add a function `get_rms_per_segment` which computes the RMS WFE per each segment.

### DIFF
--- a/webbpsf/opds.py
+++ b/webbpsf/opds.py
@@ -2195,8 +2195,8 @@ class OTE_Linear_Model_WSS(OPD):
 
         Parameters
         ----------
-        sur_file : file name
-            Path to SUR XML file
+        sur_file : file name, or SUR object instance
+            Path to SUR XML file, or a webbpsf.surs.SUR object
         group : one-based int index
             Index to a single group to run. Default is to run all groups. Note,
             this index counts up from 1 (not 0) for consistency with group indexing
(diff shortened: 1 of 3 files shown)


marcio-melendez:
I have been using the RMS function already in OPDS to calculate the RMS per segment, this new function seems to be doing the same thing or is there something different?
https://github.com/spacetelescope/webbpsf/blob/develop/webbpsf/opds.py#L545
I guess this one is easier to use just with the OPD array because for the other one I have to define an OPD object but it is the same. 
Just a comment, I think is good to have a similar function that's easier to use.

mperrin:
@obi-wan76 Yes I think this ought to yield consistent result with the one in OPDS... But checking, the two functions yield slightly different values.  One is using the RMS but the other is using the std dev (These are only identical if the mean is zero, but slightly different if not). I'll fix this so they're consistent!

mperrin:
@obi-wan76 would you be willing to review this please? Should be good to go now.

marcio-melendez (approved):
This looks good.

PR #622, opened 2022-12-08
Release prep for 1.1.1

mperrin:


mperrin:
@obi-wan76 see here for the PR you can add the updated data URL for 1.1.1 to

marcio-melendez:
I added an update to the documentation to include the new link and the corresponding name of the tar.gz file.

PR #626, opened 2022-12-16
Update to read SI pixelscales directly from pysiaf

mperrin:
A small improvement in precision: use the measured pixel scale values from on orbit when computing PSFs. This should (slightly) improve the computed PSFs.

- Read pixel scales directly from pysiaf for each instrument and detector.  Removes having "default" pixel scales of each SI hardcoded in the Python; instead it reads from the SIAF values during init.
- Update pixel scale when you switch to a different SIAF aperture (in particular, if you switch NIRCam detectors it will change the aperture to the one for that detector, which will update the pixel scale). 
- Update unit tests correspondingly to still work with the above. 

Note, this still retains the idealized assumptions of (a) square pixels, same scale in X and Y, and (b) all of NRC, FGS, NIS detectors are precisely unrotated with respect to the OTE V frame. The non-ideal departures from that, as recorded in pysiaf, continue to be handled in the rotation step and distortion model used to make the OVERDIST and DET_DIST extensions 3 and 4.

This also retains the ability to override the pixel scale with a custom value. If the user does this, then that custom value will be retained afterwards even if you switch apertures (i.e. setting a custom pixelscale disables the behavior for auto updating the pixelscale).

### DIFF
--- a/webbpsf/tests/test_nircam.py
+++ b/webbpsf/tests/test_nircam.py
@@ -192,20 +192,26 @@ def test_nircam_get_detector():
 
 
 def test_nircam_auto_pixelscale():
+    # This test now uses approximate equality in all the checks, to accomodate the fact that
+    # NIRCam pixel scales are drawn directly from SIAF for the aperture, and thus vary for each detector/
+    #
+    # 1.5% variance accomodates the differences between the various NRC detectors in each channel
+    close_enough = lambda a, b: np.isclose(a, b, rtol=0.015)
+
     nc = webbpsf_core.NIRCam()
 
     nc.filter='F200W'
-    assert nc.pixelscale == nc._pixelscale_short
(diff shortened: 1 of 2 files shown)


mperrin:
@obi-wan76 FYI this is a PR that I submitted in December, a simple bite-size standalone minor improvement.

mperrin:
Huh, this was working in December. Something seems to have broken in the rebase onto current develop. I'll investigate.

mperrin:
Updated and improved. This now works as intended to automatically read pixelscales from SIAF, per each aperture, but also to retain the ability for a user to set a custom pixelscale (and have that custom pixelscale not be overwritten by SIAF values).

This is now ready for review @obi-wan76 

Example:

Screen Shot 2023-06-01 at 2 55 49 PM

marcio-melendez (approved):
Looks good, I was able to run the example code and some other apertures.

PR #634, opened 2023-01-06
Fixed trending histogram binning so that bars add up to 1.0

Skyhawk172:
The histogram displayed by `trending.wfe_histogram_plot` was in fact a probability density function and, as a result, the sum of all the histogram bars was not 1.0 since the bin widths were not unity. This behavior is described in the [Matplolib documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html). This PR enforces a bin width of 1 nm, which now allows the user to interpret each histogram bar as the fraction of measurements at a given WFE.  

Also includes other minor display convenience fixes.

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -182,7 +182,7 @@ def wfe_histogram_plot(opdtable, start_date=None, end_date=None, thresh=None, do
     ----------
     opdtable : astropy.table.Table
         OPD table, retrieved from MAST. See functons in mast_wss.py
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Looks good!

marcio-melendez:
FYI. Just for the sake of tracking. 
 @BradleySappington this PR from Charles has been reviewed by me and it is ready to merge. 
Thanks!

Issue #627, opened 2022-12-22
Update SI filter throughput files from Pandeia 2.0


mperrin:
Use Pandeia 2.0 to update the data files for the SI throughputs to reflect on-orbit measured values.

PR #640, opened 2023-02-21
update SI throughputs using Pandeia 2.0

mperrin:
Addresses #627 . Updates SI throughput curves to be based on the Pandeia 2.0 version of the ETC data, consistent with in-flight performance.  Minimal/negligible changes for the most part, but good to have for consistency. Since we normalize the throughputs to sum=1.0 in WebbPSF calcs, the gain in total throughput compared to preflight predictions doesn't make much difference in practice to the PSFs. 

(This is work I had done some time ago, just not gotten around to committing and pushing)

- Re-ran `dev_utils/Get Throughputs from Pandeia.ipynb` to get SI throughputs from Pandeia 2.0
- Output files are in attached zip file. These will need to be placed in the source directory on central store, for each instrument, prior to the next creation of updated data file zip. 

[updated_throughputs.zip](https://github.com/spacetelescope/webbpsf/files/10797964/updated_throughputs.zip)


marcio-melendez (commented):
I am having problems running the notebook as is. I defined my reference file paths etc, but it fails quickly on `inst = pandeia.engine.jwst.NIRISS(mode='imaging')` because:
`~/anaconda3/envs/py38/lib/python3.8/posixpath.py in join(a, *p)
     74     will be discarded.  An empty last part will result in a path that
     75     ends with a separator."""
---> 76     a = os.fspath(a)
     77     sep = _get_sep(a)
     78     path = a

TypeError: expected str, bytes or os.PathLike object, not NoneType`
This seems an indication that my data path is incorrect. However, I downloaded the pandeai latest files and set up  the correct paths etc. 
In any case, if the intend of this PR is just to add the data file then it is Ok. Howver, if it is that the notebook works for everyone then I would like to get my notebook running and then add some comments that may help other people to run it too. If the latter, I need to troubleshoot my notebook.

mperrin:
I would not say the intent is "working for _everyone_", but should certainly be working for other members of the WebbPSF dev team such as you :-)

That error is definitely some Pandeia or Python environment thing, not anything in webbpsf code. Just to check you also updated the Pandeia package code to some recent version? You can get 2.0 from PyPI https://pypi.org/project/pandeia.engine/#history or can install more recent from Github of course

mperrin:
To be clear this is NOT a notebook that anybody outside of the WebbPSF dev team should ever need to run. Everybody else just gets the relevant files from installing the data zip file, as usual.  

That's true for everything in the `dev_utils` directory: we include them in the git repo, and they're available if for some reason anyone wants to look at them or even run them, but the expectation is only people on the development team will need to make use of anything in that folder.

marcio-melendez (approved):
Yes, that's what I figured. I have version 2.0 from PyPI and I have other paths as intended, which I can confirm directly from the notebook. In any case, I need to troubleshoot or open a help desk ticket :). I can approve  the new file.

mperrin:
Just to check, did you update the files in central store? Inside the webbpsf_data_source folder I mean. You can use the files from the zip attached to this PR since this notebook isn't yet working for you

marcio-melendez:
No, I have not update the files yet. Yes, I'll update with the zip file attach.

mperrin:
@obi-wan76 I wanted to follow up on your message (above) from March, in which you said you'd update the files on central store with the updated throughputs file. Did that get done?

marcio-melendez:
I just did it. Note that the NIRCam and MIRI folder have some _thru.fits files. Should we remove the _thru.fits and just have the throughput.fits? or keep both?

mperrin:
> Note that the NIRCam and MIRI folder have some _thru.fits files. Should we remove the _thru.fits and just have the throughput.fits? or keep both?

Yes, any _thru.fits are super old and should be removed

BradleySappington:
@mperrin, I'm unable to re-run the jobs or view the logs. Do you have the ability to re-run the actions?  I'd like to investigate the failed test.

mperrin:
huh, you should be able to click through to get details and logs from the CI. That doesn't work for you?

Ah I see, for me too it's showing "The logs from this run have expired and are no longer available". Fine, they don't keep them forever.  We could trigger this to re-run the tests now... 

But note that this PR _doesn't actually change anything_ in the code base of the package - it changes a notebook in the dev_utils folder.  The actual important changes are the updated data files, which happened on central store and the zip file in Box, but isn't actually in this PR. So the passing of the tests in this branch or not has no real significance.

BradleySappington:
> huh, you should be able to click through to get details and logs from the CI. That doesn't work for you?
> 
> Ah I see, for me too it's showing "The logs from this run have expired and are no longer available". Fine, they don't keep them forever. We could trigger this to re-run the tests now...
> 
> But note that this PR _doesn't actually change anything_ in the code base of the package - it changes a notebook in the dev_utils folder. The actual important changes are the updated data files, which happened on central store and the zip file in Box, but isn't actually in this PR. So the passing of the tests in this branch or not has no real significance.

Yeah, I wanted to re-run the jobs to see what was causing the fail (however there is no `re-run jobs` available for me).  
If you can initiate the re-run I'd be happy to see if I can figure out a way to make this pass.  Or, we can try to override the CI checks

mperrin:
No re-run jobs available for me either, I guess because this was from so long ago. 

I rebased the PR onto current develop and pushed that, just to have a new commit that would trigger the CI. 


My guess is the failure of the prior test was some transient failure, or some completely unrelated already-fixed thing, and I predict all will run fine now...

BradleySappington:
@mperrin, that was my assumption as well, and looks like its all good

PR #642, opened 2023-02-28
Update/enhance trending plot to show WSS proposed corrections

mperrin:
Updates the function `single_measurement_trending_plot` to show the proposed correction move for a given OPD from the WAS (retrieved from the FITS extensions in the OPD file), as opposed to previously which showed the PTT decomposition using WebbPSF. These are similar but not exactly the same


mperrin:
Example of the updated output plot. The only panel that changed is outlined in green. Third in bottom row: 

Unknown

marcio-melendez (approved):
This looks really good, great addition.

mperrin:
This was approved in March, but not merged then due to a failing CI. That was subsequently fixed in #649, so this could have been merged then. All tests do pass now. Merging now.

PR #650, opened 2023-03-22
add phase retrieval crosscheck plot and wfs obs delta plot

mperrin:
Some more code that I had written a while ago, but never yet PRed in here. 

This adds 2 functions which display the WFS weak lens data, from the cutouts included in the OPD files in MAST. 

`plot_phase_retrieval_crosscheck` shows the measured weak lens data, and the WAS' model from the PR. I.e. similar to what the PRMS folks would display and blink back and forth to visualize the quality of the PR (though the code here doesn't blink them, it just shows the images side by side.)

`plot_wfs_obs_delta` plots the difference between two weak lens observations, for instance two back-to-back sensing obs. 

Both of these can be used for instance to check if there's a binary companion or some other data quality issue. 

The PR also includes brief examples of using each, added to the trending notebook in the docs. 

Unknown-18

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -1117,3 +1117,187 @@ def monthly_trending_plot(year, month, verbose=True, instrument='NIRCam', filter
                                        names=['Date', 'Filename', 'WFS Type', 'RMS WFE (OTE+SI)', 'RMS WFE (OTE only)',
                                               'EE(2.5 pix)', 'EE(10pix)'])
     return result_table
+
+
+def plot_phase_retrieval_crosscheck(fn, vmax_fraction=1.0):
+    """ Make a plot which displays the quality of the phase retrievals,
+    comparing the WSS-generated retrieved PSF with the actual measured PSF,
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
This looks good, I was able to test both functions as expected, very useful. I noticed that to run, it requires skimage which is not in the list of require packages for webbPSF, should it be? 

image

image

mperrin:
I didn't make scikit-image a required dependency because this is an extra/optional plot, not a core part of the functionality. Most users won't use this I think. And my instinct is to try to minimize dependencies when possible. But if you want to make it required then that's fine with me

mperrin:
This was approved in March, but not merged then, maybe due to some unrelated CI issue(s). I rebased onto current develop and will see whether the CI runs cleanly.

mperrin:
CI all passes. Codecov complains, because this plotting function isn't tested in the CI, but that's consistent with most of the rest of the display functions that aren't tested. Therefore this is good to go.

PR #665, opened 2023-05-01
Finish cleaning up setup*

BradleySappington:
Remove setup* and put flake8 in tox.ini


tox.ini
```diff
@@ -5,6 +5,8 @@ envlist =
     py39-legacy-test
     py{39,310,311}-{poppydev,pysiafdev}-cov
 
+isolated_build = true
```
zacharyburnett, line 8:
```suggestion
```
this is no longer read by Tox 4: https://tox.wiki/en/latest/upgrading.html#isolated-environment-on-by-default

tox.ini
```diff
@@ -58,3 +60,7 @@ deps =
     pycodestyle
 commands =
     pycodestyle webbpsf
+
+[testenv:flake8]
+exclude= .git,__pycache__, test*
+max-line-length = 120
```
zacharyburnett, line 66: I would recommend moving this to a `.flake8` file (INI format), as putting it here might be a "back of the fridge" scenario (forgotten about)

zacharyburnett (approved):
LGTM, minor trivial suggestions

tox.ini
BradleySappington: I was getting errors after removing setup.py and this was the only thing that fixed it

tox.ini
BradleySappington: I may leave here just because I suspect a "ruff" replacement will be happening eventually

BradleySappington:
@ojustino These changes mirror that of Poppy, a review would be appreciated here as well

ojustino (approved):

PR #668, opened 2023-05-11
Implement distortion for Roman

Skyhawk172:
Implement geometric distortion for Roman using the Roman SIAF.  It appears we can use the same infrastructure already set up for JWST can be used for Roman as well.


Skyhawk172:
Current implementation will require installing Python package `soc_roman_tools`, which is where the Roman SIAF is stored.

Also will need to remove the hardcoded path to local WFI Zernike file.

Skyhawk172:
I've done a bunch of tests and I think this PR is good to go. Feedback appreciated. 

One thing I am still unsure of is whether the rotation needs to be applied.

Here are some plots to compare the effect(s) of including distortion on the Roman PSF:

image

image

image

Skyhawk172:
Regarding rotation, I believe that there is no need to apply it, like it is done for JWST NIRCam, NIRISS, and FGS. From email thread with GSFC Optical engineer back in 2020, the "Pupil FITS is FPA coordinate and no need of rotation/flip", which suggests we can use the pupil FITS files as they are for the FFT/MFT propagation.

Skyhawk172:
I also think there is no need to apply a rotation, like what is done for JWST NIRCam, NIRISS, or FGS. (Re-)Reading an email thread with a GSFC optical engineer from 2020, it's mentioned that "Pupil FITS is FPA coordinate and no need of rotation/flip", which suggests the pupil FITS files are already aligned with the detector frame.

Moreover, it also appears the Zernikes as defined in the reference data (see "Zernike NOLL def" tab in the spreadsheet) are properly aligned with the _exit_ (i.e. flipped) pupil, so I don't think we need to do any additional flip/rotation. Confirmation of this would be appreciated.

marcio-melendez (commented):
Would you need to add soc_roman_tools as a dependency?

Skyhawk172:
Yes, I think we'll have to add it to the dependencies.

marcio-melendez:
@BradleySappington is this as simple as adding this into the toml file? (soc_roman_tools dependency)

BradleySappington:
> @BradleySappington is this as simple as adding this into the toml file? (soc_roman_tools dependency)

Yes, add it with any preferred versioning to the dependencies section of pyproject.toml

Skyhawk172:
Okay, I can add it to the PR: `"soc_roman_tools>=0.1.0",`, where v0.1.0 is the current version on pypi.

marcio-melendez (approved):
This looks good.

Skyhawk172:
Other than the approximate V2/V3 location of the Roman Coronagraph, there is not much information available in the SIAF (that information simply does not exist yet), so I had to enforce no distortion for the Roman CGI.

Skyhawk172:
@obi-wan76 @BradleySappington I think this is one is good to go. If you agree, we can fold it in the next release.

PR #669, opened 2023-05-12
non-standard pixel sizes for distortion

JarronL:
Previously, distortion code assumed that webbpsf detector pixels and pysiaf apertures had similar pixel scales. However, it is possible to specify any non-standard pixel size (e.g., `inst.pixelscale = whatever`) when generating a PSF. This fix correctly applies the pysiaf distortion model for any arbitrary pixel size.


webbpsf/distortion.py
```diff
@@ -63,7 +63,7 @@ def distort_image(hdulist_or_filename, ext=0, to_frame='sci', fill_value=0,
         Value used to fill in any blank space by the skewed PSF. Default = 0.
         If set to None, values outside the domain are extrapolated.
     to_frame : str
-        Type of input coordinates. 
+        Requested of output coordinate frame.
```
mperrin: Minor: Clarify/rephrase. "Requested of" doesn't make sense...

webbpsf/distortion.py
```diff
@@ -116,29 +116,30 @@ def distort_image(hdulist_or_filename, ext=0, to_frame='sci', fill_value=0,
...
-
     # ###############################################
     # Create an array of indices (in pixels) that the final data will be interpolated onto
-    xnew_cen, ynew_cen = aper.convert(xsci_cen, ysci_cen, 'sci', to_frame)
+    if to_frame=='sci':
+        xnew_cen, ynew_cen = (xsci_cen, ysci_cen)
+    else:
+        # Choose conversion function
+        func_convert = getattr(aper, f'sci_to_{to_frame}')
+        xnew_cen, ynew_cen = func_convert(xsci_cen, ysci_cen)
```
mperrin: Can you explain the motivation behind this change?  It appears to me that this code is just replicating what the `aper.convert` function already does: check for a null transformation and if so return the input x,y; otherwise look up the correct X_to_Y function and then call it.

webbpsf/distortion.py
```diff
@@ -171,7 +179,12 @@ def distort_image(hdulist_or_filename, ext=0, to_frame='sci', fill_value=0,
         ynew += ynew_cen - np.median(ynew)
     
     # Convert requested coordinates to 'idl' coordinates
-    xnew_idl, ynew_idl = aper.convert(xnew, ynew, to_frame, 'idl')
+    if to_frame=='idl':
+        xnew_idl, ynew_idl = (xnew, ynew)
+    else:
+        # Choose conversion function
+        func_convert = getattr(aper, f'{to_frame}_to_idl')
+        xnew_idl, ynew_idl = func_convert(xnew, ynew)
```
mperrin: Same comment as above; this appears redundant with what `aper.convert` already does?

mperrin:
I'm trying to follow the logic behind this change - can you give me an example for some calculation or setup that doesn't work currently but does work properly with this change?

mperrin (commented):
Couple questions & comments after a first quick skim.

webbpsf/distortion.py
JarronL: Ah, indeed. I had looked at the pysiaf `convert` function a long time ago and (perhaps erroneously) recall it making additional unnecessary transformations, so have always written my own set of if statements to minimize the number of calls. But looks like the logic now is much improved.

webbpsf/distortion.py
JarronL: Right. Accidentally removed "type"

webbpsf/distortion.py
JarronL: Reverting to `aper.convert`.

JarronL:
> I'm trying to follow the logic behind this change - can you give me an example for some calculation or setup that doesn't work currently but does work properly with this change?

A colleague wanted to generate a MIRI PSF with a non-standard pixel size in order to match to the NIRCam PSF. So, by setting `miri.pixelscale=0.061` (for instance), the distortion calculation would not correctly generate the new coordinates to perform the regular grid transformation onto. One symptom was that the total sum of a distorted PSF would be something like 30% of the sum from an undistorted PSF (ie., undistorted PSF is close to 1.0, while distorted PSF sum was 0.3). 

I realized that when I originally authored this code, I made an assumptions about pixels scales and the meaning of `oversample` that were only applicable when the PSF pixel scale equals the SIAF pixel scale. 

The updated code now generalizes the pixel scales such that I can still use SIAF to perform various coordinate transformations, but also making sure that I'm preserving the intended pixel scale when calculating the new set of coordinates.

webbpsf/distortion.py
```diff
@@ -147,9 +143,16 @@ def distort_image(hdulist_or_filename, ext=0, to_frame='sci', fill_value=0,
             assert xnew_coords.shape==ynew_coords.shape, "If new x and y inputs are a grid, must be same shapes"
             xnew, ynew = xnew_coords, ynew_coords
     elif to_frame=='sci':
-        xnew = xarr / oversamp + xnew_cen
-        ynew = yarr / oversamp + ynew_cen
+        osamp_x = aper.XSciScale / pixelscale
+        osamp_y = aper.YSciScale / pixelscale
+        osamp = (osamp_x + osamp_y) / 2
+        xnew = xarr / osamp + xnew_cen
+        ynew = yarr / osamp + ynew_cen
```
mperrin: Can you explain the logic here for averaging `osamp`? This computes first a distinct oversampling factor in each of X and Y, but then computes `xnew` and `ynew` using the average oversampling factor. Wouldn't it be more precise to calculate using the distinct oversampling factors per axis, like `xnew = xarr / osamp_x + xnew_cen`?

mperrin (approved):
I can replicate the bug you describe, and confirm this fixes it. I added a unit test that demonstrates/verifies that fix.  

I do have the one question about oversampling per axis versus averaged; I'm honestly not sure which is the more precise/rigorous way to treat that, and would need to think through the math to double check. Apart from that clarification, this is ready to go.

webbpsf/distortion.py
JarronL: I originally tried doing it per-axis, but `test_apply_distortion_pixel_scale()` would fail. I did not have time to look into why the test was failing, as I didn't quite understand the logic behind the test. Reverting to a single oversampling applied across both axes fixed the issue. Need to investigate further.

mperrin: Huh, yeah, now that I look at that test... I do not understand the logic either. 🤣

mperrin:
@JarronL FYI I tried to look at the test  `test_apply_distortion_pixel_scale` and I can't really figure out the logic of that test and how it is supposed to work. It certainly is not doing what the docstring comment says it should do. 

I am inclined to say just to delete that test, and you can edit this PR back to what you tried first, to treat the X and Y scales separately/ 

But really I should probably hand this off to @obi-wan76 to look at and think about, or to @Skyhawk172 who has been looking at the distortion code in the context of Roman.

Skyhawk172:
I don't understand `test_apply_distortion_pixel_scale` either. First, the test doesn't seem to do exactly what the docstring states; second, I'm not sure the "linear skew" assumption is perfectly valid.

With that said, I'd have to think hard about another more intuitive test... So I'm also inclined to reinstate Jarron's per-axis approach at the expense of removing this test (at least for the time being).

marcio-melendez:
@JarronL what do you think? do you want to go back to doing it per-axis and removing `test_apply_distortion_pixel_scale()` ? Unless you have found the reason why the test is failing.

JarronL:
Yes, I'm onboard with removing `test_apply_distortion_pixel_scale()` and reverting to per-axis:

``` python
        osamp_x = aper.XSciScale / pixelscale
        osamp_y = aper.YSciScale / pixelscale
        xnew = xarr / osamp_x + xnew_cen
        ynew = yarr / osamp_y + ynew_cen
```

JarronL:
Would you like me to perform the change in this branch?

marcio-melendez:
Yes, that would be great! Thanks!

PR #671, opened 2023-05-19
Add H2RG detector effects sim framework

mperrin:
Work in progress toward implementing better models for detector systematics affecting PSFs. 

- Adds `detectors.py` containing `apply_h2rg_charge_diffusion()`, which applies a simple ad hoc Gaussian convolution to stand in for H2RG detector charge diffusion (we should make this model better) 
- For NIRCam, NIRISS, FGS call that function from within `JWInstrument. _calc_psf_format_output`. 
- Minor, cleanup: remove `SpaceTelescopeInstrument._calc_psf_format_output` which actually did nothing and just called the superclass function. (This has no functional change, I just happened to see this opportunity to simplify old unused code while working on this PR). 

To Do: 

- [x] Test and make better the model for H2RG charge diffusion. Test this for multiple datasets. 
- [x] Implement detector IPC model as an additional smoothing kernel
- [x] Consider moving the `apply_miri_scattering` code from `distortion.py` to `detectors.py`?


mperrin:
@obi-wan76 This is ready for you to take a look at and test, in comparison with the TA ePSFs for instance. 

And there is a placeholder function `apply_detector_ipc` which is blank for now but gives you a place to put the IPC models that you have been working on.

webbpsf/detectors.py
```diff
@@ -0,0 +1,40 @@
...
+
+    sigma = options.get('charge_diffusion_sigma')
+
+    if sigma is None:
+        # look up default from constants
+        inst = psf_hdulist[0].header['INSTRUME'].upper()
+        key = f"NIRCAM_{psf_hdulist[0].header['CHANNEL'][0]}W" if inst=='NIRCAM' else inst
+        sigma = webbpsf.constants.INSTRUMENT_DETECTOR_CHARGE_DIFFUSION_DEFAULT_PARAMETERS[key]
+
+    ext = 1  # Apply to the 'OVERDIST' extension
```
marcio-melendez, line 231: The charge diffusion information is showing in extension OVERSAMP  and  DET_SAMP (0,1).  In the attached example, note that the code is not updating the charge diffusion via `options['charge_diffusion_sigm']`
I think  `sigma = options.get('charge_diffusion_sigma')` needs to include the instrument so it get the sigma value from it. In other words, I get the same PSF independent of the charge_diffusion_sigm value
Screen Shot 2023-06-13 at 8 13 46 PM

marcio-melendez:
I just added IPC effects to this PR. Below a comparison for NIRCam, NIRISS and MIRI with and without IPC effect. Note that the IPC effect can be turn off via `option['add_ipc'] = False` (default is True). From the comparisons below note the small change in the FWHM when adding the IPC in agreement with expectations. 
@mperrin note that I found a different value for the NIRCam charge diffuse/BBF effect 2D Gaussian kernel proxy, with the IUPC effects included I have found that `nrc.options['charge_diffusion_sigma'] = 0.0055` gives the best results. We could change this on the constant file but people are free to change this accordingly via options.
NIRCam IPC
image
FWHM (x,y), same order as in the figure: 2.2871096601827006 2.3196544874218388 2.2988609633440427 2.3311552268678577 2.2966360119439004 2.33525820977394
For NIRCam I am using the ePSF from our TA observations downsampled to the detector pixel size. 

NIRISS IPC:
image

FWHM (x,y), same order as in the figure: 1.5226913825644413 1.4257900603334979 1.5406125390304537 1.4442808131561855
Note that NIRIS kernel selection depends on the position of the detector

MIRI IPC:
image
FWHM (x,y), same order as in the figure: 2.8745429492672843 2.970422595606055 2.965914817067365 3.04349546844798


This is a sample code used for the NIRISS comparison, with and without the IPC effect. 
```
#nominal NIRISS PSF without IPC effects
niriss = webbpsf.NIRISS()
niriss.detector_position = (200, 300)
niriss.options['add_ipc'] =  False
single_webbpsf_niriss = niriss.calc_psf(fov_pixels=100)

# NIRISS PSF with IPC effects
niriss = webbpsf.NIRISS()
niriss.detector_position = (200, 300)
single_webbpsf_niriss_ipc = niriss.calc_psf(fov_pixels=100)
```
IMPORTANT: IPC effect are only included in the 'DET_DIST' extension

marcio-melendez:
This file contains NIRISS kernels, it needs to be unpack under `$WEBBPSF_PATH/NIRISS/IPC `
This new folder, `NIRISS/IPC`, will be added to the next webbpsf-data release, > 1.1.1

[NIRISS_IPC.zip](https://github.com/spacetelescope/webbpsf/files/11850149/NIRISS_IPC.zip)

mperrin:
All sounds good. 

It makes total sense to me that we will want different (reduced) values for the charge transfer diffusion kernel once the IPC is added in. The effect is divided up among multiple physical effects. 

I see in the comments you wrote: 
```
# For NIRCam see, Moore, A.C., Ninkov, Z., Forrest, W.J., 2006, Optical Engineering, 45, 7 
```
We can surely do better and more recent for NIRcam than something 15 years pre-flight! There's IPC kernels in MAST CRDS from 2016, for instance. And @JarronL shared with me some more recent versions that he derived in-flight. Should be easy to pick some reasonably recent measurement and swap that in.

mperrin:
btw, is there a chance the labels are swapped in the NIRCam image above? The left image labeled "No IPC" looks slightly blurrier to me?

marcio-melendez:
Re: NIRCam image. I just replaced the image. The "No IPC" had the default H2RG detector effect, I found that after the IPC, a sigma of 0.0055 works better. All simulations in that figure now have 0.0055, the only thing that changes is  with/without IPC effect. 
Re: NIRCam IPC. That paper was more like a place holder. I am using the default constant value in pyNRC, which I don't think it comes from that paper and it seems in agreement with values reported in the NIRCam readthedocs, 0.0065. I can take a look at some recent kernels to update NIRCam.

JarronL:
Here are the IPC and PPC values for each NIRCam detector (labeled as 481-490) that I derived during ground I&T, primarily ISIM-CV3 after tuning all of the arrays. Flight values should be the same, although would be great spot check some of these values.

[NIRCam_IPC_PPC_kernels.zip](https://github.com/spacetelescope/webbpsf/files/11885694/NIRCam_IPC_PPC_kernels.zip)

The first image in the cube are the kernel values with a sum of 1.0. The 2nd image in the cube is the measured uncertainty for each element in the kernel.

marcio-melendez:
I added the IPC/PPC effect using @JarronL values. I combined the IPC/PPC files into a fits cube (see attach). These data cubes will be inside webbpsf-data release and once we have the flight values we can just replace the files. To test this PR you need to add these files (KERNEL_IPC_CUBE.fits and KERNEL_PPC_CUBE.fits) under $WEBBPSF_PATH/NIRCam/IPC/
[KERNEL_CUBES.zip](https://github.com/spacetelescope/webbpsf/files/11887771/KERNEL_CUBES.zip)
Below a simple test case 
```
# Nominal NIRCam 
nrc = webbpsf.NIRCam()
nrc.detector = 'NRCA3'
nrc.filter='F212N'
nrc.aperturename = 'NRCA3_FP1'
nrc.options['charge_diffusion_sigma'] = 0.0055
nrc.options['add_ipc'] =  False
single_webbpsf_nircam = nrc.calc_psf(fov_pixels=59)

#  NIRCam IPC and PPC
nrc = webbpsf.NIRCam()
nrc.detector = 'NRCA3'
nrc.filter='F212N'
nrc.aperturename = 'NRCA3_FP1'
nrc.options['charge_diffusion_sigma'] = 0.0055
single_webbpsf_nircam_ipc = nrc.calc_psf(fov_pixels=59)
```
image

marcio-melendez:
@Skyhawk172 can you take a look at this PR? this PR started as H2RG detector effects but then I added other detector effects (IPC/PPC) into it. Note that for NIRISS and NIRCam there are a few files that you need to download and add to: `$WEBBPSF_PATH/NIRCam[NIRISS]/IPC/ ` All the files are in this PR but also included in our webbpsf-data source path. I added a couple of test/example code on how to test these changes. Thanks!

Skyhawk172:
I'm finding a little of a discrepancy in how the two effects (charge diffusion and IPC) are being applied in the different extensions.  

By default, charge diffusion is applied to the OVER_DIST extension and gets carried over to the DET_DIST extension when the downsampling is done. That's also how the distortion gets applied. However, IPC is applied only to the DET_DIST extension, which results in two "distorted" extensions that do not include the same effects.  

For consistency, I propose `detectors.apply_detector_ipc` gets called right after `detectors.apply_detector_charge_diffusion` in the code, unless there is a reason to believe applying the IPC on an oversampled PSF then downsampling won't yield the same result as applying the IPC directly on the detector-sampled PSF.

marcio-melendez:
The IPC kernels are the results from the instrument performance thus, in principle,  we cannot convolve the same kernels to the oversampled PSF. Therefore,  we have to change  the kernels, meaning, resampling the kernels to the oversample factor. So, in the code, we'll need to resample the kernel to whatever the oversample factor is and then webbpsf will downsampled the image to the detector scale. 
Overall, I see the discrepancy of having 2 extensions with different effects inside but given the small uncertainties associated with interpolations, I figured that resampling and downsampling the IPC effect will defeat the purpose of adding this already small effect. However, I have not done the testing of resampling the kernel and applying it right after charge diffusion and inspect both PSFs. It is very likely that the difference will be  are very small. 
I'll give it a try.

Skyhawk172:
@obi-wan76 Just checking back on the checkboxes on the original comment, I agree that moving the `apply_miri_scattering` code from `distortion.py` to `detectors.py` would make it more consistent w.r.t. how detector effects are handled in the code, and should be easy to do.

marcio-melendez:
I moved `apply_miri_scattering` code from `distortion.py `to `detectors.py`
I performed a few other comparisons between observations and predictions in order to update the charge diffusion default sigma values after IPC effects are taken into account. 
FGS I used a set of FGS_ID images in order to find a suitable PSF to compare:
image
FWHM (no Charge diffusion, with charge diffusion and observation):
1.3450907222404094 1.5432953946411594 2.874607291260897 2.8337080558381276 2.8438461464913685 2.9624240423015613
```
# FGS no charge difussion
fgs = webbpsf.FGS()
fgs.options['add_ipc'] =  False
fgs.options['charge_diffusion_sigma'] = 0.
single_webbpsf_fgs = fgs.calc_psf(fov_pixels=100)

# FGS charge difussion
fgs = webbpsf.FGS()
fgs.options['charge_diffusion_sigma'] = 0.07
single_webbpsf_fgs_ipc = fgs.calc_psf(fov_pixels=100)
```
Note that there are no IPC kernels for FGS, thus, FGS only has the effect of charge diffusion. The extension 'DET_DIST' has a disclamer that no IPC are in FGS (header) 

NIRISS 
In order to test and update the charge diffusion parameter I used an ePSF created during MIMF-2 for the inner 20px of NIRISS (to avoid contamination from nearby  sources), F158M
image

FWHM (no effects, charge diffusion/IPC, ePSF) = 1.1812243511339484 0.9696483724261505 1.4495761008744288 1.4839461649359382 1.4550414671206466 1.428810659191833

```
niriss = webbpsf.NIRISS()
#niriss.detector_position = (200, 300)
niriss.options['add_ipc'] =  False
niriss.options['charge_diffusion_sigma'] = 0.
single_webbpsf_niriss = niriss.calc_psf(fov_pixels=21)

# NIRISS PSF with IPC effects
niriss = webbpsf.NIRISS()
#niriss.detector_position = (200, 300)
niriss.options['charge_diffusion_sigma'] = 0.028
single_webbpsf_niriss_ipc = niriss.calc_psf(fov_pixels=21)
```

For MIRI U used both an ePSF , created during MIMF-3, and a single PSF from a calibrated product (see figure for name) 
image

FWHM (no IPC/charge difuse effect, all effects, ePSF, and single PSF observations)
(1.7474256268954576, 1.8117374898638132, 2.2462261200997116, 2.201854474767499, 1.8777339833323434, 1.8451831354573198, 2.2460185279725877, 2.0968598967856082)

```
X = 699.23
Y = 822.54
#nominal MIRI PSF without IPC effects
miri = webbpsf.MIRI()
miri.filter = 'F560W'
miri.options['add_ipc'] =  False
miri.options['charge_diffusion_sigma'] = 0.
miri.detector_position = (X, Y)
single_webbpsf_miri = miri.calc_psf(fov_pixels=51)

# MIRI PSF with IPC effects
miri = webbpsf.MIRI()
miri.filter = 'F560W'
miri.options['charge_diffusion_sigma'] = 0.050
miri.detector_position = (X, Y)
single_webbpsf_miri_ipc = miri.calc_psf(fov_pixels=51)
```

marcio-melendez:
This last push checks all the checkboxes in this PR. 
Note that after some discussion we decided, for now,  to keep the IPC effects only in the 'DET_DIST' extension. The IPC effects are small thus changing the convolution kernels to match the oversampled factor (default or provided by the user) may introduce some numerical artifacts that could match the magnitude of the IPC effect itself. 
However, depending on user feedback, we may revise this implementation and change it,  if needed.

mperrin:
I'm catching up on this... Looks like this is near ready to go so I flipped the PR state from Draft to Ready for review 

@BradleySappington FYI I just updated and rebased this branch to be consistent with the latest in `develop`. 

@obi-wan76 @Skyhawk172 I think I agree with your choice to keep the IPC being used only in the detector-sampled extension. That's a discrete pixel effect so it's not intuitive to convert it for use on an oversampled PSF... Doing it only for the detector-sampled extension was my initial thought too.  _On the other hand_, recall that computation of ePSFs using webbpsf does not use the detector-sampled extension `DET_DIST`; it works from the oversampled extension (`OVERDIST`)... So with this approach we would have to separately add the IPC+PPC convolution as a step in the ePSF generator code too. Hmm. 

Recall also that convolution is commutative, so in theory we ought to be able to flip the order of the IPC+PPC convolution and the binning (which is mathematically also a kind of convolution), and get the same answer. In practice that might take a little bit of thinking to ensure getting the details right for sampling and offsets and so on. But it's a tradeoff between doing that, versus having to separately re-implement the IPC+PPC convolution as part of the ePSF code. What do you think?

mperrin:
CI tests all failing because the data files used do not yet include the new reference files for NIRCam that @obi-wan76 references above.

webbpsf/detectors.py
```diff
@@ -0,0 +1,329 @@
...
+        ipcname = 'ipc5by5median_amp' + frag1[nchannel] + '_' + \
+                      frag2[flag] + '.fits'
+        ipc_file = os.path.join(utils.get_webbpsf_data_path(), 'NIRISS', 'IPC', ipcname)
+        if os.path.exists(ipc_file):
+            ipckernel = fits.getdata(ipc_file)
+            newimage = signal.fftconvolve(image, ipckernel, mode='same')
+            psf_hdulist[ext].header['IPCINST'] = ('NIRISS', 'Interpixel capacitance (IPC)')
+            psf_hdulist[ext].header['IPCTYPA'] = (ipcname, 'kernel file used for IPC correction')
+        else:
+            newimage = numpy.copy(image)
```
BradleySappington:
```suggestion
            newimage = np.copy(image)
```

BradleySappington: small bug that breaks one of your examples in the pr
(NIRISS PSF with IPC effects)

BradleySappington (changes requested):
I'm not qualified to review the logic behind the algorithms, however I did find a small bug.

webbpsf/detectors.py
mperrin: Fixed just now, thanks.

mperrin:
These latest edits get the tests all passing locally for me; let's see if they also get it working on here.

marcio-melendez (approved):
Looks good, all comments are inside this PR.

mperrin:
Before this is merged I think the significant fix needed is to update the `psf_grid` code to also include the IPC.  

I almost worked on that today (I have been thinking about how to do so), but ended up spending most of the day on NIRSpec and MIRI datasets instead 😄.  

Adding the IPC into the oversampled ePSFs from `psf_grid` will necessarily involve writing an implementation of that which works on oversampled PSFs. At that point it might be easier and cleaner to simply make that work on the OVERDIST extension directly, rather than having separate redundant implementations in regular `calc_psf` for DET_DIST and in `psf_grid` for OVERDIST...

marcio-melendez:
Yes, sounds good, no need to rush this PR until we are happy on how is performing. I am also wondering if a created oversampled ePSF from observations preserved  IPC effects, in other words, do we know the pixel correlation on an "observed" ePSF.

mperrin:
I'm not sure I follow exactly what you're asking... empirical PSFs created from real data inherently include all detector effects that were present in the real data, yes.  (Since the pipeline could but does _not_ currently attempt to calibrate out IPC, for instance. So that effect is present in the observed PSFs that are used to make measured ePSFs.)

marcio-melendez:
Yes, I understand that an ePSF from real data will carry the detector effect. However, there are intrinsic uncertainties related to correct centroid placement and sampling of the pixel phase space when building an ePSF. So, I am wondering if  IPC effects get  diluted because of this systematic uncertainties when combining different observed PSFs during the process of creating an ePSF. I agree with the proposition of adding the effect into the OVERDIST, my comment is more about how to compare with real ePSF as, based on my photutil experience, not all ePSF are created the same way and it may be difficult to see the already small IPC effect out of the systematic uncertainties when creating a ePSF.

BradleySappington (approved):

marcio-melendez:
We are still waiting for some changes in this PR with some further review. So, no merging for now.

marcio-melendez:
TODO - If someone wants to run a simulation with `add_distortion = Fals`e  then the `apply_detector_IPC` will return an error because there is no `DET_DIST` extension. For example, the monthly trending function is  returning an error because of this.

mperrin:
Actually I fixed that TODO already in the commits I'm about to push :-)

mperrin:
Latest commits include:

- Updated `psf_grid` to include IPC(+PPC for NIRCam) in the ePSFs, implemented via an oversampled zero-padded version of the IPC convolution kernel. Also updated to copy IPC and charge diffusion FITS header keywords to the PSG grid. 
- Refactored `apply_ipc_kernel` into two functions to separate (a) obtaining the IPC kernel and (b) applying the IPC kernel. This made it cleaner and easier to implement step (c) optional conversion to oversampled form, in between a and b. 
- Added unit test for the equivalence of the two versions of the IPC convolution
- Updated checks for `add_distortion=False` and so on, to avoid trying to apply IPC model if the relevant extension isn't being computed

Skyhawk172:
Regarding WFI, note that I have tracked down the location of the IPC files on CRDS, but I have yet to determine if these are actually valid or just placeholders.

In any case, there will be Thermal-Vacuum testing in the fall and I believe we will be getting updated values.

mperrin:
This will need further updates for the tests. The addition of IPC into the epsf calculations improves model fidelity, but apparently breaks some of the simplifying assumptions used in a couple of the unit tests. 



Meanwhile, here's results from trying to manually tune up the charge diffusion sigma parameters in comparison with available ePSFs. (This is still using a Gaussian model rather than sech, and with no wavelength dependence in the model, not apart from having different coeffs per NIRCam SW and LW)

epsf_comparison_nircam_0 007_0 017

epsf_comparison_niriss_0 021

marcio-melendez:
The observations versus comparisons looks really good, only a couple of percentage difference in the worst case, ~2.7um. I ran all my previous examples, including trending plots, with no issues with the new additions and improvements. 
Just to confirm, we have the ability to add the distortion to the  `OVERDIST` extension but by default it is only in the `DET_DIST`. In other words, for a normal calculation the detector effects are not included in both extensions at the same time. Is this correct?

mperrin:
> Just to confirm, we have the ability to add the distortion to the  OVERDIST extension but by default it is only in the DET_DIST. In other words, for a normal calculation the detector effects are not included in both extensions at the same time. Is this correct?

That's the current implementation, but that could be revisited. I decided that (since you and Charles had previously said you preferred adding the IPC at the detector-pixel level, after binning to detector resolution) to leave that unchanged. So I only added the oversampled version of IPC in the psf_grid pathway only. Up to you if you want to revisit the preference for using the simpler more intuitive version of the IPC in the normal calc_psf calculations...

> The observations versus comparisons looks really good, only a couple of percentage difference in the worst case, ~2.7um. I ran all my previous examples, including trending plots, with no issues with the new additions and improvements.

I developed a test comparison notebook that varies the charge diffusion parameter, calculates ePSFs, and compares to the suite of measured ePSFs. I used this to explore a range of values for the charge diffusion for NIRCam and NIRISS to find values that seem to fit well. (Not a rigorous best-fit optimization, just a manual exploration of a range of values for each parameter). The comparison metric is the PSF FWHM on the ePSFs for both the measured and sim ePSFs. 

The results from this suggest using values around the following. These aren't perfect fits but seem reasonably decent to me. 
 - NIRCam SW: 0.065
 - NIRCam LW: 0.017
 - NIRISS: 0.021

epsf_comparison_nircam_0 0065_0 017


epsf_comparison_niriss_0 021

mperrin:
Oh, I just realized I posted those plots on Friday as well (I forgot I had already done that...) so the previous post is fairly redundant. Slight change in the recommended value for NRC SW is the only difference.

mperrin:
@obi-wan76 FYI - Tests are now passing, and all my intended code edits are complete, with the exception of adjusting the charge diffusion coefficients to better match the measured ePSFs, which could go in a separate PR. 

I think this PR is ready for either re-review and merge, as is, or we could add in here the adjustments to the diffusion parameters, if you concur with the values I have above. 

I'm planning to add in the ePSF comparison code as some extra files into dev_utils, but have not yet had a chance to add that in yet. 

My preference would be to consider merging this as-is, rather than waiting longer, then do the parameter tuning and add the ePSF comparison code in a separate PR. Motivated by getting this change merged in without further delay at this point  since it's already a good improvement as is. But your choice.

marcio-melendez:
Yes, I agree to merge as-is, my latest tests are also running. The fine tuning of values could be address in the documentation, or another PR. In any case, we need to create the documentation about detector effects that will go with the release. I'll work on that.

PR #675, opened 2023-05-31
For coronagraph Lyot masks, add reference information on alignment per each mask

mperrin:
This PR implements a capability to have default mask alignment options (`pupil_shift_x, pupil_shift_y, pupil_rotation`) defined per each coronagraph mask. This is intended to allow WebbPSF to include measured in-flight information on the various masks. 

Implementation details: 
- in `constants.py` add a dict which gives values of these options per each mask, or else None to leave undefined
- in `webbpsf_core.py`, update code to look up and use those values. 

_WORK IN PROGRESS, NOT YET COMPLETE, NOT READY TO MERGE_

### DIFF
--- a/webbpsf/constants.py
+++ b/webbpsf/constants.py
@@ -400,3 +400,15 @@ INSTRUMENT_IPC_DEFAULT_KERNEL_PARAMETERS = {
 # See Gaspar et al. 2021 for illustrative figures.
 # This is a rough approximation of a detector-position-dependent phenomenon
 MIRI_CRUCIFORM_INNER_RADIUS_PIX = 12
+
+# Alignment information about instrument internal pupil masks (
+INSTRUMENT_PUPIL_MASK_DEFAULT_POSITIONS = {
+    'NIRCam_MASKSWB': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'NIRCam_MASKLWB': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'NIRCam_MASKRND_SW': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'NIRCam_MASKRND_LW': {'pupil_shift_x': -0.012, 'pupil_shift_y': -0.023, 'pupil_rotation': -0.60},  # from K. Lawson, fits to ERS progid 1386 data
+    'MIRI_MASKFQPM_F1065C': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'MIRI_MASKFQPM_F11140': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'MIRI_MASKFQPM_F1550C': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'MIRI_MASKLYOT': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+}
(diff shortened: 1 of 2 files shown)


webbpsf/constants.py
```diff
@@ -243,3 +243,15 @@
 JWST_INSCRIBED_DIAMETER = 5.47334  # meters. Middle corners of C segments
 
 JWST_TYPICAL_LOS_JITTER_PER_AXIS = 0.0008 # milliarcseconds jitter, 1 sigma per axis. = approx 1 mas rms radial, typically
+
+# Alignment information about instrument internal pupil masks (
+INSTRUMENT_PUPIL_MASK_DEFAULT_POSITIONS = {
```
JarronL, line 404: NIRCam will need multiple values that depend on channel (SW or LW pupil wheel) and mask (field position):
 - SWA_MASK210R
 - SWA_MASK335R
 - SWA_MASK430R
 - SWA_MASKSWB
 - SWA_MASKLWB
 - LWA_MASK210R
 - LWA_MASK335R
 - LWA_MASK430R
 - LWA_MASKSWB
 - LWA_MASKLWB

We could conceivably have similar values for Module B, but we currently have zero information on those masks.

webbpsf/constants.py
```diff
@@ -243,3 +243,15 @@
...
+    'NIRCam_MASKSWB': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'NIRCam_MASKLWB': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'NIRCam_MASKRND_SW': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
+    'NIRCam_MASKRND_LW': {'pupil_shift_x': -0.012, 'pupil_shift_y': -0.023, 'pupil_rotation': -0.60},  # from K. Lawson, fits to ERS progid 1386 data
```
JarronL: This is presumably only for the M335R mask as M430R will have some pupil wander and therefore slightly different values.

webbpsf/constants.py
mperrin: I thought that physically there is in the SWA wheel and LWA only one each of the bar and round Lyot stops? Such that, for instance, the 210R, 335R, 430R focal plane masks would all be used with the same physical Lyot mask?  And likewise the SWA_MASKSWB and SWA_MASKLWB would similarly use the same  physical Lyot pupil stop?

I had thought that therefore, the Lyot alignment offsets should be independent of focal plane position? What makes that not the case? Is this from field-position-dependent pupil wander within the instrument?

JarronL: Yes, it appears that field-dependent pupil wander changes the best-fit pupil shift values for each of the occulting masks. We found this to be the case during commissioning when tuning the pupil wheel clocking to find a good balance between the M335R and M430R masks.

mperrin: OK. This is going to be a minor pain for book-keeping, because it means the pupil alignment depends on the selected image mask (as well as the pupil mask)... but sometimes it's convenient to do coronagraphic calculations with the image mask left out to get an off-axis PSF (i.e. a sim with image_mask=None and pupil_mask=MASKLWB, or similar). For calculations like that, right now there would be no way for webbpsf to know which image mask to use for the pupil   alignment parameter lookup. 

This feels like a minor practical detail of API and implementation, but it'll need at least a little thought for how we want this to work.

JarronL: The pupil shear has negligible impact on the shape of off-axis PSF. The main thing we care about is the morphology of the PSF under the influence of the occulting mask. For on-axis PSFs, the scattered light in the pupil matters (and therefore we care about alignment of the Lyot stop with the telescope pupil), whereas for off-axis PSFs, the pupil is filled. So in the end, we only care about the cases where `image_mask` is set.

mperrin:
Future development and implementation of this PR has migrated to https://github.com/spacetelescope/stpsf/pull/51

PR #690, opened 2023-07-12
Add missing astroquery dependency

larrybradley:


BradleySappington (approved):
looks good!

BradleySappington (commented):
is there a specific version we want to pin here?

larrybradley:
The `requirements.txt` file in this repo pins exactly to 0.4.6, which is the latest release: 
 https://github.com/spacetelescope/webbpsf/blob/b219a7db648681c107b5ef78a2a50574f1214fe9/requirements.txt#L10

That seems very specific, but I can update this PR to use that.  Internally, `webbpsf` is using `astroquery.mast`.  I'm not sure if that API changes frequently.

mperrin:
FYI @larrybradley, I agree having the "requirements" all pinned to exactly specifically the latest version is a little surprising and counterintuitive. That turns out to be from the use of Dependabot to manage dependencies. That service automatically updates the requirements.txt file to the latest versions of things. This has the nice effect of ensuring the CI runs with the latest version of every package, but on the other hand means that semantically the "requirements.txt" is not truly the minimum required version. Tradeoffs... 🤷‍♂️

larrybradley:
There's also a discrepancy between the *actual* package runtime requirements as specfied in the `pyproject.toml` file:
https://github.com/spacetelescope/webbpsf/blob/develop/pyproject.toml#L16

and the dependencies in `requirements.txt`.  It appears the install docs and various CI point to `requirements.txt`.  Not sure why that is needed  for CI since the latest versions will be used by default.

BradleySappington:
The pyproject dependencies are all loosely pinned (above a certain version), while the requirements.txt are all the latest (thanks to dependabot).  This gives two install options, if you install via pip with no special commands, pyproject will be used.  If you want to install with the latest version only, the user can install using requirements.txt via `pip install . -r requirements.txt`.

BradleySappington (changes requested):
Just had a discussion with @obi-wan76, since astroquery isn't needed for base functionality, we should move the dependency down to the [project.optional-dependencies] section.

larrybradley:
I've updated this PR to move `astroquery` as an optional dependency under the `all` variant.

BradleySappington (approved):
Thanks @larrybradley

PR #691, opened 2023-07-19
Add MIRI IFU (MRS) functionality to MIRI class

patapisp:
Work in progress to add functionality for the MIRI IFU - Medium Resolution Spectrometer. 

Added quite a few changes to `webbpsf_core.py:MIRI` to accommodate the instrument setup for the MRS. 

- added (hardcoded) attributes for the rotation angle of the IFU FoV, wavelength ranges of the IFU spectral channels, and updated the `_IFU_pixelscale`
-  added two new properties of `wavelength` and `band`
- added two (static methods) that calculate the in-flight measured FWHM and resolving power
- overwrote class methods for `filter.setter`, `image_mask.setter`, `detector.setter`, `aperture name.setter`
- added method `_update_detector` that deals with MIRI specific detector architecture
- overwrote class method `calc_psf` to accommodate the MIRI IFU PSF broadening
- overwrote class method `calc_datacube` that calculate multi wavelength cubes

I hope the general structure and implementation follows `webbpsf` guidelines and I tried to use as much of the already implemented functionality as possible.

To Do:

- [ ] Add to webbpsf documentation
- [ ] Add test for MIRI-IFU
- [ ] Add filters to filters.csv
- [ ] Implement IFU detector PSF projection for use in forward modelling application 
- [ ] Add detector scattering effect


marcio-melendez:
@patapisp thanks so much for this PR. I added myself as a reviewer and I'll try to help with the testing and documentation.

patapisp:
@obi-wan76 thanks! As a start, this code should produce a monochromatic PSF for the MRS broadened to match the in-flight FWHM. I named the filters for the IFU "D[subband]" where D stands for dichroic (the filter that selects the wavelength range) and sub-band being one of SHORT/MEDIUM/LONG [(see docs)](https://jwst-docs.stsci.edu/jwst-mid-infrared-instrument/miri-observing-modes/miri-medium-resolution-spectroscopy).

The main issue for compatibility I think is the handling of the siaf apertures which are a bit more complicated for the MRS and resulted in a bit of a convoluted logic in the class.

```
import webbpsf

miri = webbpsf.MIRI()
miri.image_mask = "MIRI-IFU_1"
miri.filter = "DSHORT"

miri._print_mrs_config()
out = miri.calc_psf(monochromatic=miri.wavelength, oversample=5, fov_arcsec=8, add_distortion=False, display=False, broadening="both")

import matplotlib.pyplot as plt
import numpy as np
plt.figure()
plt.imshow(np.log10(out[2].data), origin="lower", vmin=-6)
plt.show()
```

image

mperrin:
@patapisp big picture question for IFU PSFs. It seems to me that we should try to be careful to distinguish between the intrinsic PSF, versus broadening from detector effects (brighter fatter and IPC and so on), versus broadening from pipeline processing and drizzling. 

How have you been thinking about IFU cube sampling for this PR? Are you targeting to have the simulated PSFs match the drizzled PSFs for the default cube output parameters, or something else? 

I expect you have thought about this in more detail than I have, so probably have more insights than I do...

patapisp:
@mperrin that's a good point, and indeed I have been struggling to decide a clear output. The code itself is of course able to reproduce all the different PSFs of the IFU. I would give the user the ability to produce whatever they want, but I would as a default produce the standard pipeline output (in ifualign mode). An additional PSF option is the PSF projected on the detector (a feature that might come later).

Following the webbpsf convention I use an extension for the oversampled and detector sampled monochromatic PSF, and the broadened PSF as measured in the pipeline cube. Any feedback here, especially for potential user consistency between MIRI and NIRSpec is welcome!

The model currently looks like this:

image

webbpsf/webbpsf_core.py
```diff
@@ -1656,27 +1664,72 @@ def __init__(self):
         self.options['pupil_shift_x'] = -0.0069 # CV3 on-orbit estimate (RPT028027) + OTIS delta from predicted (037134)
         self.options['pupil_shift_y'] = -0.0027
 
-        self.image_mask_list = ['FQPM1065', 'FQPM1140', 'FQPM1550', 'LYOT2300', 'LRS slit']
+        self.image_mask_list = ['FQPM1065', 'FQPM1140', 'FQPM1550', 'LYOT2300', 'LRS slit',
+                                "MIRI-IFU_1", "MIRI-IFU_2", "MIRI-IFU_3",  "MIRI-IFU_4"]
+
```
mperrin, line 1669: One of the lessons we've learned from NIRSpec sims is that it does not seem to be particularly useful to include in the simulations the slicer field stop as an image mask: At times what's most useful is to simulate a PSF that's larger than the IFU FOV, so it has some margin and can be shifted around to match the position of an observed point source that's not directly centered in the cube.  In which case we do not want to simulate the image mask as a stop around the edge of the FOV. 

If I were to do it over again, I think it might be useful to have a `mode` attribute that could indicated IFU or not, independent of any field stop image mask. It might still make sense to add that to the NIRSpec class; I may implement that in a separate PR and see if it helps with the NIRSpec code. 

This may be  a pretty significant point of software design to decide about, since I think it might simplify the logic in other places, as well as being critical for enabling PSF simulations larger than the FOV. 

(Yes in practice if we wanted to _observe_ a PSF bigger than the FOV, we would need to mosaic and combine the IFU images. But we should not necessarily force ourselves to _simulate_ it that way...)

webbpsf/webbpsf_core.py
```diff
@@ -1656,27 +1664,72 @@ def __init__(self):
...
+                         "1C": [6.513777066360325, 7.669147994055998],
+                         "2A": [7.494966046398437, 8.782517027772244],
+                         "2B": [8.651469658142522, 10.168811217793243],
+                         "2C": [9.995281242621394, 11.73039280033565],
+                         "3A": [11.529088518317131, 13.491500288051483],
+                         "3B": [13.272122736770127, 15.550153182343314],
+                         "3C": [15.389530615108631, 18.04357852656418],
+                         "4A": [17.686540162850203, 20.973301482912323],
+                         "4B": [20.671069749545193, 24.476094964546686],
+                         "4C": [24.19608171436692, 28.64871057821349]}
```
mperrin, line 1695: Where do these values come from? Is it possible they will ever change slightly (with improved calibrations or similar), and if so can they be retrieved from some CRDS file?

I assume the units of these are wavelengths?

webbpsf/webbpsf_core.py
```diff
@@ -1656,27 +1664,72 @@ def __init__(self):
...
+                             "1C": 8.3,
+                             "2A": 8.15,
+                             "2B": 8.24,
+                             "2C": 8.17,
+                             "3A": 7.6,
+                             "3B": 7.6,
+                             "3C": 7.6,
+                             "4A": 8.73,
+                             "4B": 9.09,
+                             "4C": 8.43}
```
mperrin, line 1708: Can the above angle be retrieved from the SIAF apertures? 

I think it's good practice to avoid hard-coding in webbpsf numbers that we can retrieve from elsewhere (the "don't repeat yourself" principle of programming...), but sometimes it's unavoidable

webbpsf/webbpsf_core.py
```diff
@@ -1656,27 +1664,72 @@ def __init__(self):
...
+
+        # manually add miri filters: FIXME
+        self.filter_list.extend(['DSHORT', 'DMEDIUM', "DLONG"])
+        for f in ['DSHORT', 'DMEDIUM', "DLONG"]:
+            self._filters[f] = Filter(name=f, filename="N/A", default_nlambda=9)
```
mperrin, line 1713: Naming question here. I see the pipeline uses LONG, MEDIUM, SHORT like 
```
BAND    = 'LONG    '           / MRS wavelength band                            
```
Should we aim for consistency with that?

webbpsf/webbpsf_core.py
```diff
@@ -1656,27 +1664,72 @@ def __init__(self):
...
+
         self.auto_aperturename = True
 
-        self.monochromatic = 8.0
+        self.monochromatic = 5.0
```
mperrin, line 1717: There's already a `monochromatic` parameter to the calc_psf function, so I'm not sure I follow the need for adding a monochromatic attribute too? Especially since you also add a `wavelength` attribute. 

I have only read part of this PR in detail so far so I may not yet understand your plan fully. Can you explain more the thinking here?

webbpsf/webbpsf_core.py
```diff
@@ -1656,27 +1664,72 @@ def __init__(self):
...
+        # pixel scale corresponds to the slice width and the pixel size in the pipeline cube
         self._IFU_pixelscale = {
-            'Ch1': (0.18, 0.19),
-            'Ch2': (0.28, 0.19),
-            'Ch3': (0.39, 0.24),
-            'Ch4': (0.64, 0.27),
+            'Ch1': (0.18, 0.13),
+            'Ch2': (0.28, 0.17),
+            'Ch3': (0.39, 0.2),
+            'Ch4': (0.64, 0.34),
```
mperrin, line 1727: Same question as above: consider whether to retrieve this from pysiaf?  

That said, if we do want to hard-code values here it's OK, but we should add a comment indicating where they come from (which version of SIAF, or other source of calibration information etc.). I realize the detailed SIAF is difficult to work with and doesn't have a single "pixel scale" value easily accessible for the IFU apertures, so I concur it may be most practical to hard-code values here.

webbpsf/webbpsf_core.py
```diff
@@ -1686,9 +1739,73 @@ def _get_default_fov(self):
...
+    @wavelength.setter
+    def wavelength(self, value):
+        if self.MRSbands["1A"][0] <= value <= self.MRSbands["4C"][1]:
+            self._wavelength = value
+            # check if band needs to be changed
+            if not self.MRSbands[self._band][0] <= value <= self.MRSbands[self._band][1]:
+                band = self._find_wavelength_band()
+                self.band = band
+        else:
+            print(f"Wavelength not covered by MRS. Reverting to current value: {self._wavelength}")
```
mperrin, line 1755: This might be better as a `raise ValueError` to directly inform the reader that something went wrong with their input, rather than simply ignoring it.

webbpsf/webbpsf_core.py
```diff
@@ -1709,6 +1826,105 @@ def _validate_config(self, **kwargs):
             raise NotImplementedError("The MIRI MRS is not yet implemented.")
```
mperrin, line 1826: This NotImplementedError can be removed as part of this PR...

webbpsf/webbpsf_core.py
```diff
@@ -1709,6 +1826,105 @@ def _validate_config(self, **kwargs):
...
+            ch = self._image_mask.split("_")[-1]
+            if ch not in ["1", "2", "3", "4"]:
+                raise ValueError(f"MIRI MRS does not have channel {ch}")
+
+            if self.filter[0] != "D":
+                self.filter = "DSHORT"
+                print("set filter to DSHORT for the MRS")
+
+            self.band = ch + self.dichroic_to_subband[self.filter[1:]]
+            self.wavelength = np.mean(self.MRSbands[self._band])
```
mperrin, line 1856: So `.wavelength` is intended a computed property, rather than a user-settable attribute? If so it might be better implemented as a `@property`.

webbpsf/webbpsf_core.py
```diff
@@ -1709,6 +1826,105 @@ def _validate_config(self, **kwargs):
...
+
+        self._update_aperturename()
+        self._update_detector()
+
+    @JWInstrument.detector.setter
+    def detector(self, value):
+        if value.upper() not in self.detector_list:
+            raise ValueError("Invalid detector. Valid detector names are: {}".format(', '.join(self.detector_list)))
+        self._detector = value.upper()
+        self._update_aperturename()  # automatically set an appropriate aperture name
```
mperrin, line 1869: All but the last line of this function looks like it can be replaced with a call to the superclass function

webbpsf/webbpsf_core.py
```diff
@@ -1709,6 +1826,105 @@ def _validate_config(self, **kwargs):
...
+        self._update_detector()
+
+    def _update_detector(self):
+        """Determine which detector of MIRI should be used.
+        """
+        # # Check if detector is correct
+        if self._image_mask is None:
```
mperrin, line 1876: This if statement would be a good place to use a `.mode` attribute, if implemented.

webbpsf/webbpsf_core.py
```diff
@@ -1852,15 +2067,160 @@ def _update_aperturename(self):
...
+    def aperturename(self, value):
+        """Set SIAF aperture name to new value, with validation.
 
-    def _get_fits_header(self, hdulist, options):
+        This also updates the pixelscale to the local value for that aperture, for a small precision enhancement.
+        """
+        # Explicitly update detector reference coordinates,
+        # otherwise old coordinates can persist under certain circumstances
+
+        # Get NIRCam SIAF apertures
```
mperrin, line 2086: Needs update; mentions NIRCam in comment text but this is MIRI

webbpsf/webbpsf_core.py
```diff
@@ -1852,15 +2067,160 @@ def _update_aperturename(self):
...
+    def _print_mrs_config(self):
+        """
+        Print the current IFU, dichtoic, detector, band, wavelength setup
+        """
+        print(f"IFU: {self._image_mask}, Dichroic: {self._filter}, slice width: {self._slice_width} \n "
+              f"Detector: {self._detector},  pixelscale: {self.pixelscale} \n"
+              f"MRS band: {self._band}, wavelength: {self._wavelength}")
+
+    def _project_psf_to_detector(self):
+        pass
```
mperrin, line 2110: suggest `raise NotImplementedError()`

webbpsf/webbpsf_core.py
```diff
@@ -1852,15 +2067,160 @@ def _update_aperturename(self):
...
+        if alpha_width is None:
+            psf_model_alpha_beta = np.apply_along_axis(lambda m: convolve(m, kernel_beta), axis=0, arr=psf_model)
+        else:
+            kernel_alpha = Gaussian1DKernel(stddev=alpha_width)
+            psf_model_alpha = np.apply_along_axis(lambda m: convolve(m, kernel_alpha), axis=1, arr=psf_model)
+            psf_model_alpha_beta = np.apply_along_axis(lambda m: convolve(m, kernel_beta), axis=0, arr=psf_model_alpha)
+        return psf_model_alpha_beta
+
+    @utils.combine_docstrings
+    def calc_psf(self, outfile=None, source=None, nlambda=None, monochromatic=None,
```
mperrin, line 2139: Rather than overriding the parent calc_psf function, there's already related infrastructure for handling this sort of post-calculation additional effect within the `_calc_psf_format_output` superclass function.  That may be a better place to invoke some of the below code, TBD, rather than adding a layer of if statements here at the top of the function call tree.

webbpsf/webbpsf_core.py
```diff
@@ -1852,15 +2067,160 @@ def _update_aperturename(self):
...
+        else:
+            # normal PSF calculation for MIRI
+            return super().calc_psf(outfile=outfile, source=source, nlambda=nlambda, monochromatic=monochromatic,
+                                    fov_arcsec=fov_arcsec, fov_pixels=fov_pixels, oversample=oversample,
+                                    detector_oversample=detector_oversample, fft_oversample=fft_oversample,
+                                    overwrite=overwrite, display=display, save_intermediates=save_intermediates,
+                                    return_intermediates=return_intermediates,
+                                    normalize=normalize, add_distortion=add_distortion, crop_psf=crop_psf)
+
+    def calc_datacube(self, wavelengths, **kwargs):
```
mperrin, line 2202: This overrides the parent function `calc_datacube` (not in webbpsf, inherited from poppy.Instrument) but I'm not sure what, if any, additional functionality this offers?  But perhaps you were not aware of that function already existing. 

There's also a `calc_datacube_fast` which I recently implemented for NIRSpec (See #767), and is likely also applicable for MIRI. I should refactor that to generalize.

mperrin (commented):
Hi @patapisp - It's been months but I have finally had time to read your draft PR for MRS IFU functionality in WebbPSF. Thanks again for this contribution. Some comments attached, see below.  

There are a couple places where I think we can revise slightly to fit better in with existing webbpsf infrastructure, as we work to move from draft PR to completed. This is all super complex so overall already I think this is going in a good direction. There's also some of the infrastructure that may be a moving target because of things I have been thinking about for NIRSpec IFU. In particular I've decided I think it's going to be very helpful to add an instrument attribute for `.mode` to indicate imaging or IFU, rather than relying just on SIAF aperture names. And I've started work today on some added framework for IFU code that can be common to both the NIRSpec and MIRI classes. Will put in a separate PR for that shortly.

webbpsf/webbpsf_core.py
patapisp: sorry, I know this is bad practice. It's in my to do list to fetch everything from CRDS.

mperrin:
@patapisp As discussed on Slack, we know you're working on MRS PSF improvements in a new set of edits based on the work in the IFU mode PR #770 (and now FYI there is also #890 just merged in today). Those we'll expect in some new future PR. So I am going to close this now-out-of-date prior draft PR now. Thanks again for all your work on this!

PR #693, opened 2023-07-25
Tune detector effects model parameters to better match measured ePSFs

mperrin:
Adjust various free parameters to better tune simulated ePSFs to match measured ePSFs. 

**MIRI:** 
- Add new parameter `psf_location_list` to psf grid creation, which can be used to specify custom locations for the PSFs in a grid. This is needed to match Mattia's ePSFs for MIRI which have the X samples at [0, 352, 1032]. 
- Reduce charge diffusion sigma to 0.001 (essentially zero), because the IPC effect alone appears to dominate for this.
- Reduce cruciform intensity by 0.5. For the innermost region,  within the inner 12 detector pixels, further reduce it down to 0.25.  

**NIRCam:**
- For SW, adjust charge_diffusion_sigma from 0.006 to 0.0062. This provides a better fit to the ensemble of measured ePSFs across all 8 SW detectors (at a cost of slightly less good fit to NRCA3 in particular)
- For LW, adjust charge_diffusion_sigma from 0.012 to 0.018. This provides a much better fit to the ensemble of measured ePSFs across both LW detectors, versus the prior value which was a hand wave I made up without testing. 

**NIRISS:**
 - Adjust charge_diffusion_sigma from 0.018 to 0.021. This provides a better fit to the full ensemble of measured ePSFs for NIRISS, though it does so by improving the longer wavelength filters but making the shorter wavelength filters fit less well. 
 -  This is debatable and we should think about it more maybe; this seems like the best current motivation for complicating this by adding a wavelength dependence... The SW filters would be better fit by using 0.020, the LW filters by 0.021, so maybe split the difference and set the default to 0.0205? Or add in a wavelength dependence.
 
 

Figures for MIRI, NIRCam, NIRISS follow. 

epsf_comparison_miri_0 001

epsf_comparison_nircam_0 0062_0 017

epsf_comparison_niriss_0 0202

### DIFF
--- a/webbpsf/constants.py
+++ b/webbpsf/constants.py
@@ -250,9 +250,9 @@ JWST_TYPICAL_LOS_JITTER_PER_AXIS = 0.0008 # milliarcseconds jitter, 1 sigma per 
 # Note, these are parameterized as arcseconds for convenience (and consistency with the jitter paramater)
 # but the underlying physics cares more about detector pixel pitch.
 INSTRUMENT_DETECTOR_CHARGE_DIFFUSION_DEFAULT_PARAMETERS = {
-    'NIRCAM_SW': 0.006,     # Fit by Marcio to WFS TA ePSFs
-    'NIRCAM_LW': 0.012,     # Scaled up by pixel pitch
-    'NIRISS': 0.028,         # Fit by Marcio to MIMF-3 F158M (ePSF)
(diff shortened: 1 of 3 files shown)


marcio-melendez (approved):
This looks really good, great fine tune of initial parameters. Also, the cruciform artifact from MIRI.

Issue #692, opened 2023-07-25
Data files out of date


JarronL:
With the release of the latest develop version of webbpsf, I appear to no longer have a valid set of webbpsf-data files:

```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jarron/NIRCam/webbpsf-data/NIRCam/IPC/KERNEL_IPC_CUBE.fits'
```

Is there an updated pre-release version stashed somewhere that I can grab?

marcio-melendez:
Yes, the Box link is: 
https://stsci.box.com/shared/static/ykkvequ9cbble6qg88ldhjoep4m9ey6x.gz

marcio-melendez:
This was fixed in the stable release

PR #694, opened 2023-07-25
Increment minimum data files version to 1.2.0

mperrin:
Partially addresses #692, by making webbpsf give an error if you don't have the new data files yet... Another PR will be needed to update the docs download URLs etc


marcio-melendez (approved):
Looks good, this is the first step to change the version number throughout

PR #696, opened 2023-07-26
Update import path for GriddedPSFModel

bmorris3:
Importing `GriddedPSFModel` directly from `photutils` without specifying the `psf` module has been deprecated since https://github.com/astropy/photutils/pull/1433, which causes WebbPSF calls to warn:
```
webbpsf/gridded_library.py:108: DeprecationWarning: `photutils.GriddedPSFModel` is a 
deprecated alias for `photutils.psf.GriddedPSFModel` and will be removed in the future.
Instead, please use `from photutils.psf import GriddedPSFModel` to silence this warning.
```
This PR first tries to import `GriddedPSFModel` from the `psf` module, and falls back on old the top-level import if that fails. Failing both of these, the `ImportError` is raised.

### DIFF
--- a/webbpsf/gridded_library.py
+++ b/webbpsf/gridded_library.py
@@ -112,9 +112,12 @@ class CreatePSFLibrary:
 
         # Before doing anything else, check that we have GriddedPSFModel
         try:
-            from photutils import GriddedPSFModel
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
This looks good, thanks!

PR #704, opened 2023-08-01
Update tox.ini

BradleySappington:
add astroquery to all test environments


marcio-melendez (approved):
Looks good, as we discussed.

PR #705, opened 2023-08-01
add trending plot function "show_wfs_around_obs"

mperrin:
Simple PR, adds one function "show_wfs_around_obs" that displays the WFS before and after a given science observation, and the spacing in time between those observations. 

Low priority addition, not critical, but I wrote this function a while ago and wanted to get it into a PR. 

example_output_plot

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -1345,3 +1345,115 @@ def plot_wfs_obs_delta(fn1, fn2, vmax_fraction=1.0):
         ax.yaxis.set_ticks([])
 
     return fig
+
+
+def show_wfs_around_obs(filename, verbose='True'):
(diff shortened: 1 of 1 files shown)


marcio-melendez:
Great addition. Again a test with a NIRCam LW. Everything looks good. 
image

marcio-melendez:
Looking at your example, why is the WFE the same before/after if there was a big tilt event in between? The top panel is the WFE, right?

mperrin:
Top panel actually just a horizontal timeline; there is no significance to any vertical axis there. Just separation in time horizontally.  It could be changed to be an RMS, in some future PR, but I made this top panel just showing time on one axis and nothing else.

marcio-melendez (approved):
looks good.

PR #707, opened 2023-08-02
Update README.rst post-JWST launch

jdavies-st:
Since JWST is now here, it is no longer "next".  😉


marcio-melendez (approved):
Looks good. It is certainly here already.

PR #709, opened 2023-08-04
update deprecated node12 references in ci_workflows.yml

BradleySappington:
node 12 is deprecated, need to update a couple versions


marcio-melendez (approved):
looks good

PR #713, opened 2023-08-10
Minor updates in order to run the WebbPSF-Roman_Tutorial notebook

Skyhawk172:
Minor updates the notebook, with a simple example of choosing a different filter and SCA.

Updated `jupyter_gui.py` so that the notebook GUI works.


marcio-melendez (approved):
Looks good, thanks!

PR #720, opened 2023-08-11
Pypi figure display

BradleySappington:
need full path for pypi to see figure


marcio-melendez (approved):
Looks good, I can confirm the figure  from the link.

PR #721, opened 2023-08-11
Pypi figure display (#720)

BradleySappington:
* Update README.rst

figure doesn't show up in pypi, use full path


marcio-melendez (approved):
looks good

PR #749, opened 2023-10-03
use PyPI upload workflow from OpenAstronomy

zacharyburnett:
Resolves [SCSB-108](https://jira.stsci.edu/browse/SCSB-108)

This PR changes the PyPI upload process to use the OpenAstronomy workflow instead of the STScI one. The workflow in this PR is configured to test the build with each pull request push, and then publish upon an official release.


BradleySappington (approved):
looks good!

PR #750, opened 2023-10-06
Trending: add option to plot OTE-only WFE in wfe_histogram_plot

Skyhawk172:
The updates in this PR provide the option to display the OTE-only WFE in the `wfe_histogram_plot` function. The main reason is that our correction threshold is set for OTE-only WFE, so plotting Observatory WFE with our correction threshold is misleading. That’ll give us the option to go between OTE+NRC and OTE-only easily.  The title and various labels should update appropriately whether the flag `ote_only` is set to True or False.

image


marcio-melendez (approved):
Looks good.

PR #756, opened 2023-10-16
Update trending plot to subtract piston before displaying proposed correction

mperrin:
The WSS doesn't subtract piston from the proposed correction that it writes out in the OPD files. (The MCS does subsequently remove global piston before generating SURs). 

This PR changes the display code in single_measurement_trending_plot so it subtracts the piston before displaying the proposed correction. This also affects the computed RMS WFE values, removing a bias that was making those appear higher. (related to the subtle distinction between the RMS and the std dev, which the WSS code handles differently in various cases). 

Before and after plots. Only change is in the marked panel in the bottom row: 

**Before (with current code):**

Unknown

**After this PR (with proposed modified code)**

Unknown-1

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -661,6 +661,10 @@ def single_measurement_trending_plot(opdtable, row_index=-1, reference=None, ver
     iax = axes[2, 2]
 
     if show_correction:
+        # Note, the WSS does not remove piston from the correction so it's not zero mean.
+        # Fix that here before display. This also affects the derived rms value displayed.
+        correction -= np.nanmean(correction[correction_mask==1])
+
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Looks good and very useful to help visualized these last set of correction. Sometimes we subtract the piston manually from the correction.  
image

marcio-melendez:
@BradleySappington can you check this merge? Thanks!

BradleySappington:
@mperrin - please rebase to get the CI legacy test fix

BradleySappington:
@obi-wan76 @mperrin, mentioned a few weeks back (although cycle deadline was in full force when I mentioned it!).  Please re-base to get the CI legacy fix :)

PR #757, opened 2023-10-19
Install astropy dev from wheel

pllim:
You do not need to build it from source. We distribute nightly wheel.


BradleySappington (approved):
Thanks for pointing this out!

PR #755, opened 2023-10-13
import GriddedPSFModel from photutils.psf

braingram:
This PR updates the remaining `GriddedPSFModel` imports from `photutils` to prefer importing from the `psf` submodule to avoid deprecation warnings issued when importing from the top level `photutils` module.

The modifications here are equivalent to the existing `try/except` found in `gridded_library.py`:
https://github.com/spacetelescope/webbpsf/blob/be742288089484d1ac94096d0bce33d5d2adb2be/webbpsf/gridded_library.py#L113-L120

PR #758, opened 2023-10-23
Astroquery version

BradleySappington:
WebbPSF had a dependency issue during the legacy test.  Updating minimum versions and legacy pin to be in-line with poppy


BradleySappington:
@obi-wan76 - ready for review

should solve build issues for:
#757
#756
#755

marcio-melendez (approved):
This looks good.

PR #760, opened 2023-10-27
Add pytest plugin to mark tests as requiring the "remote" webbpsf related data

WilliamJamieson:
Some uses of WebbPSF require the use of specific external data. This means that if users want to test their code which uses WebbPSF they must get this data. This makes general testing of those downstream pacakges quite complex.

This PR adds a `pytest` plugin to WebbPSF so that a special marker can be defined `pytest.mark.webbpsf` to mark those tests. This marker turns those tests off by default so that users can easily test parts of their code unrelated to WebbPSF, while providing them with the flag `--webpsf` to enable these tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,9 @@ docs = [
 Homepage = "http://webbpsf.readthedocs.io/"
 github_project = "https://github.com/spacetelescope/webbpsf"
 
+[project.entry-points]
+pytest11 = {webbpsf = 'pytest_plugin.plugin'}
+
 [tool.setuptools.packages]
 find = {namespaces = false}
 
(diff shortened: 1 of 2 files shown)


BradleySappington:
@WilliamJamieson, I can certainly understand the PyTest frustration leading to a plugin.  I think it may be more prudent to make the flag --IgnoreWebbpsf and use it to disable specific tests on personal runs.  All tests created should be run by default unless there is a reason not to.  Open for conversation!

BradleySappington:
@WilliamJamieson,  Doing some housecleaning, following up to see if the above proposition is amicable.

BradleySappington:
Closing for housekeeping.  Can re-open when available.

PR #766, opened 2023-11-06
update NRC coron dispersion coeffs

JarronL:
Update NIRCam coronagraphic wavelength-dependent tilt coefficients. These were miscalculated prior to launch and have since been producing incorrect PSF morphologiers. New values represent actual dispersion from CODEV and Zemax models.

LW example for F444W:
image

SW example for F200W:
image

### DIFF
--- a/webbpsf/optics.py
+++ b/webbpsf/optics.py
@@ -1503,13 +1503,23 @@ class NIRCamFieldAndWavelengthDependentAberration(WebbFieldDependentAberration):
         self.fm_long  = np.poly1d(lw_focus_cf)
 
         # Coronagraphic tilt (`ctilt`) offset model
-        # Primarily effects the LW channel (approximately a 0.031mm diff from 3.5um to 5.0um).
-        # SW module is small compared to LW, but we include it for completeness.
-        # Values have been determined using the Zernike offsets as reported in the 
-        # NIRCam Zemax models. The center reference positions will correspond to the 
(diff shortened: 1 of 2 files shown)


mperrin:
I confirm this is a necessary fix. I had noticed a while ago the dispersion issue in particular with wide filter coronagraphic PSFs, and brought this to Jarron's attention. Realizing now I had not made a GitHub issue for any of that. 

Briefly, for the record: 
 - Comparing SW PSFs in F200W, or LW in F356W, F444W showed clearly that the sim speckles did not match well, in particular for the angle or dispersion of the speckles at left and right sides. 
unknown_720
 - Multiplying by an ad-hoc scale factor of -10 resulted in a significantly better fit. 
unknown-5_720
 - For LW I also found discrepant results. Again the sign seems flipped, but  the amplitude seems too big instead of too small. Best fit scale factor is like -0.4 for F356W, -0.6 for F444W (very coarsely)

mperrin:
This PR will require updating the test function `tests/test_nircam.py/test_nircam_coron_wfe_offset()`

JarronL:
Done!

mperrin (approved):
Looks good to me. Nice improvement in PSFs for wide filters with coronagraphy.

mperrin:
@obi-wan76 FYI this is ready for merging.  If you'd like I can go ahead and do so myself, just let me know, but by default I'm assuming you and/or @BradleySappington are handling all the PR merges now.

PR #767, opened 2023-11-08
add aperture name support for NIRSpec slit and IFU apertures; also, 100x faster data cube calculations.

mperrin:
**PART 1:** The code for handling SIAF apertures was previously set up assuming all SIAF apertures map to specific detectors and pixel coordinates (as is the case for imaging mode). This is not the case for spectrograph apertures, for instance the NIRSpec slits and IFU. Previously, trying to set a NIRSpec aperture name to e.g. "NRS_FULL_IFU" or "NRS_S200A1_SLIT" would result in errors.  

This PR fixes that, such that slit and IFU aperture names may now be used.  The main benefit of this is getting the V2V3 coordinates correct for interpolation of field-dependent OPDs. (Lots of other questions and uncertainties in how well that works for NIRSpec, given the limited spatial sampling from CV3 and the complexities of the optical train. But it's a small step in the right direction of improved IFU and slit PSF models to have this bit of the code working.)

Some test code to demonstrate this works as intended. The following will fail on `develop`, and work for this branch with the fix.  (There's also a unit test included in the PR). 

```python
nrs = webbpsf.NIRSpec()
nrs.image_mask = None
print(nrs.aperturename, nrs._tel_coords())

nrs.set_position_from_aperture_name('NRS_FULL_IFU')
print(nrs.aperturename, nrs._tel_coords())

nrs.aperturename = 'NRS_S400A1_SLIT'
print(nrs.aperturename, nrs._tel_coords())
```

**PART 2:**  This now also adds a new method `calc_datacube_fast`, which implements a simplified streamlined version of `calc_datacube()`, and can return data cubes about 100x faster with accuracy consistent within ~0.01 or better with the slower version. (I.e. consistent well within the fidelity we currently have for any of the IFU modes)

### DIFF
--- a/webbpsf/tests/test_nirspec.py
+++ b/webbpsf/tests/test_nirspec.py
@@ -2,6 +2,8 @@ import sys, os
 import numpy as np
 import matplotlib.pyplot as plt
 import astropy.io.fits as fits
+import astropy.units as u
+import pysiaf
 
 import logging
 _log = logging.getLogger('test_webbpsf')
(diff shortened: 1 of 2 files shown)


mperrin:
@tracy-beck FYI about this PR I started a little while ago to improve handling of NIRSpec apertures. 

I had asked @obi-wan76 to review it via the GitHub interface, but that hasn't happened yet. Maybe you might be available to review this? 

In particular this fix I think will be useful for (eventually) having different wavefront error maps for the IFU slicer aperture versus otherwise.

marcio-melendez (commented):
Yes, this works as intended. 
```
NRS1_FULL [ 5.34389659 -7.9813583 ] arcmin
NRS_FULL_IFU [ 4.997258   -8.30428008] arcmin
NRS_S400A1_SLIT [ 5.3598465  -7.96845703] arcmin
```
There is an issue at line 1624 in webbpsf_core with the function calc_datacube_fast that has an extra indent space thus webbpsf failed when loading. I fixed the indent for that function and it ran with no problem. I haven't include this small fix to this PR.

marcio-melendez (approved):
I added the single indent space to this PR

BradleySappington:
@mperrin happy to merge as is.  After merge,  are you able to create a unit test for the new `calc_datacube_fast` method?  That will help with the code coverage hit this PR is taking.

mperrin:
Will do @BradleySappington (though maybe not until after the holiday next week...)

PR #768, opened 2023-11-13
combining issues 726 and 736

marcio-melendez:
This PR has two components:
1.  a small addition to   match_data in order to set the correct pupil in certain situations. For example, WL images. 
2. adding interpixel capacitance for both distortion extension in the simulated PSF, i.e., detector and oversample. For this, I moved the IPC effect from the gridded_library  into webbpsf_core where we handle the IPC for the detector sampled PSF as it was originally. 

Note that you can always check if IPC effects are included in your simulation by inspecting the header. 

I tested this PR by using match_data for both the simple webbpsf simulation and a gridded simulation. Sample code:

```
file = 'jw04502071001_03104_00001_nrca3_cal.fits'
inst = webbpsf.setup_sim_to_match_file(file_input) 
sim_psf = inst.calc_psf( fov_arcsec = 3.153900235) # fov to match the size of the gridded PSF

c = webbpsf.gridded_library.CreatePSFLibrary(inst,inst.filter,  detectors = inst.detector, num_psfs = 1, psf_location = inst.detector_position )
grid = c.create_grid() # GriddedPSFModel object 
```


mperrin:
@obi-wan76 did you see that the tests are failing on this PR? Specifically the function `test_ipc_oversampling_equivalence`. This was previously set up to test that you could either bin and then apply IPC, or apply IPC and then bin, and would get the same results in either order. Seems likely to me this is getting confused by the change to already have the IPC already added in the oversampled version, so it's getting applied twice in the test, or something like that.

marcio-melendez:
@mperrin thanks, I'll take a look at it when I come back next week.

mperrin:
@obi-wan76 I'm looking at this PR but don't see any additions to match_data? 

I just ran into a situation where I needed to improve that function to handle NIRCam bar coronagraph data keywords. I could add those lines of code to this PR, but I don't want to conflict with changes you may already have done?

marcio-melendez:
@mperrin I forgot to add the file to the push. It is showing now. It should handle coronagraph data keywords but please add any missing setup. Thanks!

mperrin:
Thanks that change does what I needed :-) 

(there may be an edge case still TBD for some of the dual channel coronagraph cases, using the SW channel with the LWB mask and so forth; but for right now those masks are treated interchangeably so this can be dealt with later)

mperrin:
I added one additional enhancement to this, relevant for NIRCam bar mask sims:The `match_data.setup_sim_to_match_file` function will set the `aperturename` attribute.  I added a few lines of code to use that aperture name value when looking up the offset along the bar occulter. This allows `setup_sim_to_match_file` to correctly setup a sim for cases in which the user used the Fiducial Point Override special requirement to override the default position on the coronagraph. (relevant for cycle 1 PIDs 1194 and 1412)

mperrin:
To handle dual-channel NIRCam coronagraphy data is tricky, since the FITS header APERNAME keyword cannot be relied upon to correctly give the aperture name used. (Known issue, discovered during commissioning of the dual channel operations mode, not yet prioritized to be addressed in PPS and DMS).  Thankfully @JarronL already wrote code to work around this in `webbpsf_ext`. With his permission I'm copying those files here into webbpsf so they can be used in`setup_sim_to_match_file`.

Example: 
```
> nrc = webbpsf.setup_sim_to_match_file('jw01194002001_0310a_00001_nrca4_uncal.fits')
> nrc.aperturename
'NRCA4_400X256_MASKLWB_NARROW'
```

marcio-melendez:
I fixed a problem with the previous commit that the IPC effects were apply twice inside the gridded PSF path and also to apply IPC effects before the especial normalization inside gridded PSF.  There is also a 
 new check in `detectors.py` so not to apply IPC effects twice in any of the `calc_psf` paths. I tested all the different combinations and they are all in agreement with previous results. 
 All distortion extension  of calc_psf should have IPC by default.

marcio-melendez:
@Skyhawk172 can you take a look at this PR? I just fixed a problem with my previous commit. Thanks!

Skyhawk172:
Ran a few cases along with the test mentioned above and everything runs successfully and as expected. Looks good!

Skyhawk172 (approved):

mperrin:
@obi-wan76 good morning. I've been reminded that this PR was approved but hasn't been merged yet, I guess perhaps because the tests were failing? Do you have a plan for looking into this, or would you like me to? 

Do you know, does the fix you made to `match_data` in this address the case of NIRCam using the filters in the pupil wheel, for instance F323N as reported in issue #806 ?

marcio-melendez:
@mperrin the test that's failing is `test_return_intermediates` with  `KeyError: 'MASKSWB'`. I get the following message:
`ValueError: Filter MASKSWB does not have a defined nominal offset position along MASKSWB`
I think the code should be picking the bar_offset from the filter but it is using MASKSWB and thus not finding the offset. 
If I do the offset position by hand, e.g., `nc.options['bar_offset']  = 0.5  ` then the test pass. However, this may be an indication of a larger configuration issue with MASKSWB.

marcio-melendez:
This PR also solves the problem from #806
For example:
```
file = './jw01059-o004_t029_nircam_f322w2-f323n_i2d.fits'
file_input = file
inst = webbpsf.setup_sim_to_match_file(file_input)

Setting up sim to match ./jw01059-o004_t029_nircam_f322w2-f323n_i2d.fits
iterating query, tdelta=3.0

MAST OPD query around UTC: 2022-03-14T18:59:16.344
                        MJD: 59652.79116138889

OPD immediately preceding the given datetime:
	URI:	 mast:JWST/product/R2022031405-NRCA3_FP1-1.fits
	Date (MJD):	 59652.6556
	Delta time:	 -0.1355 days

OPD immediately following the given datetime:
	URI:	 mast:JWST/product/R2022032002-NRCA3_FP1-0.fits
	Date (MJD):	 59655.2847
	Delta time:	 2.4936 days
User requested choosing OPD time closest in time to 2022-03-14T18:59:16.344, which is R2022031405-NRCA3_FP1-1.fits, delta time -0.136 days
Downloading URL https://mast.stsci.edu/api/v0.1/Download/file?uri=mast:JWST/product/R2022031405-NRCA3_FP1-1.fits to /Users/melendez/WFSC_Flight/webbpsf-data_1_2_1/webbpsf-data/MAST_JWST_WSS_OPDs/R2022031405-NRCA3_FP1-1.fits ... [Done]
Importing and format-converting OPD from /Users/melendez/WFSC_Flight/webbpsf-data_1_2_1/webbpsf-data/MAST_JWST_WSS_OPDs/R2022031405-NRCA3_FP1-1.fits
Backing out SI WFE and OTE field dependence at the WF sensing field point

Configured simulation instrument for:
    Instrument: NIRCam
    Filter: F323N
    Detector: NRCB5
    Apername: NRCB5_FULL
    Det. Pos.: (1024, 1024) 
    Image plane mask: None
    Pupil plane mask: None


psf = inst.calc_psf()

psf['DET_DIST'].header

PUPILINT= 'jwst_pupil_RevW_npix1024.fits.gz' / Pupil aperture intensity source  
PUPILOPD= 'OPD from supplied FITS HDUlist object' / Pupil OPD source            
OPD_FILE= 'R2022031405-NRCA3_FP1-1.fits' / Pupil OPD file name                  
OPDSLICE=                    0 / Pupil OPD slice number, if file is a datacube  
INSTRUME= 'NIRCam  '           / Instrument                                     
FILTER  = 'F323N   '           / Filter name                                    
EXTNAME = 'DET_DIST'           / This extension is oversampled.        

```

webbpsf/webbpsf_core.py
```diff
@@ -2252,13 +2256,22 @@ def _addAdditionalOptics(self, optsys, oversample=2):
...
             bar_offset = self.options.get('bar_offset', None)
-            # If the bar offset is not provided, use the filter name to lookup the default
+            # If the bar offset is not provided, use the SIAF aperture name, or else the filter name to lookup the default
             # position. If an offset is provided and is a floating point value, use that
             # directly as the offset. Otherwise assume it's a filter name and try passing
             # that in to the auto offset. (that allows for selecting the narrow position, or
             # for simulating using a given filter at some other filter's position.)
             if bar_offset is None:
-                auto_offset = self.filter
+                # Try to use the SIAF aperture name to determine the offset
```
marcio-melendez, line 2446: @mperrin I think this is the addition that's creating problems in test_return_intermediates. In the test,  auto_offset = maskswb because aperture name, 'NRCA4_FULL_MASKSWB' and that's not an option in offset_swb. In the cases of maskswb/masklwb, it should  be setting auto_offset = self.filter as before but I guess this breaks something for some observations when matching observations.  We could change the test by assigning the offset manually:

`filt = 'F182M'
nc = webbpsf_core.NIRCam()
nc.filter = filt
nc.image_mask='maskswb'
nc.pupil_mask='wedgelyot'
coron = webbpsf.optics.NIRCam_BandLimitedCoron('MASKSWB')
bar_offset = coron.get_bar_offset_from_siaf(filt, channel='SW')
nc.options['bar_offset']  = bar_offset

osys = nc.get_optical_system()

psf, intermediates = nc.calc_psf(monochromatic=2e-6, return_intermediates=True)
assert len(intermediates) == len(osys.planes)
assert isinstance(intermediates[0], poppy.Wavefront)
assert isinstance(psf, astropy.io.fits.HDUList)`

mperrin: I fixed the code to eliminate this bug. This test passes now. 

Changing the test code itself is rarely the right way to go, not as a first step. The starting assumption should be that a failing test is telling you there's some kind of bug or logic issue in the code. In the case, the test failure _correctly_ caught that the set of code I'd written here had not considered that sort of input. I wrote it expecting aperture names like "NRCA4_FULL_MASKSWB_F210M" but neglected the case of it being like "NRCA4_FULL_MASKSWB".

mperrin:
@obi-wan76 I made a fix to the PR code so that the test in `test_nircam` passes, from my work on the part of this that was for issue 726. 

But there's still several test failures from `test_psfgrid`, presumably from your work towards implementing 736. So those will still need to be addressed.

mperrin:
@obi-wan76 I re-added the two commits that got lost in your rebase. And did a little bit of code cleanup (whitespace, etc)  and one fairly minor revision to your distortion code: for the `add_ipc_gridded` variable that you added, since it was only needed internal to that one function it could just be a local variable. Does not need to be saved into the options dict for access anywhere else, as far as I could tell. Yes?

The tests all pass locally for me, but are failing on here still. This is pretty weird.

mperrin:
Wait, no, I take that back: actually the test suite DOES fail for me locally, reproducing the error seen online. Hmm. Will investigate.

PR #770, opened 2023-11-30
Infrastructure improvements for improved IFU sims

mperrin:
Work in progress for adding infrastructure for better IFU simulations, shared between NIRSpec and MIRI classes. 

This PR is heavily inspired by #691 by @patapisp, but has generalized much of the implementation to work across both NIRSpec and MIRI. 

This adds a new intermediate abstract subclass `JWInstrument_with_IFU`, and changes the NIRSpec and MIRI classes to  subclass from that. The benefit will be to allow adding IFU-specialized code shared between the two instruments, but not needed in the other three JW SIs. 

That shared code includes:

- [x] Addition of a `mode` attribute to indicated imaging or IFU major mode
- [x]  move the faster data cube function to this subclass
- [x] Add a `.band` attribute to specify IFU bandpass. The implementation and behavior differs for the two classes, following the instrument designs. For NIRSpec the `band` is like `"G295H/F290LP"`, derived from the `.disperser` and `filter` attributes. For MIRI the `band` is an MRS sub-band, like `3A`. 
   - [x] For NIRSpec, this necessitated also adding filter support for the long-pass filters. This requires an update to the data files, and the data files min version. 
- [x] For both classes, reasonable automatic support for toggling SIAF aperture name and mode, in either direction: setting `mode='IFU'` explicitly will change the aperturename to a relevant IFU aperture. Setting the aperture name to an IFU aperture will invoke IFU mode. The intent is to make it seamless and as automatic as possible for the user to set up a self-consistent calculation. 
- [x] For both classes, a lookup table attribute `._IFU_bands_cubepars`, derived from current CRDS cubepar files, which gives the wavelength range, wavelength sampling, and spaxel size for all bands. 
- [x] A method `get_IFU_wavelengths()` that returns the wavelength sampling for the selected band, based on the cubers values.  This function takes an optional `nlambda` argument to specify some desired sampling across the band; by default it matches the wave step from cubepars, which can be >1000 wavelengths depending on mode. 
- [x] Auto adjust pixel (spaxel) scale based on IFU/imaging mode, and IFU band (for MIRI)
- [x] For NIRSpec, in IFU mode, rotate images to match the IFUalign output orientation. 
- [ ] Do we need any such rotation for MIRI to better match the IFUalign outputs for MIRI too? 
- [ ] Consider for IFU mode updating the default pixel scale to match IFUalign output cubes? Or leave it at the actual detector pixel scale as the default? The user can always simply set the pixel scale manually to a custom value, e.g. `nrs.pixelscale=0.1` 

**Example code:**
_MIRI:_
```py
miri = webbpsf.MIRI()
miri.mode = 'IFU'
miri.band= '2A'
waves = miri.get_IFU_wavelengths()
cube = miri.calc_datacube_fast(waves)
```
_NIRSpec:_
```py
nrs = webbpsf.NIRSpec()
nrs.mode = 'IFU'
nrs.disperser = 'PRISM'
nrs.filter = 'CLEAR'
waves = nrs.get_IFU_wavelengths()
cube = nrs.calc_datacube_fast(waves)
```

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -58,7 +58,7 @@ jobs:
 
     - name: Get WebbPSF Data
       run: | # Get WebbPSF data files (just a subset of the full dataset!) and set up environment variable
-           wget https://stsci.box.com/shared/static/0ojjfg3cieqdpd18vl1bjnpe63r82dx8.gz -O /tmp/minimal-webbpsf-data.tar.gz
+           wget https://stsci.box.com/shared/static/gkbnn3jwcootr9riwv0rv1t8rpvxl3vr.gz -O /tmp/minimal-webbpsf-data.tar.gz
            tar -xzvf /tmp/minimal-webbpsf-data.tar.gz
            echo "WEBBPSF_PATH=${{github.workspace}}/webbpsf-data" >> $GITHUB_ENV
 
--- a/docs/relnotes.rst
+++ b/docs/relnotes.rst
@@ -27,12 +27,24 @@ See https://github.com/spacetelescope/webbpsf/issues for currently open issues a
 Version History and Change Log
 -------------------------------
 
+Version 1.3.0
+=============
+*Date TBD*
+
+This release comes with new features and improvements including but not limited to:
+
+X. Improved support for NIRSpec and MIRI IFU PSF calculations, including addition of a ``mode`` attribute for toggling between imaging mode and IFU mode simulations; an option for much faster (but slightly simplified) IFU datacube PSF calculations. Spectral bandpass information added for the IFU bands for both NIRSpec and MIRI. For NIRSpec, IFU mode PSF outputs are rotated by an additional 90 degrees to match the convention used in pipeline-output s3d datacubes made using the IFUAlign orientation. This extra rotation can be optionally disabled if so desired; see the NIRSpec class docstring. 
+
+
 Version 1.2.1
 =============
 Minor documentation updates
 
 Version 1.2.0
 =============
+
+*2023 August*
+
 We are pleased to announce the release of the latest version of WebbPSF version 1.2.0, now available on PyPi and GitHub. This release comes with new features and improvements including but not limited to:
 
 1. The addition of detector effects for JWST simulations. H2RG detector effects are included in two flavors, a simple ad hoc Gaussian convolution to model charge diffusion effects and a set of convolution kernels to model interpixel capacitance (IPC) and post-pixel coupling effects. We have found that these effects greatly improve the agreement between observations and simulations. See `JWST Detector Effects for more details. <https://webbpsf.readthedocs.io/en/latest/jwst_detector_effects.html>`_
--- a/webbpsf/__init__.py
+++ b/webbpsf/__init__.py
@@ -40,7 +40,7 @@ if sys.version_info < tuple(
 # required. If changes to the code and data mean WebbPSF won't work
 # properly with an old data package, increment this version number.
 # (It's checked against $WEBBPSF_DATA/version.txt)
-DATA_VERSION_MIN = (1, 2, 1)
+DATA_VERSION_MIN = (1, 3, 0)
 
 
 class Conf(_config.ConfigNamespace):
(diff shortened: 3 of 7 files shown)


mperrin:
FYI @patapisp @tracy-beck

mperrin:
Here's a general question for @patapisp @tracy-beck and anyone else thinking about IFU mode PSF characterization. IFU pixel scales are irregular, and in general not square. How should we approach this in sim PSFs for the IFU? I can imaging cases being made for trying to simulate on the native sampled spaxels, or on the pixel scale that the pipeline makes drizzled datacubes.  (in principle it seems better to simulate on the native detector sampling, but those are in general not square! Would need a bunch of effort to simulate different pixel scales in the alpha and beta directions.)

My hand waved suggestion is to, for now, simulate by default at the pipeline output pixel size (and, as usual, potentially oversampled relative to that scale). That at least gives us an isotropic pixel scale, avoiding the complexity of different X and Y scales. In principle we could then handle the different X and Y scales as a distortion effect, applied by interpolation after the optical propagation calculation. Thoughts?

patapisp:
@mperrin, regarding pixel sizes. My thoughts are as follows: 
- I would by default use the square pixel sizes that come out after cube building. This is to my understanding already working with the infrastructure you set up for the IFU mode. 
- I have concrete plans to enable MIRI IFU PSFs on the detector level which would then have the native detector sampling. Maybe we can use the same code for NIRSpec? I have not looked into the details, but I presume the distortion model might be similar. 

I am looking into the PR now to try to catch up to the new features you implemented! :)

mperrin:
Hi @patapisp - I wanted to check in since it's been a while since we chatted. Did you happen to have any time to work more on any code for MRS in webbpsf?
Right now, I think I am inclined to try to get this PR merged in with the current functionality (after de-conflicting with the changes from other PRs already merged). And then we can leave additional functionality to further improve MRS PSF models for some later PR.  If you have no objections I think that's what I will do?  (Past experience has taught me that it can be more trouble to pack too much stuff into one PR...) Thanks!

mperrin:
@obi-wan76 @tracy-beck the webbpsf IFU-mode PR is ready for review. 

I added the new v1.3.0 data files to Box which are needed for testing this, and the tests pass here on Github so that part seems good.   

I'd still like to add in some more user documentation of this, but I think that might best be left for a separate PR.

marcio-melendez:
The example code above works for me with the relevant files from the webbpsf-data 1.3. I tried to save the cube via the outfile option, `cube = miri.calc_datacube_fast(waves, outfile = 'test_cube.fits')`, however, it did not save it as a cube but a single frame with all the 4 webbpsf ext 
```
  0  OVERSAMP      1 PrimaryHDU     107   (284, 284)   float64   
  1  DET_SAMP      1 ImageHDU       108   (71, 71)   float64   
  2  OVERDIST      1 ImageHDU       112   (284, 284)   float64   
  3  DET_DIST      1 ImageHDU       113   (71, 71)   float64  
```

mperrin:
Nice catch on the outfile parameter not being implemented for that method. Added. Grab the latest and try again now, please.

marcio-melendez:
The save file  is working as intended. Thanks! 

Looking at some of the MIRI cubes, inside the fits file header, there are a few places that refers to the mean wavelength and the wavelength "0" that show a value of 2um, i.e., WAVELEN and WAVE0. Also the history shows that it was created with a wavefront at 2um. This seems to be very NIRSpec specific, so I am wondering if the header keywords are not being updated correctly for MIRI. I think it is only an issue with some of the header information because the wave parameter is set correctly for the MIRI calculation, from 7.51 to 8.7697um, and the keyword WVLN0000 is correct for both MIRI and NIRSpec.

mperrin:
Good catch with the wavelengths in the headers. I think actually there may indeed be some bug with that. I'll take a look!

mperrin:
I pushed some improvements to the wavelength handling in the calc_datacube_fast code, so (a) now it will use a more reasonable wavelength for MIRI calculations for the initial optical propagation, and (b) the redundant WAVE0 keyword is deleted, and instead there are WAVELN00, WAVELN01 etc. (consistent with the keywords from the regular calc_datacube output.)

mperrin:
Should be ready for re-test now

marcio-melendez (approved):
Great, it looks good now. Thanks!

PR #778, opened 2023-12-07
Coronagraph calcs: add 'coron_include_pre_lyot_plane' option for extra output plane

mperrin:
This is a relatively specialized PR for a technical use case in coronagraph calculations. This adds an option for, when using the return_intermediates option to calc_psf, optionally outputting an additional wavefront plane to return the wavefront in the Lyot plane right _before_ the Lyot stop. 

Example usage: 
```
# setup
nrc = webbpsf.NIRCam()
nrc.pupil_mask = 'MASKLWB'
nrc.image_mask = 'MASKLWB'
nrc.filter='F460M'

# new option
nrc.options['coron_include_pre_lyot_plane'] = True
psf, planes = nrc.calc_psf(nlambda=1, return_intermediates=True, display=True)
```
The output planes will include an extra plane compared to the default, giving the electric field before the Lyot stop. This is useful for visualizing and diagnosing certain aspects of coronagraphic starlight suppression, by visualizing the intensity before/after the Lyot stop: 

Unknown-5

Unknown-6

### DIFF
--- a/webbpsf/tests/test_nircam.py
+++ b/webbpsf/tests/test_nircam.py
@@ -592,3 +592,18 @@ def test_coron_shift(offset_npix_x=4, offset_npix_y=-3, plot=False):
     assert np.isclose(cutout_1.sum(), cutout_2.sum()), "PSF cutout sums should be consistent"
 
     assert np.allclose(cutout_1, cutout_2), "PSF cutouts should be consistent"
+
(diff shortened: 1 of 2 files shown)


mperrin:
FYi @lpueyo

marcio-melendez (approved):
This looks good. The test code works as expected. 

image

PR #779, opened 2023-12-08
Fix ci failures and update latest/min versions

BradleySappington:
Updating versions and tests to handle new photutils version 1.10.0


BradleySappington:
@obi-wan76 @mperrin @larrybradley -
Updated tests to handle new grid_xypos addition from photutils.  
Looks like everything is okay except for `test_psfgrid.py::test_saving`.

Attaching the output from the failed comparison that checks a saved file with what was originally written.
(test_array1.txt is from the file write)

[test_array1.txt](https://github.com/spacetelescope/webbpsf/files/13621241/test_array1.txt)
[test_array2.txt](https://github.com/spacetelescope/webbpsf/files/13621246/test_array2.txt)

BradleySappington:
@obi-wan76 @mperrin  -  When saving model, to maintain xygrid sorting, now writing using `model.data` instead of `psf_array`.  Please advise if there are any science related issues with this change in particular.

mperrin:
> @obi-wan76 @mperrin - When saving model, to maintain xygrid sorting, now writing using model.data instead of psf_array. Please advise if there are any science related issues with this change in particular.

I think what we really need is @larrybradley to advise on interface issues with the newest version of photutils and the save/read functionality there. I'm not familiar enough in detail with the changes made in x,y order to know if this would cause any problems or not. 

One small piece of advice though, is that if we're making changes in how the PSF grid files are saved, let's make sure that there's a version number keyword that's being saved into the files as metadata, so that when reading in any file it's 100% unambiguous which version of the software it was created with (and thus which x,y ordering). The science-related issue I can imagine would be reading in a saved file created with some software version X into code running version Y, and being confused about data ordering in such a way that the data in the PSF grid for field-dependence gets mixed up. Want to make sure that doesn't happen accidentally for any users!

mperrin (approved):
I have tested this PR and verified that
1. If you create a psf grid, and save it to disk then read back in, then the computed grid and read-from-disk-file grid match precisely (as expected)
2. The PSF in a given grid position matches the expected PSF for that detector location as given by the grid_xypos metadata. (I.e. the reordering of grid data in https://github.com/astropy/photutils/pull/1661 did not introduce any accidental surprises when making or reading grids using WebbPSF). 

Tested with photutils 1.10.0.

LGTM.

BradleySappington:
@mperrin , thank you for the review.  Would you still like me to add the version of webbpsf in the header information?

mperrin:
I checked and the version of WebbPSF is already in the header information!

mperrin:
@obi-wan76 would you also want to review this? I see @BradleySappington tagged both of us. Or one of us can go ahead and merge now. Your choice. I think it would be nice to ge this merged sooner rather than later so the CI is working again and can be used to test all the other PRs...

marcio-melendez (approved):
This looks good to me, specially after Marshall testing.

PR #781, opened 2023-12-22
Support MIRI LRS slit in aperturename and setup_sim_to_match_file functions

mperrin:
I encountered a bug when trying to use `setup_sim_to_match_file` with a MIRI LRS observation, using SIAF aperture name 'MIRIM_SLIT'. This is a SLIT-type SIAF aperture, which defines V2,V3 coordinates but does not specify a particular set of detector coordinates or subarray (it's read out using the MIRIM full array).  You can also reproduce this bug simply with:
```
webbpsf.MIRI().set_position_from_aperture_name('MIRIM_SLIT')
```
The necessary code for handling this kind of aperture was previously only present in the NIRSpec class, which has several slit apertures. I'd missed that MIRI SIAF also uses this for the LRS. 

This PR refactors the code from the `NIRSpec` class `aperturename` and `_tel_coords` functions, moving it up to the `JWInstrument` class and generalizing it to work on any slit aperture for any instrument.   

Overall this PR actually makes the code shorter and simpler (more deleted lines than added lines!), which is pretty satisfying.

### DIFF
--- a/webbpsf/tests/test_miri.py
+++ b/webbpsf/tests/test_miri.py
@@ -1,6 +1,7 @@
 import logging
 import os
-
+import astropy.units as u
+import pysiaf
 import numpy as np
 
 _log = logging.getLogger('test_webbpsf')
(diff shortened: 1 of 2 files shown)


marcio-melendez (approved):
Looks good. I reproduced the original bug
`TypeError                                 Traceback (most recent call last)
Cell In[15], line 1
----> 1 webbpsf.MIRI().set_position_from_aperture_name('MIRIM_SLIT')

File ~/webbpsf_issues_726_736/webbpsf/webbpsf_core.py:981, in JWInstrument.set_position_from_aperture_name(self, aperture_name)
    978 detname = aperture_name.split('_')[0]
    979 self.detector = detname  # As a side effect this auto reloads SIAF info, see detector.setter
--> 981 self.aperturename = aperture_name
    983 if self.name != 'NIRSpec' and ap.AperType != 'SLIT':
    984     # Regular imaging apertures, so we can just look up the reference coords directly
    985     self.detector_position = (ap.XSciRef, ap.YSciRef)  # set this AFTER the SIAF reload

File ~/webbpsf_issues_726_736/webbpsf/webbpsf_core.py:937, in JWInstrument.aperturename(self, value)
    935 self._aperturename = value
    936 # Update detector reference coordinates
--> 937 self.detector_position = (ap.XSciRef, ap.YSciRef)
    939 # Update DetectorGeometry class
    940 self._detector_geom_info = DetectorGeometry(self.siaf, self._aperturename)

File ~/webbpsf_issues_726_736/webbpsf/webbpsf_core.py:301, in SpaceTelescopeInstrument.detector_position(self, position)
    298 @detector_position.setter
    299 def detector_position(self, position):
    300     try:
--> 301         x, y = map(int, position)
    302     except ValueError:
    303         raise ValueError("Detector pixel coordinates must be pairs of nonnegative numbers, not {}".format(position))

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
`
This PR addressed this issue with no errors and the test functions returns True for both V2/V3 for `tel_coords`, that they are the same.

PR #784, opened 2024-01-08
fix some display issues in monthly_trending_plot

mperrin:
Some of the annotations in `monthly_trending_plot` were not showing as intended, in at least some versions of matplotlib. This fixes that. 

Two unrelated issues: the subfigure white backgrounds were hiding some figure annotations, and the connection patches had a hard coded Y value but should instead adapt to the y axis max of the time series plot. 

**Before**:  Note the absence of suptitle and connection patch lines. 

Unknown-1

**With this fix:** Those annotations appear as intended. 

Unknown


marcio-melendez (approved):
Looks good.

PR #786, opened 2024-01-10
add test for calc_datacube_fast, and fix some imports

mperrin:
Thanks to @JarronL for reporting a couple small issues in calc_datacube_fast. This fixes those.


JarronL:
Looks good to me!

JarronL (approved):
Looks good to me!

PR #787, opened 2024-01-10
improve MIRI LRS model details

mperrin:
Address spacetelescope/stpsf#41.

- [x]  Update LRS slit geometry for better consistency with SIAF
- [x] Fixes SI WFE model to be after coordinate rotation to detector frame for LRS mode. (I believe this fix also will slightly improve the model setup for MIRI coronagraphy too. )
- [x] Tunes up cruciform artifact amplitude in LRS mode, using an estimate from F1000W and F1130W.
- [x] Adjust the distortion code to gracefully skip trying to add distortion for LRS, rather than raising an exception.
- [x] Add a basic unit test for running an LRS mode PSF sim. 
 
Remaining open items from spacetelescope/stpsf#41:
- [x]  Pupil stop is correctly set to the oversized tricontagon for the P750L pupil mask, but ought to be set to that for any imaging filter used with the LRS. (or, strictly speaking, in general, but we don't model the effects of the oversized internal pupil stop for regular imaging mode).
- [x]  Need to update pupil shear for this mode using the values from commissioning. See [Wright, Sabatke, and Telfer 2022](https://www.spiedigitallibrary.org/conference-proceedings-of-spie/12180/2632087/James-Webb-Space-Telescope-MIRI-shear-pupil-analysis/10.1117/12.2632087.full#_=_)

Also:
 - [x] Major improvement in LRS slit PSF calculation fidelity through changing the optical propagation to use a much more finely sampled pixelscale for a small region around the LRS slit, computed via MFT. This makes a very significant difference in fine details of LRS PSF sims. 

Defer these as out-of-scope; could be done in some future PR:
- [ ]  Need to understand/double check relative about filter-dependent apparent offsets and how to account for them. In discussion with @skendrew . This may be outside of the scope of webbpsf but relevant for fitting physically-optics-based forward models to data for complex scenes seen via the LRS.
- [ ]  Wavelength-dependent continuous version of the cruciform convolution kernel needed. (Actually I now think this is out of scope for this PR)
- [ ]  SI field dependent WFE: can we get a better estimate of it near the LRS field point? Extrapolate from MIMF sensing near there?  _Edit:_ Descope from this PR. The current estimate is good enough for now.

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -1960,8 +1960,13 @@ class MIRI(JWInstrument):
             # Per Klaus Pontoppidan: The LRS slit is aligned with the detector x-axis, so that the
             # dispersion direction is along the y-axis.
             # Slit width and height values derived from SIAF PRDOPSSOC-063, 2024 January
+            # Undocumented options allow for offsetting the slit relative to the output pixel grid, to
+            # more precisely match the actual instrument alignment
             optsys.add_image(optic=poppy.RectangularFieldStop(width=4.72345, height=0.51525,
-                                                              rotation=self._rotation, name=self.image_mask))
+                                                              rotation=self._rotation, name=self.image_mask,
+                                                              shift_x=self.options.get('lrs_slit_offset_x', None),
+                                                              shift_y=self.options.get('lrs_slit_offset_y', None),
+                                                              ))
             trySAM = False
         else:
             optsys.add_image()
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -1823,8 +1823,10 @@ class MIRI(JWInstrument):
         self._rotation = 4.83544897  # V3IdlYAngle, Source: SIAF PRDOPSSOC-059
                                 # This is rotation counterclockwise; when summed with V3PA it will yield the Y axis PA on sky
 
-        self.options['pupil_shift_x'] = -0.0069 # CV3 on-orbit estimate (RPT028027) + OTIS delta from predicted (037134)
-        self.options['pupil_shift_y'] = -0.0027
+        # Coordinate system note: The pupil shifts get applied at the instrument pupil, which is an image of the OTE exit pupil
+        # and is thus flipped in Y relative to the V frame entrance pupil. Therefore flip sign of pupil_shift_y
+        self.options['pupil_shift_x'] = -0.0068  # In flight measurement. See Wright, Sabatke, Telfer 2022, Proc SPIE 
+        self.options['pupil_shift_y'] = -0.0110  # Sign intentionally flipped relative to that paper!! See note above.
 
         self.image_mask_list = ['FQPM1065', 'FQPM1140', 'FQPM1550', 'LYOT2300', 'LRS slit']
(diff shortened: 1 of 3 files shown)


mperrin:
In the course of some science analyses and discussions, I discovered/realized that the LRS calc has been (until now) using a very coarse pixel scale via default FFT propagation from pupil to image plane. This is not sufficiently precise. Much better outputs achieved via using an adjustable matrix Fourier transform to set much finer spatial scale right around the slit. 

Figures before and after: 
lrs_slit_sampling


As a bonus, this actually makes the code faster too since in this case we can restrict to computing a quite small region of the image plane. But that's not the main motivation.

marcio-melendez:
@mperrin how do I get the same image/output that you are showing in your comment above? (with the LRS slit). I checked your modifications and I was able to create a LRS PSF based on the test setup but I would like to play around a bit more, specially to get the same images. Thanks!

mperrin:
The above plots were obtained by using the `return_intermediates` option to calc_psf, to return the wavefront at each intermediate optical plane of the calculation. Like this:

```
psf, waves = miri.calc_psf(monochromatic=10e-6, fov_pixels=101, return_intermediates=True, display=True)
waves[2].display()
```

webbpsf/webbpsf_core.py
```diff
@@ -1822,8 +1823,10 @@ def __init__(self):
         self._rotation = 4.83544897  # V3IdlYAngle, Source: SIAF PRDOPSSOC-059
                                 # This is rotation counterclockwise; when summed with V3PA it will yield the Y axis PA on sky
 
-        self.options['pupil_shift_x'] = -0.0069 # CV3 on-orbit estimate (RPT028027) + OTIS delta from predicted (037134)
-        self.options['pupil_shift_y'] = -0.0027
+        # Coordinate system note: The pupil shifts get applied at the instrument pupil, which is an image of the OTE exit pupil
+        # and is thus flipped in Y relative to the V frame entrance pupil. Therefore flip sign of pupil_shift_y
+        self.options['pupil_shift_x'] = -0.0068  # In flight measurement. See Wright, Sabatke, Telfer 2022, Proc SPIE 
+        self.options['pupil_shift_y'] = -0.0110  # Sign intentionally flipped relative to that paper!! See note above.
```
marcio-melendez, line 1829: For my own clarification. In the paper the values are V2 = +0.0068 and V3 = -0.0110, this is the measured shear projected to the OTE Exit pupil. However, this is the relative shear from the primary mirror to the pupil and webbpsf needs the pupil shear. Therefore, we change the sign of both values, -0.0068 and 0.0110 and then change the y shift sign because it is applied at the instrument pupil. Thus, the final numbers, -0.0068 and -0.0110.

mperrin: 👍 . And you saw the emails with Randal asking him to help confirm I had the signs right here.

marcio-melendez: oh you are right, I forgot about the email. Yes, you explained everything there.

marcio-melendez (approved):
This looks good to me. I did the example from the test, the intermediate planes, and double check the values from the papers.

PR #789, opened 2024-01-20
Add `nrc_ta_image_comparison` function to trending tools

mperrin:
This adds a function `nrc_ta_image_comparison` which retrieves the TA image for a given visit, generates a matching simulation, and displays their difference. Useful as a check for binary. 

For example, in particular 4502:126...

Screenshot 2024-01-20 at 1 21 46 PM

### DIFF
--- a/webbpsf/mast_wss.py
+++ b/webbpsf/mast_wss.py
@@ -2,6 +2,7 @@
 
 
 import os
+import functools
 
 import astropy
 import astropy.io.fits as fits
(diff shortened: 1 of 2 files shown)


marcio-melendez (commented):
This is really good and it works as expected. I have a similar "residual" function that uses  `make_residual_image` from `PSFPhotometry` in `photutils.psf`.  I wonder if there is a way to quantify the residual  image beside the visual inspection (stdev?)

mperrin:
To quantify, a straightforward step would be to compute Chi^2, using the pipeline ERR extension estimate of the uncertainties.

mperrin:
Yeah that was pretty easy to add. Now it looks like this. Note annotation text in rightmost panel:

Unknown-7
Unknown-8

marcio-melendez (approved):
Looks really good, nice addition. Thanks!

Skyhawk172:
I ran this function on all of WFS&C observations and found that, sometimes, the chi^2 does not catch all binaries, if that's its main purpose. I'm looking into other options.

Screenshot 2024-01-30 at 12 14 11 PM

PR #794, opened 2024-02-01
NRC TA plot enhancements

mperrin:
Some minor enhancements to the `nrc_ta_image_comparison` function. 
- Adds an optional `show_centroid` parameter, which toggles a plot annotation showing the centroid location of the TA image. This includes a centroid measured using the fwcentroid code in poppy+webbpsf, and (optionally but automatically, if my `misc_jwst` toolkit is available, the onboard OSS centroids retrieved from the MAST engineering database).
- Generalizes some bits of this function, so it will work on any NIRCam TA exposure, not just the WFS program. This allows the same code to be used for similar analyses of coronagraphic or TSO target acquisitions with NIRCam. 

Probably fairly specialized use cases, but since I have this code now for myself it seems worthwhile to include it here to share.

Example with centroid over plotting:
 
Unknown-4

Example showing a NIRCam coron target acq: 
Unknown-3


marcio-melendez:
@Skyhawk172 since you have been using this function you may want to try the new minor additions.

webbpsf/trending.py
```diff
@@ -1641,30 +1638,61 @@ def nrc_ta_image_comparison(visitid):
...
+    scalefactor = np.nanmax(im_obs_clean) / im_sim.max()
+    if verbose:
+        print(f"Scale factor to match sim to data: {scalefactor}")
 
     im_sim_scaled_aligned = im_sim_shifted*scalefactor
 
     # Plot
+    if show_centroid:
+        # First, see if we can retrieve the on-board TA centroid measurment from the OSS engineering DB in MAST
+        try:
```
Skyhawk172, line 1908: To test it out, I installed `misc_jwst` but it crashed because I apparently haven't set up my `MAST_API_TOKEN` environment variable. There's a built-in exception handler in `engdb.get_ictm_event_log` that terminates the execution. So, until I set the env. variable, I'm now stuck and can't run the code because it will always go with `misc_jwst` since it's installed. I think in such cases, the code should naturally revert to calculating the centroid using webbpsf methods, instead of halting altogether. 

I think it'd be nice to be able to compare the two ways to measure the centroids; could the code calculate and report both centroids and (1) use the OSS if `misc_jwst` is successful  or (2) use the WebbPSF centroid otherwise?

mperrin:
ok, I can rework/improve the exception handling here. 

Yes, the intent of this part of the code is to compare the two ways to measure the centroids. It tries to get the OSS centroid, and the webbpsf centroid, and display and print both of them. As of Friday, in code not yet added to this PR, it also uses the WCS to retrieve where the star is "supposed to be" after the FGS acq, so it can also quantify directly what the pointing accuracy is after the FGS acq.

mperrin:
OK @Skyhawk172, please update to the latest misc_jwst (as of a minute ago), and retry. I think it should allow access to the eng db without requiring the API token to be set now.

Skyhawk172:
Hmmm, I'm getting a different error now:
```
File ~/opt/anaconda3/envs/py310/lib/python3.10/site-packages/misc_jwst/engdb.py:40, in get_ictm_event_log(startdate, enddate, mast_api_token, verbose, return_as_table)
---> 40 start = datetime.fromisoformat(f'{startdate}+00:00')
     41 if enddate is None:
     42     end = datetime.now(tz=tz_utc)

ValueError: Invalid isoformat string: '2024-01-14 18:55:49.4420000+00:00'
```

mperrin:
Oh, that parameter is just right now expecting startdate to be like "2024-01-14" with nothing after that. Can fix that too to be more general.

mperrin:
@obi-wan76 @obi-wan76 Wanted to ping you about this and ask if either of you will be able to review this PR, and if so by when?

marcio-melendez (commented):
@mperrin  when I ran it on one of our sensing programs I got:
`ValueError: Instrument NIRCam doesn't have a pupil mask called 'CLEAR'.`
This should be fixed in the PR #768 . Do you need to rebase this PR to get those changes?

mperrin:
yeah I'm working on rebasing and deconflicting now 👍

mperrin:
@obi-wan76 eh, looks like some other change broke things: that error shows up for the code in `develop` too now, not just this PR. I'll fix it in this branch in any case.

marcio-melendez (approved):
This is working for me now, looks good. 
Screen Shot 2024-04-03 at 10 39 57 AM

Issue #728, opened 2023-08-23
Efficiency improvement for webbpsf.trending.monthly_trending_plot()


bchen2:
when plotting using webbpsf.trending.monthly_trending_plot(2022, 3, opdtable=opdtable ,verbose=True), with a test opdtable of 50 rows ranging from March to May, the method processed all 50 rows without filtering. This is a request for efficiency improvement by filtering the provided opdtable by the year and month.

PR #801, opened 2024-02-12
Filter provided opd table by month

BradleySappington:
Fix for https://github.com/spacetelescope/webbpsf/issues/728
When providing an opdtable parameter to `monthly_trending_plot` do not assume it is already filtered for the correct month.

Filter accordingly before proceeding.

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -1362,6 +1362,8 @@ def filter_opdtable_for_month(year, mon, opdtable):
     opdtable = webbpsf.mast_wss.filter_opd_table(
         opdtable, start_time=pre_start_date, end_time=end_date
     )
+    if len(opdtable) == 0:
+        raise ValueError("The opdtable is empty for this date range.")
 
     # Trim the table to have 1 and only 1 precursor measurement -
     # we'll use this to compute the drift for the first WFS in the time period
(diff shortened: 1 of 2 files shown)


BradleySappington:
removed a lot of linting updates so the logic changes aren't lost in the jumble

webbpsf/trending.py
```diff
@@ -941,30 +1350,43 @@ def get_month_start_end(year, month):
     return start_date, end_date
```
BradleySappington, line 1352: filter_opdtable_for_month is new method to be used for auto generating, and filtering existing opdtable

webbpsf/trending.py
```diff
@@ -1038,24 +1471,31 @@ def monthly_trending_plot(year, month, verbose=True, instrument='NIRCam', filter
...
         if pid_dates is not None:
             pid_dates = pid_dates[(pid_dates >= start_date) & (pid_dates <= end_date)]
         else:
             pid = None
-        
-    if opdtable is None:
-        opdtable = get_opdtable_for_month(year, month)
-    else:
-        opdtable = check_colnames(opdtable)
+
```
BradleySappington, line 1487: logic change here to account for actual filtering if we pass our own opdtable, and also to handle errors

BradleySappington:
ruff formatting was too much, starting fresh with differen't PR for simplicity sake

PR #802, opened 2024-02-13
filter custom opbtable for month

### LINKED ISSUE #728: Efficiency improvement for webbpsf.trending.monthly_trending_plot()

BradleySappington:
Fix for https://github.com/spacetelescope/webbpsf/issues/728
When providing an opdtable parameter to monthly_trending_plot do not assume it is already filtered for the correct month.

Filter accordingly before proceeding.

### DIFF
--- a/webbpsf/tests/test_trending.py
+++ b/webbpsf/tests/test_trending.py
@@ -1,18 +1,7 @@
 import astropy
 import astropy.time
 import astropy.units as u
-import calendar
 import webbpsf
-
-def get_month_start_end(year, month):
-    _, ndays = calendar.monthrange(year, month)
-    start_date_str = f"{year:04d}-{month:02d}-01"
-    end_date_str = f"{year:04d}-{month:02d}-{ndays:02d}"
-
-    start_date = astropy.time.Time(start_date_str)
-    end_date = astropy.time.Time(end_date_str)
-
-    return start_date, end_date
 
 
 def test_monthly_trending_plot_auto_opdtable():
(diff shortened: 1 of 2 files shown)


BradleySappington:
I created a test_trending.py which has brought to light that there is an error with the latest astropy and how trending.py interacts with it.

webbpsf/tests/test_trending.py
```diff
@@ -0,0 +1,35 @@
...
+
+def get_month_start_end(year, month):
+    _, ndays = calendar.monthrange(year, month)
+    start_date_str = f"{year:04d}-{month:02d}-01"
+    end_date_str = f"{year:04d}-{month:02d}-{ndays:02d}"
+
+    start_date = astropy.time.Time(start_date_str)
+    end_date = astropy.time.Time(end_date_str)
+
+    return start_date, end_date
```
mperrin: This looks like an exact copy of the `get_month_start_end` function from webbpsf.trending? Why is it useful to duplicate that function here into the test code, instead of just calling the existing function?

mperrin (approved):
Looks good overall. 

My one question is about why the get_month_start_end function got duplicated into test code? It's a small thing but I would suggest it's simpler to call the existing function rather than duplicate the exact same code in two places.

webbpsf/tests/test_trending.py
BradleySappington: Good catch, it wasn't deliberate, I forgot to swap that out.  Thank you!

PR #804, opened 2024-02-15
consistently allow astropy.time.Time or string inputs to all trending functions

### LINKED ISSUE #790: Use consistent Time objects in trending.py
In `webbpsf/trending.py`, only the `wavefront_time_series_plot` function uses `datetime.datetime` objects for the `start_date` and `end_date`, while in other functions, `astropy.time.Time` objects are used. I think it makes sense to use `astropy.time.Time` object across the board.

mperrin:
Fixes #790 to make the API more consistent.  

Previously, some trending functions had a `start_date` parameter which wanted `astropy.time.Time` objects, and another had `start_date` which had to be a `datetime.datetime`. 

This PR fixes it so that all start_date and end_date input parameters can accept an `astropy.time.Time` object, or a string giving the date (which will be turned into an astropy.time.Time automatically)

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -42,8 +42,9 @@ def wavefront_time_series_plot(opdtable, start_date=None, end_date=None, ymin=0,
     ----------
     opdtable : astropy.table.Table
         OPD table, retrieved from MAST. See functons in mast_wss.py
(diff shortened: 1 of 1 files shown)


BradleySappington (approved):
LGTM

PR #807, opened 2024-02-20
Fix obscure error in FITS writing

eteq:
This one is a doozy insofar as I think it's really a corner case that triggers it. I have run a code using webbpsf for over a year without issues, and suddenly today it started giving me errors when writing out the final fits file.  I *believe* this is because in this specific configurations I have configured cupy to use my GPU, and some part of something that is being output is outputting a cupy scalar instead of a numpy scalar.  And `astropy.io.fits` is unhappy with that because it doesn't know what to do with cupy arrays (since they aren't supposed to auto-cast since they live in the GPU memory).

The fix, in this PR, is just to explicitly cast the values to be put into the FITS header into floats, rather than depending on the numpy implicit behavior.

I have no idea how to write a test for this given that it is pretty hardware-specific... But I can attest at least that on a g5.xlarge AWS instance (using NVIDIA A10G GPUs), this fixes the error...

### DIFF
--- a/webbpsf/webbpsf_core.py
+++ b/webbpsf/webbpsf_core.py
@@ -419,7 +419,7 @@ class SpaceTelescopeInstrument(poppy.instrument.Instrument):
         optsys.add_pupil(pupil_optic)
 
         pupil_rms_wfe_nm = np.sqrt(np.mean(pupil_optic.opd[pupil_optic.amplitude == 1] ** 2)) * 1e9
-        self._extra_keywords['TEL_WFE'] = (pupil_rms_wfe_nm, '[nm] Telescope pupil RMS wavefront error')
+        self._extra_keywords['TEL_WFE'] = (float(pupil_rms_wfe_nm), '[nm] Telescope pupil RMS wavefront error')
         if hasattr(pupil_optic, 'header_keywords'):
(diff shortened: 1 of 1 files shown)


mperrin:
ha ok. I am not entirely surprised -- the CI system doesn't include any tests on GPUs, and the testing of the CUDA code has in the past relied on my running the test suite manually on a Windows laptop, occasionally, so this is not exactly well tested, and it seems pretty plausible that something like this might slip through in a corner case. 
Not quite sure how/why that testing missed this corner case but maybe it's hardware-specific or cupy-specific in some regards. In any case the fix makes sense.

eteq:
yeah, if I understand right (and I may well not...) the cupy arrays sometimes behave more like numpy arrays depending on whether you are using its cpu mode vs the gpu mode, and depending on the details it'll sometimes swap between them.  It's all kinda magic to me but I *have* seen this kind of "behaves slightly differently depending on the hardware" situation with cupy in the past.

mperrin (approved):
Looks good to me.

mperrin:
By the way, glad to hear you're making use of the cupy GPU functionality! I wish it were easier to make use of the GPUs on our laptops, as opposed to all the extra startup overhead of getting a machine in the cloud with a GPU.

PR #817, opened 2024-03-14
[SCSB-145] require Python 3.10

zacharyburnett:
resolves [SCSB-145](https://jira.stsci.edu/browse/SCSB-145)

propagate Astropy's deprecation of Python 3.9 to downstream packages

> [!NOTE]
> this PR was generated automatically by ``batchpr`` :robot:

PR #819, opened 2024-03-21
Allow setup_sim_to_match_data to select choice of WFS before or after

mperrin:
Simple PR to add a keyword passthrough. This lets the  `choice='before'` or `'after'` parameter be used with the `setup_sim_to_match_data` function, and get passed to `load_wss_opd_by_date`. 

Also adds a missing docstring for setup_sim_to_match_data.

### DIFF
--- a/webbpsf/match_data.py
+++ b/webbpsf/match_data.py
@@ -8,8 +8,19 @@ import scipy.optimize
 import webbpsf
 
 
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Looks good,

mperrin:
The test failures are due to trying to run on py 3.9, now deprecated. That's fixed elsewhere (#817) so is not a concern for this PR. Going to merge anyway regardless of those failed tests.

PR #820, opened 2024-03-22
minor: improve X axis tick marks on trending timeseries plot

mperrin:
Very minor PR to avoid a too-crowded X axis. If showing a time series plot for > 1 year duration, make the major tick marks every 1 month instead of 1 week. 

Unknown-5


marcio-melendez (commented):
For the large events the labels overlap with the event itself and it is hard to read. I wonder if it is possible to add an offset for those events. I do agree that as we keep collecting data the X-axis started to look busier, so I like it now.

marcio-melendez (approved):
Looks good.

PR #825, opened 2024-04-01
Performance enhancement: avoid repeated slow loads of the SIAF

mperrin:
Benchmarking revealed that repeated loads of the SIAF information were a very significant overhead on PSF calculations: as much as _half a second_ per total calc_psf call, tested on a Mac M1 Max. Each load of the SIAF costs about 150 milliseconds, and there were (unintentionally) multiple reloads happening within the distortion code, and in some optics and OPD subclasses. This is unnecessary because the SIAF information will always be the same in any given python session (it only changes rarely, with new versions of pysiaf). 

This PR adds a cache function `get_siaf_with_caching` and replaces all JWST SIAF loads with a call to that, resulting in significant speedups: 

**Examples for a single PSF calculation:**

_Without this PR:_
```
%timeit psf = nrc.calc_psf(nlambda=1, fov_pixels=101)
691 ms ± 4.93 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

_With this PR:_
```
%timeit psf = nrc.calc_psf(nlambda=1, fov_pixels=101)
268 ms ± 3.02 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

**Examples for running the entire test suite:**

_Without this PR:_
```
> pytest   # run entire test suite
=== 115 passed, 1 xfailed in 303.20s (0:05:03) ===
```
_With this PR:_
```
> pytest   # run entire test suite
=== 115 passed, 1 xfailed in 240.77s (0:04:00) ===
```

```

### DIFF
--- a/webbpsf/distortion.py
+++ b/webbpsf/distortion.py
@@ -7,6 +7,7 @@ from scipy.interpolate import RegularGridInterpolator
 from scipy.ndimage import rotate
 
 from soc_roman_tools.siaf.siaf import RomanSiaf
+import webbpsf.webbpsf_core
 
 def _get_default_siaf(instrument, aper_name):
     """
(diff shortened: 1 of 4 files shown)


marcio-melendez (approved):
Looks good and it make sense.

PR #826, opened 2024-04-01
add delta_wfe_around_time function

mperrin:
@obi-wan76 This is the function I mentioned briefly in slack. Written a while ago but I hadn't yet had a chance to PR it. 

Given any date (or date and time, in the usual string format), this returns the delta OPD between the two measurements around that time.  This should make it relatively easy to make plots of the drifts corresponding to the times when you think you detect strain sensor steps. 

Screen Shot 2024-04-01 at 5 27 38 PM


marcio-melendez (approved):
This is great, thanks!

PR #827, opened 2024-04-05
add functions to download WFSC WL image data

mperrin:
Sometimes it's useful to download the weak lens image data. This PR adds to `mast_wss` some functions for doing so, either by specifying a WFSC program ID and observation number, or else by default just getting the most recent WFSC image data. The `download_wfsc_images` function retrieves the level-3 wfscmb data products from MAST and downloads into the current working directory. 

This is probably less useful to general users of webbpsf, and is intended mainly as a convenience for us on the mission team. But can also be useful for expert-level external users too, in some select cases. In any case I think this fits in well as part of the mast_wss submodule. 

(For right now, defaults to only downloading from NRCA3, but this is a user-selectable parameter in case data from one of the other detectors is wanted)

### DIFF
--- a/webbpsf/mast_wss.py
+++ b/webbpsf/mast_wss.py
@@ -656,16 +656,18 @@ def download_wfsc_images(program=None, obs=None, verbose=False, **kwargs):
 
     Other kwargs for detector and productlevel are passed through
     to the query functions.
+
+    Returns list of image filenames
     """
 
     if program is None and obs is None:
         if verbose:
             print("Querying latest available WFSC images")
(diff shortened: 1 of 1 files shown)


mperrin:
FYI to @grbrady. (This feature for webbpsf was partially inspired by wanting an quick tool to download more WL data, for testing out with stpr)

marcio-melendez (approved):
Looks good, this works as intended. I tested with NIRCam A3 and A1. For example: 
`webbpsf.mast_wss.download_wfsc_images(program = '4503', obs = '091', detector='NRCA`1')

Issue #839, opened 2024-05-06
Visit Id naming smarts in get_visit_nrc_ta_image


kulpster85:
This is a request to make _mast_wss.get_visit_nrc_ta_image()_  more flexible which assumes that _visitid_ has a leading "V" but doesn't check this assumption.

PR #840, opened 2024-05-06
#839 Visit Id naming smarts in get_visit_nrc_ta_image

kulpster85:
#839 - add robustness to the visitid formatting to replace a "V" if present and otherwise ignore


mperrin (approved):
Tested and working as intended. 

With this, both the following work equally well, yielding the same output without error.

```
webbpsf.trending.nrc_ta_image_comparison('V04503091001')
webbpsf.trending.nrc_ta_image_comparison('04503091001')

```

mperrin:
Fixed #839.

PR #841, opened 2024-05-08
Autofix linting, and formatting updates

BradleySappington:
Massive non functional ruff formatting/linting updates


BradleySappington:
@mperrin and @obi-wan76 - 
The changes in these files have all been completed using `ruff check . --autofix`
It should predominantly be formatting changes, and most of them are quote style changes.  You shouldn't see any "highlighted" changes that look like logic changes in your review, and if you do we should discuss (I will also personally review everything).

This is the first of the code clean PRS and should be pretty straight forward.

BradleySappington:
@mperrin and @obi-wan76 - I'm going to include one more commit with more auto-formatting changes for files that require manual intervention.  That way, when I do the manual intervention, there wont be formatting changes mucking up the compare.  I will tag you when ready for review (probably tomorrow)

BradleySappington:
@mperrin and @obi-wan76 - Ready for review.  Again, all of these changes were deemed the "safe" options for autoformat.  I didn't include any "unsafe" changes, that will be a future PR.  So this massive change set should be straightforward.

BradleySappington:
FYI: I just went through all the code and feel comfortable with this change set.

webbpsf/constants.py
```diff
@@ -32,132 +24,27 @@
...
-        [-1.53090954, -8.90571587e-07],
-        [-1.90872454, -.654384118]])),
-    ('C6-18', np.array([
-        [-1.8928951, 1.97644151],
-        [-1.51795694, 2.62917746],
-        [-0.76520012, 2.62751652],
-        [-0.38764887, 1.98019646],
-        [-0.76545554, 1.32580611],
-        [-1.52107554, 1.32581188]])),
+    ("A1-1", np.array([[-0.38101, 0.667604], [-0.758826, 1.321999], [-0.38101, 1.976407], [0.38101, 1.976407], [0.758826, 1.321999], [0.38101, 0.667604]])),
```
mperrin, line 27: I do not love this particular formatting change; these columns of numbers were intentionally arranged in two columns so you can see the X, Y values clearly.  But this is a tiny point and I'm going to let it go for this PR>

webbpsf/opds.py
```diff
@@ -95,8 +93,7 @@ class OPD(poppy.FITSOpticalElement):
 
     """
 
-    def __init__(self, name='unnamed OPD', opd=None, opd_index=0, transmission=None,
-                 segment_mask_file=None, npix=1024, **kwargs):
+    def __init__(self, name="unnamed OPD", opd=None, opd_index=0, transmission=None, segment_mask_file=None, npix=1024, **kwargs):
```
mperrin, line 96: it's super not obvious to me how ruff is deciding where to split lines or not. Compare this edit, where it compresses two lines into one, versus the next...

webbpsf/opds.py
```diff
@@ -271,67 +267,78 @@ def tvcircle(radius=1, xcen=0, ycen=0, center=None, **kwargs):
...
+        unit="nm",
+        pupil_orientation="entrance_pupil",
+        cbpad=None,
+        colorbar_orientation="vertical",
+        show_axes=False,
+        show_rms=True,
+        show_v2v3=False,
+        cmap=None,
+    ):
+        """Draw on screen the perturbed OPD
```
mperrin, line 341: ...here, where it has taken a def statement and expanded it out so that each parameter is on its own line. 

There must be some kind of logic for this, but it's not apparent to me?!

webbpsf/opds.py
```diff
@@ -444,63 +451,69 @@ def label_seg(self, segment, ax=None, show_axes=False, color='black', pupil_orie
...
+                    )  # in units of mm
 
                     xoffset = 0.1 if i == 2 else 0
-                    ax.text(cx + ax_arrow_len * b[0] * 1.5 + xoffset, (cy + ax_arrow_len * b[1] * 1.5)*ysign, label,
-                            color=color, fontsize=8,
-                            horizontalalignment='center', verticalalignment='center'
-                            )
+                    ax.text(
+                        cx + ax_arrow_len * b[0] * 1.5 + xoffset, (cy + ax_arrow_len * b[1] * 1.5) * ysign, label, color=color, fontsize=8, horizontalalignment="center", verticalalignment="center"
+                    )
```
mperrin, line 480: I also don't understand about the line length choices. As I've said, I'm not a fan of 80 character line limits, but here it's taken code and compressed it to one line, which as a result ends up being like 180 characters long. That seems a bit excessive... What do you think?   

Was there any specific line length limit value set when using ruff for this?

webbpsf/opds.py
```diff
@@ -1924,10 +1935,24 @@ def move_seg_local(self, segment, xtilt=0.0, ytilt=0.0, clocking=0.0, rot_unit='
...
+        xtrans=0.0,
+        ytrans=0.0,
+        piston=0.0,
+        roc=0.0,
+        trans_unit="micron",
+        display=False,
+        delay_update=False,
+        absolute=False,
+    ):
+        """Move a segment in pose and/or ROC, using PM global V coordinates..
```
mperrin, line 1955: And again this is inconsistent. This one gets expanded out to many lines, but the very next function is not.

webbpsf/opds.py
```diff
@@ -2061,10 +2088,8 @@ def move_seg_global(self, segment, xtilt=0.0, ytilt=0.0, clocking=0.0, rot_unit=
         if not delay_update:
             self.update_opd(display=display)
 
-    def move_sm_local(self, xtilt=0.0, ytilt=0.0, rot_unit='urad',
-                      xtrans=0.0, ytrans=0.0, piston=0.0, trans_unit='micron', display=False,
-                      delay_update=False):
-        """ Move the secondary mirror in pose, using segment-local control coordinates.
+    def move_sm_local(self, xtilt=0.0, ytilt=0.0, rot_unit="urad", xtrans=0.0, ytrans=0.0, piston=0.0, trans_unit="micron", display=False, delay_update=False):
+        """Move the secondary mirror in pose, using segment-local control coordinates.
```
mperrin, line 2092: These two functions, move_seg_local and move_sm_local, have almost exactly the same set of arguments, so it bugs me that ruff treats them inconsistently. Not sure I care enough to think we should change any of this further, but it just puzzles me about what the tool is doing.

mperrin (commented):
Only made it through about half the files, will try to look at the rest later. 

There's parts of the reformatting that I like, and parts I don't.... I think I will have to defer to you @BradleySappington on whether or not this cleanup actually makes the codebase better and more maintainable for the long run...

BradleySappington:
switching to new branch with different rules

PR #843, opened 2024-05-10
bug fix an edge case for switching MIRI apertures

mperrin:
Fix an edge case in some of the pixel scale checking code which happens when changing apertures. 

Currently, can't switch from a SLIT type aperture back to a regular aperture. The following will fail:
```
miri = webbpsf.MIRI()
miri.set_position_from_aperture_name("MIRIM_SLIT")
miri.set_position_from_aperture_name('MIRIM_FULL')
```

With this PR, the above works without error.


marcio-melendez (approved):
Look goods to me, it works without the warning/error

PR #844, opened 2024-05-10
Ruff changes

BradleySappington:
Initial Format only changes to establish "safe" formatting changes before doing any linting which will come in a subsequent PR


mperrin (approved):
I read through most of the auto-changed files, and all looked ok. Didn't see anything of major concern. And the tests are all passing which is the main reassurance. 

I'm going to mark this good and merge. Thanks @BradleySappington !

BradleySappington:
Thanks @mperrin!  I changed back any large lists where the auto-formatting looked like it was sabotaging intentional "non-customary" decisions.

PR #846, opened 2024-05-13
ruff "safe" auto-fix and format imports only

BradleySappington:
ruff "safe changes" auto formatting only.
Lint PR 2 of 3

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -1,5 +1,4 @@
 import calendar
-import datetime
 import functools
 import os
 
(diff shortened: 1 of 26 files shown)


webbpsf/trending.py
```diff
@@ -1,20 +1,19 @@
...
 import astropy.units as u
-from astroquery.mast import Observations
 import matplotlib
 import matplotlib.pyplot as plt
-from matplotlib.backends.backend_pdf import PdfPages
 import numpy as np
+import poppy
 import scipy.interpolate
+from astroquery.mast import Observations
+from matplotlib.backends.backend_pdf import PdfPages
```
mperrin, line 15: super minor comment/question: should this `from matplotlib` import be a couple lines higher up, grouped with the other matplotlib imports?

mperrin (approved):
Looks good to me!

webbpsf/trending.py
BradleySappington: I believe the best practice is all direct `import` statements are grouped above all `from ... import ....` statements

PR #847, opened 2024-05-14
Manual lint

BradleySappington:
Whole repo final linting procedure.  Marked with TODO for code review feedback on changes that needed a second glance or opinion.


dev_utils/package_filters.py
```diff
@@ -4,6 +4,8 @@
```
mperrin, line 4: This whole file package_filters.py can probably be dropped. Separate PR.

webbpsf/gridded_library.py
```diff
@@ -155,7 +155,7 @@ def __init__(
             from photutils.psf import GriddedPSFModel
         except ImportError:
             try:
-                from photutils import GriddedPSFModel
+                from photutils import GriddedPSFModel # noqa
             except ImportError:
                 raise ImportError('This method requires photutils >= 0.6')
```
mperrin: This nested import check can probably be simplified or dropped, since we're now requiring photutils >1.10 in pyproject.toml

webbpsf/jupyter_gui.py
```diff
@@ -36,7 +36,7 @@ def show_notebook_interface_wfi(instrument):
     from matplotlib import pyplot as plt
```
mperrin, line 36: jupyter_gui is another candidate file to just drop entirely. I can't remember if we formally deprecated it, but it's never been widely used, is not essential functionality, and I think should not be kept. Separate PR.

webbpsf/opds.py
```diff
@@ -261,7 +261,7 @@ def powerspectrum(self, max_cycles=50, sampling=5, vmax=100, iterate=False):
 
         """
 
-        # import SFT
+        import SFT
```
mperrin: This edit is incorrect; the commented-out import statement should just be deleted entirely. (SFT.py hasn't existed in this repo for many many years; that evolved into poppy's matrix_dft.py)

webbpsf/opds.py
```diff
@@ -2979,7 +2977,7 @@ def random_unstack(ote, radius=1, verbose=False):
 
 def test_OPDbender():
```
mperrin: These two test functions should be deleted. Better tests were long ago added in the tests subdirectory.

webbpsf/opds.py
```diff
@@ -3003,7 +3001,7 @@ def test_OPDbender():
 
 
 def test2_OPDbender(filename='OPD_RevV_nircam_132.fits'):
-    orig = OPDbender(filename)
+    orig = OPDbender(filename) # TODO remove this whole test?
```
mperrin: yep, remove

webbpsf/optical_budget.py
```diff
@@ -344,7 +346,7 @@ def visualize_wfe_budget(inst, slew_delta_time=14 * u.day, slew_case='EOL', ptt_
     ax.text(
         0.96,
         0.92,
-        f'45$^\circ$, {slew_delta_time}',
+        f'45$^\circ$, {slew_delta_time}', # noqa
```
mperrin: actually I think this should be fixed rather than noqa'ed. Using a `\\` rather than a single `\` works to escape the back slash properly. (Both single and double backslashes yield the desired output in plot text, which I just verified.)

webbpsf/optical_budget.py
```diff
@@ -359,7 +361,7 @@ def visualize_wfe_budget(inst, slew_delta_time=14 * u.day, slew_case='EOL', ptt_
         annotate_budget='Image motion (as equiv. WFE)',
         instname=inst.name,
     )
-    ax.text(0.04, 0.92, f'LOS jitter: {jitter}, 1$\sigma$/axis', transform=ax.transAxes)
+    ax.text(0.04, 0.92, f'LOS jitter: {jitter}, 1$\sigma$/axis', transform=ax.transAxes) # noqa sigma not escape char
```
mperrin: Same as above, use `\\` for a literal backslash.

webbpsf/optical_budget.py
```diff
@@ -425,8 +427,8 @@ def bounds(axis):
         )
 
     # Annotate first column to show WFE summation
-    between_row1_row2 = (bounds(axes[1, 0])[0, 1] + bounds(axes[2, 0])[1, 1]) / 2
-    between_row2_row3 = (bounds(axes[2, 0])[0, 1] + bounds(axes[3, 0])[1, 1]) / 2
+    between_row1_row2 = (bounds(axes[1, 0])[0, 1] + bounds(axes[2, 0])[1, 1]) / 2   # TODO - Unused values okay to delete?
+    between_row2_row3 = (bounds(axes[2, 0])[0, 1] + bounds(axes[3, 0])[1, 1]) / 2   # TODO - Unused values okay to delete?
```
mperrin: Yes OK to delete. I think this is leftover dev code fragments.

webbpsf/optics.py
```diff
@@ -233,8 +233,8 @@ def get_transmission(self, wave):
         The walls separating adjacent shutters are 0.06 arcsec wide.
         """
 
-        msa_width = 0.2
-        msa_height = 0.45
+        msa_width = 0.2     # TODO - Unused values okay to delete?
+        msa_height = 0.45   # TODO - Unused values okay to delete?
```
mperrin: OK to delete

webbpsf/optics.py
```diff
@@ -388,7 +388,7 @@ def __init__(
         if which == 'LLNL':
             raise NotImplementedError('Rotated field mask for LLNL grism not yet implemented!')
         elif which == 'Bach':
-            transmission = os.path.join(utils.get_webbpsf_data_path(), 'NIRISS/optics/MASKGR700XD.fits.gz')
+            transmission = os.path.join(utils.get_webbpsf_data_path(), 'NIRISS/optics/MASKGR700XD.fits.gz')  # TODO - Unused value, delete entire statement?
```
mperrin, line 389: Probably, but I'll defer to @obi-wan76 for this since he's going to be revising the GR700XD models at some point.

webbpsf/optics.py
```diff
@@ -841,7 +841,7 @@ def get_transmission(self, wave):
             if poppy.accel_math._USE_NUMEXPR:
                 import numexpr as ne
 
-                jn1 = scipy.special.j1(sigmar)
+                jn1 = scipy.special.j1(sigmar)  # noqa TODO - this looks like a bug, should jn1 be formatted in the following line?
```
mperrin: Not a bug. The following line uses `numexpr` which is a JIT compiler for fast optimized math. It's given a string input which is the expression to compile, then JITs it into byte code which accesses local variables by name.

webbpsf/roman.py
```diff
@@ -310,7 +309,7 @@ def calc_psf(
         self.options['crop_psf'] = crop_psf
 
         # add_distortion keyword is not implemented for RomanCoronagraph Class
-        if self.name == 'RomanCoronagraph' and add_distortion == True:
+        if self.name == 'RomanCoronagraph' and add_distortion is True:
```
mperrin, line 312: This could be further simplified I think to just `and add_distortion`, yes? The comparison to True can be implicit.

webbpsf/tests/test_opds.py
```diff
@@ -343,7 +343,8 @@ def test_apply_field_dependence_model():
     dependence. Thus there are several calls to manually set only the nominal field dep to True
 
     """
-    rms = lambda array, mask: np.sqrt((array[mask] ** 2).mean())
+    def rms(array, mask):
+        return np.sqrt((array[mask] ** 2).mean())
```
mperrin: This exact function is already defined in `utils.py` so we could just import it from there rather than defining a copy here.

webbpsf/tests/test_roman.py
```diff
@@ -185,23 +185,12 @@ def test_swapping_modes(wfi=None):
...
-        ],
-    ).format(detector_substr(self.detector))
-
     tests = [
         # [filter, mode, pupil_file]
-        ['F146', 'imaging', pupil_path],
-        ['F213', 'imaging', pupil_path],
-        [PRISM_FILTERS[0], 'prism', pupil_path],
-        [GRISM_FILTERS[0], 'grism', pupil_path],
+        ['F146', 'imaging', pupil_path(wfi)],
```
mperrin: I don't follow this edit. I see the lambda for pupil_path was deleted from the existing version, but I don't see a replacement implementation in the revised file?

webbpsf/tests/validate_webbpsf.py
```diff
@@ -99,7 +99,7 @@ def validate_vs_russ_plot7(base_opd='OPD_RevV_nircam_155.fits'):
     P.draw()
```
mperrin, line 99: This whole file validate_webbpsf is another one to delete. It's leftover testing from long ago in very early development of webbpsf. Separate PR.

webbpsf/trending.py
```diff
@@ -1380,7 +1376,7 @@ def basic_show_image(image, ax, vmax=0.3, nanmask=1):
             (ees_at_rad - median_ee) / median_ee,
             ls='-',
             color=color,
-            label=f'$\Delta$EE within {ee_rad:.2f} arcsec ({ee_npix} pix)',
+            label=f'$\Delta$EE within {ee_rad:.2f} arcsec ({ee_npix} pix)', # noqa
```
mperrin: I think this is another one to make use a double back slash instead of noqa

webbpsf/trending.py
```diff
@@ -1502,7 +1498,7 @@ def basic_show_image(image, ax, vmax=0.3, nanmask=1):
                 20, 20, f'{webbpsf.utils.rms(delta_opd, mask=apmask)*1e9:.1f}', color='yellow', fontsize=fs * 0.6
             )
 
-    for i, l in enumerate(['Measured\nWFE', 'Drifts', 'Mirror\nCorrections']):
+    for i, l in enumerate(['Measured\nWFE', 'Drifts', 'Mirror\nCorrections']): # noqa
```
mperrin: why does this need a noqa?

webbpsf/trending.py
```diff
@@ -1645,15 +1641,15 @@ def plot_wfs_obs_delta(fn1, fn2, vmax_fraction=1.0, download_opds=True):
...
+    wlp8_c1 = poppy.utils.pad_or_crop_to_shape(hdul1[11].data, wlm8_m1.shape)   # noqa TODO - any reason not to delete?
 
     opd, hdul2 = webbpsf.trending._read_opd(fn2)
 
     wlm8_m2 = hdul2[5].data
     wlp8_m2 = hdul2[10].data
-    wlm8_c2 = poppy.utils.pad_or_crop_to_shape(hdul2[6].data, wlm8_m2.shape)
-    wlp8_c2 = poppy.utils.pad_or_crop_to_shape(hdul2[11].data, wlm8_m2.shape)
+    wlm8_c2 = poppy.utils.pad_or_crop_to_shape(hdul2[6].data, wlm8_m2.shape)    # noqa TODO - any reason not to delete?
+    wlp8_c2 = poppy.utils.pad_or_crop_to_shape(hdul2[11].data, wlm8_m2.shape)   # noqa TODO - any reason not to delete?
```
mperrin: Concur with deleting all of the above 4 unused lines

webbpsf/trending.py
```diff
@@ -1897,7 +1897,7 @@ def show_wfs_during_program(
 
     # Look up wavefront sensing and mirror move corrections for that range
     opdtable = get_opdtable_for_daterange(start_date, end_date)
-    corrections_table = webbpsf.mast_wss.get_corrections(opdtable)
+    corrections_table = webbpsf.mast_wss.get_corrections(opdtable) # noqa TODO - this is unused and has no ramifications, is there a reason to keep or can I delete this line?
```
mperrin: Probably leftover from copy/paste when developing this function. I think OK to delete

webbpsf/trending.py
```diff
@@ -2017,7 +2017,7 @@ def delta_wfe_around_time(datetime, plot=True, ax=None, vmax=0.05, return_filena
         show_opd_image(delta_opd * nanmask, ax=ax, vmax=vmax)
         plt.colorbar(mappable=ax.images[0], label='WFE [microns]')
 
-        ax.set_title(f'$\Delta$WFE in the {post_delta_t - prev_delta_t:.2f} d around {datetime}')
+        ax.set_title(f'$\Delta$WFE in the {post_delta_t - prev_delta_t:.2f} d around {datetime}') # noqa
```
mperrin: double backslash

webbpsf/webbpsf_core.py
```diff
@@ -1745,6 +1745,17 @@ def load_wss_opd_by_date(self, date=None, choice='closest', verbose=True, plot=F
         opd_fn = webbpsf.mast_wss.get_opd_at_time(date, verbose=verbose, choice=choice, **kwargs)
         self.load_wss_opd(opd_fn, verbose=verbose, plot=plot, **kwargs)
 
+    def _label_wl (nwavelengths, wavelength_slices):
```
mperrin: If you're defining this as class method, it needs a `self` at the start. But since this doesn't depend on `self`, it might be better placed as a standalone function in utils rather than as part of the class definition.

webbpsf/webbpsf_core.py
```diff
@@ -2671,11 +2673,11 @@ def _addAdditionalOptics(self, optsys, oversample=2):
...
-        # NIRCam as-built weak lenses, from WSS config file, PRDOPSFLT-027
-        WLP4_diversity = 8.3443  # microns
-        WLP8_diversity = 16.5932  # microns
-        WLM8_diversity = -16.5593  # microns
-        WL_wavelength = 2.12  # microns
+        # NIRCam as-built weak lenses, from WSS config file, PRDOPSFLT-027 # TODO - Okay to delete these unused defines?
+        # WLP4_diversity = 8.3443  # microns
+        # WLP8_diversity = 16.5932  # microns
+        # WLM8_diversity = -16.5593  # microns
+        # WL_wavelength = 2.12  # microns
```
mperrin: Yeah, I think these values were used in a prior version of the code but are not currently used, so OK to delete.

mperrin (commented):
Comments and suggestions on TODOs

webbpsf/gridded_library.py
BradleySappington: moving single import from photutils.psf to top

webbpsf/opds.py
BradleySappington: @mperrin `def tvcircle` uses SFT.  Should that entire method be removed?

webbpsf/optical_budget.py
BradleySappington: I thought that `\circ` created a special symbol.  You are saying that `\\circ` yields the same result? If so, happy to change all these #noqa as there are plenty of them

webbpsf/optical_budget.py
BradleySappington: expanding, I thought these were latex specific symbols.

webbpsf/opds.py
mperrin: There are clearly some functions here which haven't been used or called in forever, in particular the `powerspectrum`function . Not included in the test suite, since clearly it doesn't work given that it's calling a file that doesn't exist.

webbpsf/optical_budget.py
mperrin: This is subtle about which piece of code is parsing the string. 

`\circ` în _LaTeX_ (invoked via Matplotlib) does yield the special symbol. But in _python string parsing_ by the python interpreter, for f strings any backslash gets interpreted as indicating an escape sequence, so it checks for `\c` which isn't a valid escape sequence, then falls back to pass that string to LaTeX . Passing the python interpreter the double backslash results in the double backslash being parsed to a single backslash, then a single backslash being passed to LaTeX via Matplotlib, directly, without first trying to interpret `\c` as an escape sequence.

webbpsf/roman.py
BradleySappington: Thats correct, I kept it intentionally with this check for code consistency.

webbpsf/tests/test_roman.py
BradleySappington: the lambda was a redefine of `def pupil_path` on line 22

webbpsf/trending.py
BradleySappington: the 'l' is ambiguous naming convention

mperrin: ah, yeah. I guess fine to leave as noqa but that's actually a valid critique from the linting. Could refactor that variable name to `label` to avoid the ambiguous single-letter-l.

webbpsf/webbpsf_core.py
BradleySappington: yup, realized that right away, have it fixed locally but will transfer it to utils.

webbpsf/optical_budget.py
BradleySappington: Awesome, I haven't used latex and was researching it, I appreciate the insight.

BradleySappington:
@pep8speaks suggest diff

mperrin:
@BradleySappington I think this all looks good, except now there's a few conflicts with edits from other PRs. I'm going to go ahead and resolve those for you...

mperrin (approved):
I approve, based on my earlier review;  all the comments from that were addressed in subsequent edits in this PR.

mperrin:
Ah but I see I broke something in my de-conflicting with other recent edits. Will fix...

BradleySappington:
Thanks for handling the merging @mperrin!

PR #848, opened 2024-05-14
Update for WFI to use pysiaf instead of soc_roman_tools

Skyhawk172:
The Roman SIAF is now available from PySIAF. This PR updates the source code that handles WFI distortion to pull the necessary data from PySIAF instead of `soc_roman_tools`.  I've done a few tests and code runs fins and results are as expected.

Also, as a result of this update, there is no more dependency on `soc_roman_tools`, and it was removed from "pyproject.toml".


BradleySappington (approved):
LGTM,
thanks for removing the dependency with this PR!

Issue #660, opened 2023-04-26
Actual detector values not respected by WebbPSF


dlakaplan:
I am trying to use WebbPSF on NIRCam LW data.  The FITS file says that the detector is `NRCBLONG` but WebbPSF will only accept `NRCB5`.  It seems that those should be compatible.

(Please forgive me if I should have asked this through the JWST helpdesk - I wasn't sure if this was a code or telescope question)

marcio-melendez:
Yes. We'll make a note to update the filter list so it can take A5/B5 and NRCBLONG/NRCALONG in order to match flight expectations. 
Thanks!

JarronL:
Any example function I wrote that parses a number of the detector name aliases and spits out what webbpsf expects. For instances, you could feed it one of A5, NRCA5, ALONG, NRCALONG, or 485 and it will return NRCA5.

```python

def get_detname(det_id, use_long=False):
    """Return NRC[A-B][1-4,5/LONG] for valid detector/SCA IDs
    
    Parameters
    ==========
    det_id : int or str
        Detector ID, either integer SCA ID or string detector name.
    use_long : bool
        For longwave detectors, return 'LONG' instead of '5' in detector name.
    """

    # For longwave devices, do we use 'LONG' or '5'?
    long_str_use = 'LONG' if use_long else '5'
    long_str_not = '5' if use_long else 'LONG'


    det_dict = {481:'A1', 482:'A2', 483:'A3', 484:'A4', 485:f'A{long_str_use}',
                486:'B1', 487:'B2', 488:'B3', 489:'B4', 490:f'B{long_str_use}'}
    scaids = det_dict.keys()
    detids = det_dict.values()
    detnames = ['NRC' + idval for idval in detids]

    # If already valid, then return
    if det_id in detnames:
        return det_id
    elif det_id in scaids:
        detname = 'NRC' + det_dict[det_id]
    elif det_id.upper() in detids:
        detname = 'NRC' + det_id.upper()
    else:
        detname = det_id

    # If NRCA5 or NRCB5, change '5' to 'LONG' (or vice-versa)
    detname = detname.upper()
    if long_str_not in detname:
        detname = detname.replace(long_str_not, long_str_use)
        # Ensure NRC is prepended
        if detname[0:3]!='NRC':
            detname = 'NRC' + detname

    if detname not in detnames:
        all_names = ', '.join(detnames)
        err_str = f"Invalid detector: {detname} \n\tValid names are: {all_names}"
[...]

PR #849, opened 2024-05-15
Allow specifying NIRCam LW detectors equivalently like 'NRCA5' or 'NRCALONG'

mperrin:
Issue #660 has been pending and slipped through the cracks, not dealt with for over a year now, about specifying NRCALONG as a synonym for NRCA5 and similarly for B. This is a quick easy fix so I just went ahead and fixed it for this release as well..


marcio-melendez (approved):
Looks good.

PR #850, opened 2024-05-16
Couple small fixes to LRS support

mperrin:
Fixes two small issues missed in recent PR spacetelescope/stpsf#41.

 - makes setup_sim_to_match_file better at inferring LRS slit for the specific case of LRS TACONFIRM exposures
 - Fixes a bug which would happen if you set the aperture name to MIRIM_SLIT then tried to set it back to some regular aperture such as MIRIM_FULL, due to the SLIT type apertures in pysiaf not having a defined pixescale. 

For testing, an example MIRI LRS TACONFIRM exposure is `jw01529001001_03102_00001_mirimage_cal.fits`.


marcio-melendez (approved):
Looks good, works as intended.

PR #851, opened 2024-05-17
minor bug fix to IFU mode after #770

mperrin:
Minor fix: When switching to NIRSpec IFU mode, don't leave the image_mask at a slit or MSA mask (which is impossible to see with the IFU). 

Specifically the default `image_mask` For NIRSpec is 'MSA all open', and setting `nrs.mode = 'IFU'` would leave `nrs.image_mask == 'MSA all open'` which makes no sense. This PR fixes that. 

It also deletes some very-very-old and obsolete warning checks about filter names starting with IFU and saying the IFU modes are not yet implemented; this is way out of date and no longer needed (and these lines of code were never getting executed anyway).

### DIFF
--- a/webbpsf/tests/test_nirspec.py
+++ b/webbpsf/tests/test_nirspec.py
@@ -55,6 +55,7 @@ def test_mode_switch():
     nrs.mode = 'IFU'
     assert 'IFU' in nrs.aperturename
     assert nrs.band == 'PRISM/CLEAR'
+    assert nrs.image_mask is None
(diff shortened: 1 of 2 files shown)


mperrin:
@obi-wan76 apologies for keeping making yet more PRs, but this one at least is a tiny quick bug fix...

marcio-melendez (approved):
Looks good to me, I did try it and it changed the mask to None.

PR #856, opened 2024-05-21
Additional Pep8speaks updates

BradleySappington:
Found a few missed items from the no-formatting transition


.pep8speaks.yml
```diff
@@ -12,7 +12,7 @@ message:  # Customize the comment made by the bot
     no_errors: "There are no PEP8 issues in this Pull Request."
 
 scanner:
-    diff_only: False  # If True, errors caused by only the patch are shown
+    diff_only: True  # If True, errors caused by only the patch are shown
```
mperrin, line 15: 👏

dev_utils/field_dependence/read_codev_dat_fits.py
```diff
@@ -45,12 +45,13 @@ def main():
...
-            'nirspec': 15,
-            'niriss': 15
-            }
+    order = {
+        'fgs': 15,
+        'nircam': 15,
+        'miri': 15,
+        'nirspec': 15,
+        'niriss': 15
+    }
```
mperrin, line 54: (minor) I'm curious what PEP8 rule is relevant for this? What's the style rule or reasoning for preferring the revised version with the one extra newline?   (to be clear I'm fine with this edit; I'm just asking for my own better understanding of why this is a pep8 change)

webbpsf/detectors.py
```diff
@@ -167,7 +169,7 @@ def apply_detector_ipc(psf_hdulist, extname='DET_DIST'):
 
     """
 
-    # In cases for which the user has asked for the IPC to be applied to a not-present extension, we have nothing to add this to
+    # Cases for which user has asked for the IPC to be applied to a not-present extension, we have nothing to add this to
```
mperrin: minor, but please change the text back here. Dropping those two words results in a non-sentence which is not clear.

webbpsf/mast_wss.py
```diff
@@ -103,12 +103,14 @@ def mast_wss_opds_around_date_query(date, verbose=True):
         if tdelta >= 6 * u.day:
             if verbose:
                 print(
-                    'Could not find JWST OPDs both before and after the specified date. Date outside of the available range of WFS data.'
+                    'Could not find JWST OPDs both before and after the specified date. \
+                        Date outside of the available range of WFS data.'
```
mperrin: Doesn't breaking the string this way result in a bunch of extra added whitespace? I don't think the whitespace at the start of the second line is stripped.  Test running this and the next edit below in both cases do result in strings containing a bunch of whitespace between the two parts. 

Consider splitting this just with two strings, like the following.  

```
                        'Could not find JWST OPDs both before and after the specified date. '
                        'Date outside of the available range of WFS data.'
```

webbpsf/opds.py
```diff
@@ -851,7 +851,7 @@ def _move(self, segment, type='tilt', vector=None, display=False, return_zernike
         if type == 'tilt':
             local_vector[
                 2
-            ] /= 1000  # convert Z tilt to milliradians instead of microradians because that is what the sensitivity tables use
+            ] /= 1000  # convert Z tilt to milliradians instead of microradians, that is what the sensitivity tables use
```
mperrin: minor: This edit results in a grammatically incorrect run-on sentence. I'm not sure what the point of this edit is?

webbpsf/webbpsf_core.py
```diff
@@ -2013,10 +2031,10 @@ def __init__(self):
         # This is rotation counterclockwise; when summed with V3PA it will yield the Y axis PA on sky
 
         # Modes and default SIAF apertures for each
-        self._modes_list = {'imaging': 'MIRIM_FULL',
-                            'IFU': 'MIRIFU_CHANNEL1A'}
+        self._modes_list = {'imaging': 'MIRIM_FULL', 'IFU': 'MIRIFU_CHANNEL1A'}
```
mperrin: This is another case where I'm curious what the reasoning is for the edit. I get that too-long lines should be avoided, but is there some PEP8 reason why it's frowned upon to occasionally add line breaks for clarity in a list or dict literal?   Is this an auto-reformatting that got overzealous or is there some deeper principle why it's not desirable?  I routinely see code with dict literals that has each key as its own line, because that often makes it clearer and easier to read and understand later.

webbpsf/webbpsf_core.py
```diff
@@ -977,21 +979,25 @@ def aperturename(self, value):
...
                     )
-            elif ap.AperType == 'COMPOUND' and self.name=='MIRI':
+            elif ap.AperType == 'COMPOUND' and self.name == 'MIRI':
                 # For MIRI, many of the relevant IFU apertures are of COMPOUND type.
                 has_custom_pixelscale = False  # custom scales not supported for MIRI IFU (yet?)
                 # Unlike NIRSpec, there simply do not exist full-detector SIAF apertures for the MIRI IFU detectors
-                _log.info(f'Aperture {value} is of type COMPOUND for MIRI; There do not exist corresponding SIAF apertures, so we ignore setting detector geometry.')
+                _log.info(
+                    f'Aperture {value} is of type COMPOUND for MIRI; \
+                      There do not exist corresponding SIAF apertures, so we ignore setting detector geometry.'
```
mperrin: another instance of line continuation needing revision

webbpsf/webbpsf_core.py
```diff
@@ -1025,7 +1035,8 @@ def aperturename(self, value):
                 if not has_custom_pixelscale:
                     self.pixelscale = self._get_pixelscale_from_apername(self._aperturename)
                     _log.debug(
-                        f'Pixelscale updated to {self.pixelscale} based on average X+Y SciScale at SIAF aperture {self._aperturename}'
+                        f'Pixelscale updated to {self.pixelscale} \
+                          based on average X+Y SciScale at SIAF aperture {self._aperturename}'
```
mperrin: here too

webbpsf/webbpsf_core.py
```diff
@@ -1842,9 +1860,12 @@ def calc_datacube_fast(self, wavelengths, compare_methods=False, outfile=None, *
         MIN_REF_WAVE = 2e-6  # This must not be too short, to avoid phase wrapping for the C3 bump
         if ref_wave < MIN_REF_WAVE:
             ref_wave = MIN_REF_WAVE
-            _log.info(f"Performing initial propagation at minimum wavelength {MIN_REF_WAVE*1e6:.2f} microns; minimum set to avoid phase wrap of segment C3 surface.")
+            _log.info(
+                f'Performing initial propagation at minimum wavelength {MIN_REF_WAVE*1e6:.2f} microns; \
+                  minimum set to avoid phase wrap of segment C3 surface.'
```
mperrin: also here

webbpsf/webbpsf_core.py
```diff
@@ -2642,7 +2663,8 @@ def aperturename(self, value):
                 # Set alternative aperture name as bandaid to continue
                 value = newval
                 _log.warning(
-                    'Possibly running an old version of pysiaf missing some NIRCam apertures. Continuing with old aperture names.'
+                    'Possibly running an old version of pysiaf missing some NIRCam apertures. \
+                     Continuing with old aperture names.'
```
mperrin: also here

webbpsf/webbpsf_core.py
```diff
@@ -2676,7 +2699,8 @@ def aperturename(self, value):
             if not has_custom_pixelscale:
                 self.pixelscale = self._get_pixelscale_from_apername(self._aperturename)
                 _log.debug(
-                    f'Pixelscale updated to {self.pixelscale} based on average X+Y SciScale at SIAF aperture {self._aperturename}'
+                    f'Pixelscale updated to {self.pixelscale} \
+                      based on average X+Y SciScale at SIAF aperture {self._aperturename}'
```
mperrin: also here

webbpsf/webbpsf_core.py
```diff
@@ -3247,26 +3281,31 @@ def aperturename(self, value):
...
                 self._detector_geom_info = DetectorGeometry(self.siaf, self._aperturename)
-                _log.info(f"{self.name} SIAF aperture name updated to {self._aperturename}")
+                _log.info(f'{self.name} SIAF aperture name updated to {self._aperturename}')
 
                 if not has_custom_pixelscale:
                     self.pixelscale = self._get_pixelscale_from_apername(detector_apername)
-                    _log.debug(f"Pixelscale updated to {self.pixelscale} based on average X+Y SciScale at SIAF aperture {self._aperturename}")
+                    _log.debug(
+                        f'Pixelscale updated to {self.pixelscale} \
+                          based on average X+Y SciScale at SIAF aperture {self._aperturename}'
```
mperrin: also here

webbpsf/webbpsf_core.py
```diff
@@ -3247,26 +3281,31 @@ def aperturename(self, value):
...
                 if 'IFU' in self.aperturename:
                     self._mode = 'IFU'
                     if self._disperser is None:
                         self.disperser = 'PRISM'  # Set some default spectral mode
                         self.filter = 'CLEAR'
                     if self.image_mask not in ['IFU', None]:
-                        _log.info("The currently-selected image mask (slit) is not compatible with IFU mode. Setting image_mask=None")
+                        _log.info(
+                            'The currently-selected image mask (slit) is not compatible with IFU mode. \
+                             Setting image_mask=None'
```
mperrin: also here

webbpsf/opds.py
```diff
@@ -3289,7 +3290,9 @@ def sur_to_opd(sur_filename, ignore_missing=False, npix=256):
     if not os.path.exists(sur_filename):
         if not ignore_missing:
             raise FileNotFoundError(
-                f'Missing SUR: {sur_filename}. Download of these should eventually be automated; for now, manually retrieve from WSSTAS at https://wsstas.stsci.edu/wsstas/staticPage/showContent/RecentSURs?primary=master.png'
+                f'Missing SUR: {sur_filename}. Download of these should eventually be automated; \
+                  for now, manually retrieve from WSSTAS at \
+                  https://wsstas.stsci.edu/wsstas/staticPage/showContent/RecentSURs?primary=master.png'
```
mperrin: line continuation needs revision

webbpsf/opds.py
```diff
@@ -1747,7 +1749,10 @@ def _get_zernikes_for_ote_field_dep(
             if clip_dist > 0.1 * u.arcsec:
                 # warn the user we're making an adjustment here (but no need to do so if the distance is trivially small)
                 warnings.warn(
-                    f'For (V2,V3) = {v2v3}, Field point {x_field_pt}, {y_field_pt} not within valid region for field dependence model of OTE WFE for {instrument}: {min_x_field}-{max_x_field}, {min_y_field}-{max_y_field}. Clipping to closest available valid location, {clip_dist} away from the requested coordinates.'
+                    f'For (V2,V3) = {v2v3}, Field point {x_field_pt}, {y_field_pt} \
+                      not within valid region for field dependence model of OTE WFE for \
+                      {instrument}: {min_x_field}-{max_x_field}, {min_y_field}-{max_y_field}. \
+                      Clipping to closest available valid location, {clip_dist} away from the requested coordinates.'
```
mperrin: line continuation needs revision

mperrin (changes requested):
Hi @BradleySappington thanks for these edits.  Most looks fine but I have a couple concerns, particularly about line breaks in strings. 

I think using a backslash to break a string across two lines doesn't work in the way you're intending it to here: the whitespace at the start of the second line isn't trimmed out automatically. I tested several of those edits, and the proposed edits do result in extra unintended whitespace in the output strings, so I think those will need to be revised. There are quite a few instances of this, as marked below (I may not have caught them all). 

This is a small thing but I don't think we want to accidentally introduce formatting errors in output messages in the course of cleaning up formatting errors of the code 😆

dev_utils/field_dependence/read_codev_dat_fits.py
BradleySappington: This is in the "[indentation](https://peps.python.org/pep-0008/#indentation)" section of the Pep8 standards.  Both `hanging` and `aligned` are fine, however the rest of this document uses `aligned` with its closure indicators, this was the only `hanging` instance that I saw.

webbpsf/detectors.py
BradleySappington: No problem!

webbpsf/mast_wss.py
BradleySappington: I forgot to go back to verify this, thanks for the catch!

webbpsf/opds.py
BradleySappington: To keep comment under 125 on one line.  I'll just move to leading line.  Thanks!

webbpsf/webbpsf_core.py
BradleySappington: This was an auto-format addition.  For the few larger files in this codeset that had a lot of pep8 issues, I used the auto-formatting and then went through and fixed the items that I deemed too much.  This was short enough that I let it stick, but happy to put back as I personally also prefer line delineations for clarity.

BradleySappington (commented):
Thanks again for pointing out the formatting mistake!

BradleySappington:
@mperrin - updates complete

mperrin (approved):
Looks good. Thanks @BradleySappington

PR #845, opened 2024-05-12
Implement support for NIRCam DHS sub apertures

mperrin:
Implements support for computing monochromatic PSFs for the NIRCam Dispersed Hartmann Sensor (DHS) mode. Intended for ETC PSF generation in support of implementing a TSO mode using the DHS.

Requires updated data files to work.

PR #857, opened 2024-05-21
add test function for the NIRCam DHS support.

mperrin:
Add a unit test for the DHS support added in  #845

This tests that we can select one of the DHS sub apertures as a pupil mask, make a PSF, and that the resulting PSF has only a couple percent of the total intensity (as expected, given the small size of these sub apertures)


marcio-melendez (approved):
It works as intended. I did the test for all the DHS apertures and I see that DHS1 and DHS6 have throughputs < 0.03

PR #870, opened 2024-06-11
Improve trending plots, particularly wfe_histogram arrows display

mperrin:
Improves the `trending.wfe_histogram_plot` arrow markers to show corrections. 
 - switches to use `matplotlib.annotate` to draw the arrows, instead of separate calls to plot lines and draw triangle markers. This ultimately is cleaner but ended up taking a while to figure out since there's a ridiculously complicated set of options to the arrow-drawing functions.
 - Adds `min_wfe` and `max_wfe` options to that function, which can be used to directly set custom min and max ranges if the defaults aren't suitable in some particular case. The histogram and cumulative histogram display is updates to clip values, such that values outside of min and max are still included in the histogram at the extreme bottom or top bin of the histogram, respectively. 

Example output:

Unknown-12

Separately, this PR also has a minor fix to some unrelated display issues elsewhere in `trending.py`, by setting `clip_on=True` in a couple function calls, to prevent labels being drawn outside of axes sometimes depending on x and y ranges there. This is mostly unrelated but is convenient to combine into this same PR.

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -201,7 +201,7 @@ def wavefront_time_series_plot(
             # Limit events that happened after start_date:
             if d >= start_date and d <= end_date:
                 plt.axvline(d.plot_date, color=color, ls=':', alpha=0.5)
-                ax.text(d.plot_date + 0.25, ymax * 0.95, event, color=color, rotation=90, verticalalignment='top', alpha=0.7)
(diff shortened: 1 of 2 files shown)


kulpster85:
Looks great!  Did some testing with the various options and it works well with changing the min,max values as well as the correction markers.

kulpster85 (approved):

kulpster85 (approved):

PR #878, opened 2024-07-08
minor: trending function add cycle 3 routine WFS PIDs

mperrin:
Very minor PR: Adds the cycle 3 WFS program IDs to the list inside `wavefront_time_series_plot`wavefront_time_series_plot`. This has only the trivial effect of changing the color of the dot used to plot those datapoints.

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -125,6 +125,7 @@ def wavefront_time_series_plot(
         4508,
         4509,
     ]  # Cycle 2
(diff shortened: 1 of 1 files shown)


mperrin:
@nflagey-stsci I'm wondering if you could step in to provide a Github review and approval of this fairly simple and trivial PR, please, since Marcio is away this week?

marcio-melendez (approved):
This looks good. We should think of a way to add the relevant sensing programs without hard coding their PIDs.

PR #879, opened 2024-07-08
more strict units handling; fixes some issues for astropy 6.0.0 compatibility

mperrin:
Fixes problem with `calc_datacube_fast` as reported in Slack.


marcio-melendez (approved):
This fixed the previous error: `UnitConversionError: Can only apply 'less' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan).`

PR #880, opened 2024-07-11
Remove unused modules

AlexKurek:


mperrin:
Hi, thanks for the contribution. However, I believe we would prefer to retain these various files. The `dev_utils` subfolder is exactly as it says, development utilities and scripts which are from time to time useful to the development team. For example the notebook for getting throughputs from Pandeia is used to produce updated instrument throughput files to be included in the webbpsf data zip files.  That script and the plot SI field dependence notebooks certainly should be retained.

PR #881, opened 2024-07-16
[CI] add names to workflows and fix data version parsing with `echo` and `awk`

zacharyburnett:
the workflows don't have names so it defaults to their filename, which is ugly:
[image](https://github.com/spacetelescope/webbpsf/actions)

Additionally, the `version` output was broken because I forgot to include `echo` when piping to `awk`; that's fixed here


mperrin (approved):
Looks good to me

PR #883, opened 2024-07-16
[CI] set a generalized cache path, allowing multiple data caches to be combined downstream

zacharyburnett:
since GitHub caching has a quirk where caches are distinguished by both the cache key AND cache path, this should make it much easier to cache WebbPSF data and also CRDS files when running the `romancal` tests, by setting `restore-keys`

This PR changes the cache path as so:
```diff
- ${{ runner.temp }}/webbpsf-data
+ ${{ runner.temp }}/data/
```
where `webbpsf-data/` is stored within `${{ runner.temp }}/data/`

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -64,11 +64,11 @@ jobs:
     - name: Get WebbPSF Data
       uses: actions/cache/restore@v4
       with:
-        path: ${{ runner.temp }}/webbpsf-data
+        path: ${{ needs.retrieve_cache.outputs.cache_path }}
         key: ${{ needs.retrieve_cache.outputs.cache_key }}
 
     - name: Set WebbPSF data path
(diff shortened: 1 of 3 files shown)


zacharyburnett:
@mperrin I apologize for all the notifications; when working on the `romancal` tests I realized that a more generalized cache path was needed. This and #881 should be the last two polishing PRs and then it should be maintenance-free. Thanks for being so patient!

mperrin:
@zacharyburnett I definitely don't understand the GitHub caching and cache paths at all (I was completely unaware of this functionality before you contributed this PR, to be clear) so it's easy to be patient with lots of PRs. Plus I know how much of a hassle it can be to get working these sorts of infrastructure things which can't be entirely tested locally!

mperrin (approved):

zacharyburnett:
Thanks! This just needs a manual run of https://github.com/spacetelescope/webbpsf/actions/workflows/download_data.yml to re-make the cache, and then it's all set

EDIT: one more PR to fix the `download_data.yml`: #885

PR #889, opened 2024-07-22
Fix #888, typo in NIRSpec disperser list

### LINKED ISSUE #888: A typo in NIRSpec.disperser_list
Hi there,

I find that the G395M is not accepted when it is set to `disperser` of class `NIRSpec`. It is possibly due to a typo at the line 3171 of [webbpsf_core.py](https://github.com/spacetelescope/webbpsf/blob/develop/webbpsf/webbpsf_core.py), which wrongly spells the name as G394M. I'm not sure if this typo would caused any further error. Please check, thanks.

mperrin:
Fixes #888


BradleySappington (approved):
Challenging review, but I pulled through

Issue #868, opened 2024-06-04
Check and perhaps slightly update NIRSpec aperture PAs


mperrin:
Tiny effect, only makes a difference  by < 1 degree, but nonzero. 

NRS1_FULL: 139.23
NRS2_FULL: 137.80
NRS_FULL_IFU: 138.97

The current hard-coded value in webbpsf is the average of the NRS1_FULL and NRS2_FULL values. We have the existing code infrastructure from other classes to update based on APERNAME and make that more precise (slightly).

PR #897, opened 2024-08-02
Update NIRSpec aperture PAs

mperrin:
This addresses issue #868 by updating the NIRSpec aperture PAs from PySIAF, including the very slightly different aperture PAs for NRS1 vs NRS2 and for different apertures such as the IFU or slit apertures. Tiny effect, fraction of a degree, so will have only a small effect on PSFs at most. But it's easy to do so we might as well. (Much simpler than the corresponding code for MIRI which I already wrote in #770).


mperrin:
This could go into either 1.3.1 or be saved for later. @obi-wan76 your choice, just let me know, or tag the Milestone on this PR, please.

marcio-melendez (approved):
It looks good to me, I ran the  test and run the function `_get_aperture_rotation`, they work as expected.

PR #899, opened 2024-08-02
Docs improvements, including for matching PSFs to data

mperrin:
Documentation improvements. 

- [x] Improve docs about detector_position and source_offset, and how they relate. Addresses spacetelescope/stpsf#39, and several recent user questions.
- [x] Update psf_grid docs. Addresses spacetelescope/webbpsf#887. 
- [x] Great big bunch of docs improvements to the main "Using WebbPSF" page, including converting it from .rst to .ipynb format, ensuring all code examples are runnable, and expanding/improving doc text extensively. 
- [x] Switch sphinx style file and some conf.py options to match those used by the JWST pipeline readthedocs, improving the  aesthetics a bit.

Issue #887, opened 2024-07-17
Using PSF Grids - webbpsf.gridded_library.display_psf_grid(grid) does not provide correct labels


ecoughli-trinity:
When following the provided sample code from [https://github.com/spacetelescope/webbpsf/blob/develop/docs/psf_grids.rst](url),  the resulting output is not correctly labeled and instead labels every image as 'None'.  Definitely really minor, but not sure how to rectify.

mperrin:
Hi @ecoughli-trinity, thanks for reporting this bug. I can reproduce the issue here with the unhelpful None output labels. 

The good news is, the functionality in webbpsf's `display_psf_grid` was migrated into `photutils` last fall, and the version in photutils is working correctly.  (I think the  cause of the problem with the plot output from `display_psf_grid`  is due to some internal changes in the  recent versions of photutils, which that function in webbpsf was not updated for). 

I recommend you simply use the photutils `plot_grid` function instead, like this: 

```
grid = nrc.psf_grid(num_psfs=36, all_detectors=False)
grid.plot_grid()
grid.plot_grid(deltas=True)
```

It seems to me like we should probably deprecate the function in webbpsf, and certainly should update the docs to just point users towards that photutils function call instead. @larrybradley @BradleySappington I presume you concur?

BradleySappington:
@mperrin - agreed that this is the best course of action

larrybradley:
@mperrin - yep I agree.

PR #901, opened 2024-08-05
mark display_psf_grid as deprecated

mperrin:
This addresses #887 by marking `display_psf_grid` as deprecated, using an astropy decorator for marking deprecation. 

I'm not actually sure this is the best way to do this, and welcome suggestions if there's a better way. I see there's also a PEP 702 to add a similar `@deprecated` to the standard library but it's not yet available even in py 3.12, so that doesn't work.   In any case we can informally deprecate (and I'm separately revising the docs and example code in #899 already), so this PR is just attempting to mark the deprecation in code, in some suitable manner.

### DIFF
--- a/webbpsf/gridded_library.py
+++ b/webbpsf/gridded_library.py
@@ -1,6 +1,7 @@
 import itertools
 import os
 from collections import OrderedDict
+import astropy.utils.decorators
 
 import astropy.convolution
 import numpy as np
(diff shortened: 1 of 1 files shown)


BradleySappington (commented):
@mperrin - this is a working solution, however the warning looks like an astropy method is what is being deprecated.  This may be confusing to the user (or it may not?)

I don't think we should included a deprecated library install for something like this, so we can either keep as you proposed, or just construct our own warning as so:

`
import warnings`

`
warnings.simplefilter('always', DeprecationWarning)`


`
def display_psf_grid((grid, zoom_in=True, figsize=(14, 12), scale_range=1e-4, diff_scale_range=1, cmap=None):`

`
    warnings.warn(
        "The display_psf_grid function is deprecated and may be removed in a future version."
        "Use photutils.psf.GriddedPSFModel.plot_grid instead.",
        DeprecationWarning,
        stacklevel=2
    )`

BradleySappington:
@mperrin , I updated the dependencies (and raised an issue), this one should be good to go.

PR #908, opened 2024-09-03
Enhance capabilities to simulate PSFs with larger JWST pupil 

marcio-melendez:
This PR is for enhance capabilities to simulate PSFs with larger JWST pupils. The changes are only in the load_wss functions in order to match the observed OPD to the larger pupil sizes. This PR also adds documentation for this enhance support.


docs/jwst_large_psf.ipynb
```diff
@@ -220,18 +220,54 @@
...
     "- jwst_pupil_RevW_npix4096.fits.gz\n",
     "- jwst_pupil_RevW_npix16384.fits.gz\n",
     "\n",
-    "In this case, npix refers to the number of pixels scaled from the same pupil side length (~ 6.6 meters). "
+    "In each case, npix refers to the number of pixels used to sample across the same pupil side length (~ 6.6 meters). \n",
+    "\n",
+    "To perform a large PSF simulation we need to provide the name of one of the larger size JWST pupil files via the `pupil` attribute of the instrument. \n",
+    "\n",
+    "Then, as before, we use the function `load_wss_opd_by_date` or, if you know the appropriate OPD, `load_wss_opd`. These functions will take the size of the user-specified JWST pupil and scaled the observed wavefront measurement (OPD) accordingly."
+   ]
```
marcio-melendez: @mperrin There is a new cell in the notebook (cell 5) that seems more like a "test" cell , in order to check the pupil parameters, but not necessarily intended for the documentation, is this correct? Unless we add some text, e.g.,  " Before changing the pupil file we can check the default values" , or something like that. 

`nrc = webbpsf.NIRCam()
print(nrc.pupil)

webbpsf.optics._get_initial_pupil_sampling(nrc), nrc._datapath`

marcio-melendez (commented):
Thanks for reviewing and making the changes.

docs/jwst_large_psf.ipynb
mperrin: yeah good catch, I'll take that back out later today, or else you can

mperrin: I had to debug and make a fix in a file elsewhere in webbpsf to get the notebook to run correctly. In my tests, it was erroneously looking in the wrong directory for pupil files specified with a relative path. I'm not sure how this worked for you before, since it didn't work for me... In any case that is test code and can come out

mperrin: I removed that extra cell, and pushed the revised notebook

marcio-melendez:
@mperrin  thanks for making the changes, it looks good. I cannot approve it because I am the author but you can go ahead and approve it. 
btw, you are right about the OPD path, I changed the file locally but did not add it to this PR, thanks for fixing my mistake!

mperrin (approved):
Tested, minor fixes applied as needed, all good now.

PR #909, opened 2024-09-04
using z range in requirements.txt

BradleySappington:
same as in poppy, lets allow all z updates for dependencies.


mperrin (approved):
Also looks good to me

Issue #666, opened 2023-05-03
Roman Cycle 9 Reference Data file corrupted


Skyhawk172:
While looking more closely at the Zernike aberrations, I realized that the `"wim_zernike_cycle9"` data file used by WebbPSF does not match the original Cycle 9 file delivered by Goddard. See screenshot.  It appears something went wrong when the original file got manually parsed for use by WebbPSF.  @ojustino is currently investigating the root cause of how this happened.

![image (1)](https://user-images.githubusercontent.com/9323123/236033913-8358f468-8568-44da-967f-f6d27a2cefa3.png)

I did a quick test by properly parsing the original file and the results are as expected (i.e. identical to what Goddard reports), so clearly the `"wim_zernike_cycle9"` file currently used by WebbPSF is wrong but the fix is straightforward.

ojustino:
The root cause here was a mistake in copying values from Goddard's original spreadsheet for Roman's imaging mode to `wim_zernike_cycle9_OG.csv`. I remade it and ran it through the steps in `Process_zern_and_filters.ipynb` to generate a new `wim_zernike_cycle9.csv` file. @Skyhawk172 and I verified that the new file's Zernike values align with the spreadsheet's.

I also noticed that the aforementioned files, `wsm_grism_zernikes_cycle9.csv`, and `wsm_prism_zernikes_cycle9.csv` all had values from their corresponding spreadsheets' "RMS (nm)" columns where the CSV files' headers actually call for values from the "PV (waves)" columns instead. I remade all of these files using the proper columns.

These changes are now reflected in the ITAR storage location's `webbpsf-data-source/WFI` directory. I've also added a `cycle9` directory to `sources` there so we have a mutually accessible record of the updates we made for Cycle 9 in #466.

PR #910, opened 2024-09-11
Roman docs & figures updates

Skyhawk172:
I updated the figures under the "Roman Instrument Model Details" section of the documentation as it appeared they had not been updated with the latest optical model (see GIF). The `how_tos.py` was also modified in order to rearrange the first figure.

I also updated the caption to one of the figures to clearly state that the prism and grism PSFs shown are monochromatic.

Finally, I updated the Roman notebook, in particular by adding how to use the `datacube()` function for the spectroscopic "filters".

Note: there are no differences in the Roman CGI figures after re-running the code to generate the figure, so no update needed.

POST-FACTO EDIT: The root cause of the difference was tracked down to the fix for https://github.com/spacetelescope/webbpsf/issues/666, which was rolled out as part of v1.2.0.

RomanWebbPSF


marcio-melendez (approved):
Looks good, I tested the notebook and confirmed that the output is as expected. thanks!

PR #911, opened 2024-09-11
add CI tests for mast_wss, in test_mast_wss.py

mperrin:
In the course of testing #908 I wanted to add a unit test to test that new code in the CI. And I realized I had long ago wrote -- but never committed, by accident -- a `test_mast_wss.py` file. 

This PR adds that file, including with an added test for the OPD rescaling feature added in #908. As such, we should expect the CI tests for this PR to fail until #908 is merged, and then subsequently pass after that.

### DIFF
--- a/webbpsf/tests/test_mast_wss.py
+++ b/webbpsf/tests/test_mast_wss.py
@@ -1,0 +1,57 @@
+import pytest
+import webbpsf
+from astropy.table import Table
+
+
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Looks good, thanks for adding the test for the OPD rescaling. I'll approve now and merge once the CI tests pass, after merging the large scale PR.

PR #912, opened 2024-09-13
Pre release 140

BradleySappington:
All release specific updates for the 1.4.0


BradleySappington:
Due to cache issues, merging with failure assuming that once merged, all tests will pass.  Will re-assess before doing release

PR #914, opened 2024-09-13
[CI] allow manual run of `download_data.yml` to specify data URL

zacharyburnett:
adds the `url` input to the `Run Workflow` box (auto-populated with the minimal data URL by default). This way, if a new data cache is needed, simply run `download_data.yml` (https://github.com/spacetelescope/webbpsf/actions/workflows/download_data.yml) on the default branch (`develop`) and enter the new data URL

### DIFF
--- a/.github/workflows/download_data.yml
+++ b/.github/workflows/download_data.yml
@@ -27,8 +27,13 @@ name: download and cache data
 on:
   workflow_call:
     inputs:
+      url:
+        description: URL to gzip file
+        type: string
+        required: false
+        default: https://stsci.box.com/shared/static/0dt9z6b927iqgtify2a4cvls9hvapi6k.gz
       minimal:
-        description: minimal dataset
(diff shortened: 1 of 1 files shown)


.github/workflows/download_data.yml
```diff
@@ -41,25 +46,33 @@ on:
...
   schedule:
     - cron: "0 0 * * 0"
   release:
+  push:
+    branches:
+      - develop
 
 env:
-  DATA_URL: https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz
+  FULL_DATA_URL: https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz
```
BradleySappington, line 67: FULL_DATA_URL not needed, but I guess it makes sense to keep it in case we switch anything down the line

BradleySappington (approved):

PR #918, opened 2024-09-27
trivial fix to string formatting for a debug message

mperrin:
Super minor two-character syntax fix.


marcio-melendez (approved):
Looks good.

PR #923, opened 2024-10-30
update ifu docs to describe about coord_system and pixel scale

mperrin:
Improve docs to address some of the questions from that help desk ticket.


marcio-melendez (approved):
Looks good!

mperrin:
Huh. A few of the tests for this docs PR fail, in a way that’s completely unrelated to docs, which is the only thing that changed in this PR. Must be some kind of upstream change or dependency issue?? FYI @obi-wan76 @BradleySappington 
Maybe one of you could investigate

BradleySappington:
@mperrin, looks like photutils updated to version 2.0.2, last successful run it was 1.13.0

BradleySappington:
@mperrin - I pushed updates to develop that I believe should fix this issue.  If you would please merge develop into your update_ifu_docs repo, this pr should fall out and pass all tests.  I also raised an issue for the cause of this failure that will need to be addressed separately.

PR #928, opened 2024-11-03
Implemeted MRI IFU empirical cruciform

patapisp:
This is the the start of a PR for finalising the broadening and cruciform implementation for the MIRI IFU mode. Added an additional option for ```model_type = "cruciform"```

Remaining tasks: 

- [ ] check normalisation convention when modifying PSF models
- [ ] changing the ```oversample``` parameter yields quite significant differences in the PSF models
- [ ] switch empirical broadening (now following Law+2023) to broadening due to sampling by detector and cube pixels

image


mperrin:
I'm noting here for the record that the workflow on this one is going to be atypical. Rather than giving a code review to Polychronis and then having him revise the PR, I'm going to take over and be editing and finishing this one myself.

mperrin:
See #930 instead

PR #930, opened 2024-11-08
Implemeted MRI IFU empirical cruciform [v2]

mperrin:
_Cloned from #928, and relocated into a PR branch owned by @mperrin.  As noted in #928, the workflow for this one is atypical in that @patapisp started the branch and I @mperrin am now picking up to revise and finish it. By @patapisp:_

This is the the start of a PR for finalising the broadening and cruciform implementation for the MIRI IFU mode. Added an additional option for ```model_type = "cruciform"```

Remaining tasks: 

- [x] check normalisation convention when modifying PSF models
- [ ] changing the ```oversample``` parameter yields quite significant differences in the PSF models
- [ ] switch empirical broadening (now following Law+2023) to broadening due to sampling by detector and cube pixels

image


mperrin:
Future development and implementation of this PR will be at https://github.com/spacetelescope/stpsf/pull/53

PR #932, opened 2024-11-19
Add automatic data download

WilliamJamieson:
This PR makes it so that instead of erroring out when `$WEBBPSF_PATH` is not set, that instead it attempts to download the data using the link in the [documentation](https://webbpsf.readthedocs.io/en/latest/installation.html#installing-the-required-data-files) to the directory suggested for storing the data by the documentation. Note that no download will occur if the location suggested by the docs is non-empty; this saves users from constantly re-downloading the data. Warnings are issued to the user in the case that `$WEBBPSF_PATH` is not set and then another warning if a download is attempted.

### DIFF
--- a/webbpsf/utils.py
+++ b/webbpsf/utils.py
@@ -180,6 +180,39 @@ MISSING_WEBBPSF_DATA_MESSAGE = """
 """
 
 
+def auto_download_webbpsf_data():
+    import os
+    import tarfile
+    from pathlib import Path
+    from tempfile import TemporaryDirectory
+    from urllib.request import urlretrieve
+
+    # Create a default directory for the data files
+    default_path = Path.home() / "data" / "webbpsf-data"
+    default_path.mkdir(parents=True, exist_ok=True)
+
+    os.environ["WEBBPSF_PATH"] = str(default_path)
+
+    # Download the data files if the directory is empty
+    if not any(default_path.iterdir()):
+        warnings.warn("WebbPSF data files not found in default location, attempting to download them now...")
+
+        with TemporaryDirectory() as tmpdir:
+            # Download the data files to a temporary directory
+            url = "https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz"
+            filename = Path(tmpdir) / "webbpsf-data-LATEST.tar.gz"
+            urlretrieve(url, filename)
+
+            # Extract the tarball
+            with tarfile.open(filename, "r:gz") as tar:
+                tar.extractall(default_path.parent, filter="fully_trusted")
+
+        if not any(default_path.iterdir()):
+            raise IOError(f"Failed to get and extract WebbPSF data files to {default_path}")
+
+    return default_path
+
+
 def get_webbpsf_data_path(data_version_min=None, return_version=False):
     """Get the WebbPSF data path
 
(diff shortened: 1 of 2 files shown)


webbpsf/utils.py
```diff
@@ -180,6 +180,39 @@ def setup_logging(level='INFO', filename=None):
...
+
+        with TemporaryDirectory() as tmpdir:
+            # Download the data files to a temporary directory
+            url = "https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz"
+            filename = Path(tmpdir) / "webbpsf-data-LATEST.tar.gz"
+            urlretrieve(url, filename)
+
+            # Extract the tarball
+            with tarfile.open(filename, "r:gz") as tar:
+                tar.extractall(default_path.parent, filter="fully_trusted")
```
BradleySappington, line 208: the filter parameter is implemented with python 3.12.  This repo currently supports >=3.10

BradleySappington (changes requested):
Found a small issue, but overall this looks great

mperrin:
Thanks @WilliamJamieson!

This is a small point but I wanted to suggest we think about a different default path. FYI we are shortly going to be migrating “webbpsf” to “stpsf” to reflect that it’s Webb + Roman both. So perhaps we can preemptively avoid creating directories with the string “webbpsf” hard coded in them.

WilliamJamieson:
> Thanks @WilliamJamieson!
> 
> 
> 
> This is a small point but I wanted to suggest we think about a different default path. FYI we are shortly going to be migrating “webbpsf” to “stpsf” to reflect that it’s Webb + Roman both. So perhaps we can preemptively avoid creating directories with the string “webbpsf” hard coded in them. 
> 
> 

This is fine with me, I only picked that as a default because that's what the docs used. I'm happy to change it to whatever you would like and update the docs to reflect the new default.

webbpsf/utils.py
WilliamJamieson: It is in Python 3.10, see https://docs.python.org/3.10/library/tarfile.html#tarfile.TarFile.extractall. In fact, it exists all the way back to 3.8, see https://docs.python.org/3.8/library/tarfile.html#tarfile.TarFile.extractall... I think there is a bug in the python docs related to this.

BradleySappington:
@WilliamJamieson @mperrin, I'm actually okay with it staying at WebbPSF as long as this gets settled and merged in sooner than later.  I will be doing plenty of replacements, so keeping it in line with the documentation is probably our best bet.  Or if you choose not to use webbpsf, make sure you change the documentation to match.  However if you think this will not be in within a week or so please change to something other than webbpsf (and include the doc change).

WilliamJamieson:
> @WilliamJamieson @mperrin, I'm actually okay with it staying at WebbPSF as long as this gets settled and merged in sooner than later. I will be doing plenty of replacements, so keeping it in line with the documentation is probably our best bet. Or if you choose not to use webbpsf, make sure you change the documentation to match. However if you think this will not be in within a week or so please change to something other than webbpsf (and include the doc change).

I wonder if it is possible to hard code the URL and the default location and then just have the [doc page](https://webbpsf.readthedocs.io/en/latest/installation.html#installing-the-required-data-files) use that value

webbpsf/utils.py
BradleySappington: sounds good to me then

WilliamJamieson:
Note that I think it should be possible to have it pull the latest data version when this fails: https://github.com/spacetelescope/webbpsf/blob/629d73e878a7d2ed36e3e3876e57884a26a11b9b/webbpsf/utils.py#L214-L238 rather than erroring out. But that was beyond the scope of what I needed.

BradleySappington (approved):

PR #934, opened 2024-11-22
switch a test case float comparison to np.allclose for robustness

mperrin:
Minor test case fix. Found a case where it was doing a precise `==` comparison on two floats, which led to intermittent sometimes-failing and sometimes-working due to floating point precision issues. I switched it to use `np.isclose()` which will apply appropriate tolerances for a floating point comparison.


BradleySappington (approved):
lgtm

PR #935, opened 2024-11-22
update photutils upper bound

BradleySappington:
fix for photutils has been merged to develop.  Remove the upper bound


mperrin (approved):
👍 Thanks!

PR #936, opened 2024-11-27
Minor update to docs about the WFI optical model

Skyhawk172:
This PR adds a note about the current WFI optical model and the ongoing work to implement the new one. The note is added to (1) README.rst and (2) roman.rst.

After a few tries, I think the rendition on Github and `readthedocs` should be as expected.


marcio-melendez (approved):
looks good

PR #937, opened 2024-11-30
Add support for new sensing point and target phase map

marcio-melendez:
This PR adds support for the upcoming new field point in the wavefront sensing maintenance program 
Rationale:
Around December 2024 the wavefront sensing and control maintenance program will start using a new field point as we changed from NIRCAm A3 FP1 to NIRCam A1 FP6. The reason for this change is to reduce the effect of persistence that could affect nircam observations that are schedule right after the sensing program observations that use a bright target. 

1. For this PR we generated a new webbpsf target phase map from the official phase map created  by Randal Telfer using the sensing images collected during PID= 04556, observation007. The official target phase map has been release under: PRDOPSFLT-047/WAS_ConfigXML/FLIGHT_2564021O/3_was_conf_fa_trg_phase_maps.fits (extension PHASE_37). This new webbpsf target phase map is named : wss_target_phase_fp6.fits. 

For this PR we need pysiaf [0.23.3](https://github.com/spacetelescope/pysiaf/tree/0.23.3) that uses PRDOPSSOC-067. This release is the one that add support for the new NIRCam apertures, namely:
pps_aper         detector  subarray   exp_type   filter
NRCA1_FP6        NRCA1     SUB8FP6A   NRC_IMAGE  F212N
NRCA1_FP6_SUB8   NRCA1     SUB8FP6A   NRC_IMAGE  F212N
NRCA1_FP6_SUB64  NRCA1     SUB64FP6A  NRC_TACQ   F212N

The new target phase file, wss_target_phase_fp6.fits,  is under webbpsf-data-source (internal only); this file is just an upscaled version of the original phase map 3_was_conf_fa_trg_phase_maps.fits (256x256) to a 1024x1024 size in order to match webbpsf default 1024x1024 JWST pupil.  

This is a comparison of the target phase map, current (left) and new (right) 
image
siaf


marcio-melendez:
This is an example by using load_wss_opd with the new target phase map. For the OPD I used PID 4556 Obs. 7 in order to create an OPD using WAS. Then I replaced the aperture name "NRCA1_FULL" to "NRCA1_FP6" in order to match the SIAF aperture name.  
image

marcio-melendez:
For testing. Here is an OPD file created using  PID 4556 Obs. 7 (test of new field point) with the correct aperture name as expected from SIAF. This file can be add to the webbpsf-data folder, under MAST_JWST_WSS_OPDs. Then you can simply load this file:
```
nrc = webbpsf.NIRCam()
nrc.filter='F200W'
nrc.detector = 'NRCB2'
nrc.detector_position = (1024,1024)
nrc.load_wss_opd('R2024120101-NRCA1_FP6-0.fits', plot = True)
```
Note that this is not an official MAST OPD, this file is for testing only. 
[R2024120101-NRCA1_FP6-0.fits.zip](https://github.com/user-attachments/files/17979919/R2024120101-NRCA1_FP6-0.fits.zip)

webbpsf/trending.py
```diff
@@ -313,11 +313,20 @@ def wfe_histogram_plot(
...
+                # Get WSS Target Phase Map for the sensing aperture
+                # Note that the sensing maintenance program changed field point from NRC A3 to A1 around Dec 2024.
+                if sensing_apername == 'NRCA3_FP1':
+                    was_targ_file = os.path.join(
                     webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp1.fits'
                 )
+                elif sensing_apername == 'NRCA1_FP6':
+                    was_targ_file = os.path.join(
+                        webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp6.fits'
+                    )
```
mperrin: This if/elif statement is repeated five times, with no significant variation. This cries out to be refactored into a function, following the "Don't repeat yourself" principle. Here: 

```
def get_target_phase_map_filename(apername):
    """Get WSS Target Phase Map for the specified aperture
    Note that the sensing maintenance program changed field point from NRC A3 to A1 around Dec 2024.
    """
    if apername == 'NRCA3_FP1':
        fn = 'wss_target_phase_fp1.fits'
    elif sensing_apername == 'NRCA1_FP6':
        fn = 'wss_target_phase_fp6.fits'
    else:
        raise ValueError(f"Target phase map not available for aperture = {apername}")

    was_targ_file = os.path.join(
        webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', fn)
    return was_targ_file
```

I suggest putting that in utils.py, then in each of the places in the code you can just call that function.

webbpsf/trending.py
```diff
@@ -628,18 +638,44 @@ def single_measurement_trending_plot(
...
+        if prev_sensing_apername == 'NRCA3_FP1':
+            prev_was_targ_file = os.path.join(
+                webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp1.fits'
+            )
+        elif prev_sensing_apername == 'NRCA1_FP6':
+            prev_was_targ_file = os.path.join(
+                webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp6.fits'
+            )
+
+
```
mperrin, line 645: So for instance this block would become just 
```
was_targ_file = utils.get_target_phase_map_filename(sensing_apername)
prev_was_targ_file = utils.get_target_phase_map_filename(prev_sensing_apername)
```

webbpsf/trending.py
```diff
@@ -961,9 +997,20 @@ def vprint(*text):
...
+    # Note that the sensing maintenance program changed field point from NRC A3 to A1 around Dec 2024.
+    if sensing_apername == 'NRCA3_FP1':
+        was_targ_file = os.path.join(
+            webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp1.fits'
+        )
+    elif sensing_apername == 'NRCA1_FP6':
+        was_targ_file = os.path.join(
+            webbpsf.utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp6.fits'
+        )
+
```
mperrin, line 984: and similarly here
```
was_targ_file = utils.get_target_phase_map_filename(sensing_apername)
```

webbpsf/webbpsf_core.py
```diff
@@ -1720,10 +1720,17 @@ def load_wss_opd(self, filename, output_path=None, backout_si_wfe=True, verbose=
...
+            # special case: for the main sensing points FP1 or FP6, we use the official WAS target phase map,
+            # rather than the WebbPSF-internal SI WFE model.
+
+            # Select correct target phase map based on sensing field point.
+            # Note that the sensing maintenance program changed field point from NRC A3 to A1 around Dec 2024.
+            if sensing_apername == 'NRCA3_FP1':
+                was_targ_file = os.path.join(utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp1.fits')
+            elif sensing_apername == 'NRCA1_FP6':
+                was_targ_file = os.path.join(utils.get_webbpsf_data_path(), 'NIRCam', 'OPD', 'wss_target_phase_fp6.fits')
+
```
mperrin: And similarly here, though an if will still be needed in this part of the code since it can also have any other arbitrary apernames here too

```
if sensing_apername in ['NRCA3_FP1', 'NRCA1_FP6']:
    was_targ_file = utils.get_target_phase_map_filename(sensing_apername)    
```

mperrin (changes requested):
Thanks for putting this together, and getting the data files ready too! I think the code can be simplified significantly by writing a utility function to avoid so much code repetition. See comments...

webbpsf/trending.py
marcio-melendez: Thanks, much appreciated your suggestion and I fully agree with you. I added the function to utils and change all the code accordingly.

webbpsf/trending.py
marcio-melendez: done

webbpsf/trending.py
marcio-melendez: done

webbpsf/webbpsf_core.py
marcio-melendez: done

requirements.txt
```diff
@@ -4,7 +4,7 @@ matplotlib>=3.9.1,<3.10.0
 numpy>=2.1.0,<2.2.0
 photutils>=1.13.0,<2.1.0
 poppy>=1.0.0
-pysiaf>=0.22.0,<0.24.0
+pysiaf>=0.23.3,<0.24.1
```
mperrin: Should this be "<=0.24.1"? 

I would think that the current most recent released version 0.24.1 is usable here, yes?

webbpsf/trending.py
```diff
@@ -628,18 +631,27 @@ def single_measurement_trending_plot(
...
+        # Get WSS Target Phase Map for the sensing aperture
+        # Note that the sensing maintenance program changed field point from NRC A3 to A1 around Dec 2024.
+        was_targ_file = webbpsf.utils.get_target_phase_map_filename(sensing_apername)
+        prev_was_targ_file = webbpsf.utils.get_target_phase_map_filename(prev_sensing_apername)
+
+
         target_1024 = astropy.io.fits.getdata(was_targ_file)
         target_256 = poppy.utils.krebin(target_1024, (256, 256)) / 16  # scale factor for rebinning w/out increasing values
 
+        prev_target_1024 = astropy.io.fits.getdata(prev_was_targ_file)
```
mperrin: A small performance efficiency improvement would be to wrap this in a check for whether the target file is the same or not: 
```
    if prev_was_targ_file != was_targ_file:
        prev_target_1024 = astropy.io.fits.getdata(prev_was_targ_file)
        prev_target_256 = poppy.utils.krebin(prev_target_1024,
                                         (256, 256)) / 16  # scale factor for rebinning w/out increasing values
    else:
        prev_target_256 = target_256
```

That would save a bunch of redundant math for the common case of the prev and current phase maps being the same.

mperrin (commented):
👍 Looks good overall. I wanted to check one thing about the requirements: I think it should be <= instead of < for the current pysiaf 0.24.1. Yes?  

Also, we should also update the dependency versions in `pyproject.toml` too. (That's the more official way to specify the dependencies in the latest standards for python packaging, and pyproject.toml is what pip uses when computing dependencies unless you explicitly tell it to install using the requirements.txt file.) 

One additional optional suggested small revision, see comment below about a minor runtime efficiency improvement.

webbpsf/trending.py
marcio-melendez: Good point, fixed it. Thanks!

requirements.txt
marcio-melendez: Yes, good catch. Fixed.

marcio-melendez:
> Also, we should also update the dependency versions in `pyproject.toml` too. (That's the more official way to specify the dependencies in the latest standards for python packaging, and pyproject.toml is what pip uses when computing dependencies unless you explicitly tell it to install using the requirements.txt file.)
> 

I added the dependency on `pyproject.toml`

mperrin (approved):
Looks good thanks!

PR #943, opened 2024-12-04
update monthly trending plot to allow shifted date ranges

mperrin:
A small enhancement to `monthly_trending_plot` inspired by having a relatively big mirror tilt event on Dec 1.  This adds a new parameter `shift_dates_offset` to allow shifting the time period, for better display of trends around month boundaries.

```
webbpsf.trending.monthly_trending_plot(2024, 11, vmax=50, shift_dates_offset=7)
```

Unknown-6

### DIFF
--- a/webbpsf/trending.py
+++ b/webbpsf/trending.py
@@ -1106,12 +1106,16 @@ def filter_opdtable_for_daterange(start_date, end_date, opdtable):
     return opdtable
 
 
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
Looks good. I did the test on another large event at the end of February and it works as intended. 
image

PR #945, opened 2024-12-06
update MAST query for OPDs to also retrieve OPDs from FP6

mperrin:
A little bit more for the field point switch, from testing this evening. 
- Update `retrieve_mast_opd_table` to retrieve from both FP1 + FP6 by default.  This is necessary for it to find the OPDs from the new field point, for use in all the trending tools. 
- Update `load_wss_opd_by_date` debug/verbose message to include which field point was used.  This is just a small cosmetic convenience.


marcio-melendez (approved):
Looks good, thanks!

PR #948, opened 2024-12-09
v1.5.0

BradleySappington:
Version 1.5.0
Final functional changes to webbpsf before transition to STPSF


marcio-melendez (approved):
It looks good. I checked the URLs for the data release and everything looks as expected. I also performed some ofher checks, everything is a expected from previous merges  Thanks!

PR #959, opened 2025-02-13
fix download workflow

braingram:
The download workflow is failing for romancal resulting in failures downloading webbpsf data:
https://github.com/spacetelescope/romancal/actions/runs/13316268671

This is likely due to changes to the `actions/cache` infrastruction:
https://github.com/actions/cache/releases/tag/v4.2.0
with the important bit being:
> We are deprecating some versions of this action. We recommend upgrading to version v4 or v3 as soon as possible before February 1st, 2025. (Upgrade instructions below).
> 
> If you are using pinned SHAs, please use the SHAs of versions v4.2.0 or v3.4.0
>
> If you do not upgrade, all workflow runs using any of the deprecated [actions/cache](https://github.com/actions/cache) will fail.

Since the workflow has an exact pin to v4.0.2 this may explain why the caches suddenly started failing.

### DIFF
--- a/.github/workflows/download_data.yml
+++ b/.github/workflows/download_data.yml
@@ -81,7 +81,7 @@ jobs:
         run: echo "version=$(cat ${{ steps.cache_path.outputs.cache_path }}/webbpsf-data/version.txt)" >> $GITHUB_OUTPUT
       - id: cache_key
         run: echo "cache_key=webbpsf-data-${{ (github.event_name == 'schedule' || github.event_name == 'release') && 'mini' || inputs.minimal && 'mini' || 'full' }}-${{ steps.version.outputs.version }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


braingram:
@BradleySappington I can't seem to request reviewers. Would you be able to do this for me?

braingram:
It looks like the caches here are similarly impacted. The fix in this PR won't work until merged, a cache is updated (I see none for this repository) and the pin updated here:
https://github.com/spacetelescope/webbpsf/blob/207f724a881a224225f2e57243cefa09ffe4c04b/.github/workflows/ci_workflows.yml#L7

BradleySappington (approved):
LGTM

BradleySappington (changes requested):
this is being handled already in a separate PR, thank you though!

PR #960, opened 2025-02-14
only using mini in workflow.  name accordingly

BradleySappington:
stop populating "full" cache filename as we are only using "mini"


york-stsci:
The change looks reasonable. As I understand it it, this PR is expected to fail testing until merged into develop? I'd like to specifically confirm that here before approving it.

BradleySappington:
current cache checks will fail here, correct, however once I run them again in develop they should pass.

BradleySappington:
Its the same update we did in STPSF

york-stsci (approved):
As noted in main thread, looks good.

PR #961, opened 2025-02-14
V2 release prep

BradleySappington:
release preparation for v2.0.0

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,0 @@
-version: 2
-updates:
-- package-ecosystem: pip
-  directory: "/"
-  schedule:
-    interval: monthly
-    time: "10:00"
-  open-pull-requests-limit: 10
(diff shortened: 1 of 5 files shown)


BradleySappington:
awaiting feedback from folks looking at the testpypi install:
`pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ legacy-webbpsf`

marcio-melendez (approved):
I tested the alias on a clean environment and installation. I imported webbpsf and it loaded the alias with the deprecation warning as expected:
DeprecationWarning: "
    This message is for information only and WebbPSF will continue to function as normal.
    The WebbPSF library has been moved/renamed to STPSF.
    Please see https://stpsf.readthedocs.io/en/stable/ for more information.
    WebbPSF is now an alias of STPSF and is running code from the STPSF library.

I was able to run webbpsf as expected. The alias uses STPSF_DATA path but if the users set WEBBPSF_DATA then it gets copied to STPSF_DATA . 
Overall, a webbpsf user is transparent to STPSF running behind as an alias, i.e., a user can install webbpsf, download the latest webbpsf-data files and run webbpsf with no problems despite the fact that STPSF is the functional code.

